PR #12, opened 2015-05-03
Added support for new Map and List elements

nealshail:


brandongoode:
First thanks for the pull request.  It a big one and will take me some time to go through it.

There are a few things that I think could break backwards compatibility.  There are several users (including myself) that manage existing DB's and I must ensure that it is fully backwards compatible.  I  noticed that the array and object types had been replaced with the list and map types to align with DynomoDB's "new" types. I like the new type, but we need to keep the old ones too.  

Also, Travis CI failed the build.  Haven't had time to look into why.  I'll need to resolve these before merging. 

I will continue to review the changes.

Again, thanks for the excellent PR.  It has some nice additional functionality that has been needed.

nealshail:
Hi Brandon,  I understand there may well be some compatibility issues with other use cases (array and object), I just made sure the unit tests were running ok, and there were tests for those. Perhaps these can be added?

Please let me know if you have any questions about the features I've added. There will definitely be some things i could have made a little neater, so feel free to modify as you wish.

One thing I now remember I haven't done is add documentation for the Virtual, static and methods. But essentially they should operate exactly as mongoose.

nealshail:
The last feature I added (allow undefined values to be handled as strings) was a poor mans version of the SchemaType::Mixed whereby I simply pass back any undefined and unparsed parts of the document as is, instead of discarding them.. In my application code I've got a large variable format document (actually an XML document converted to JSON) which has a variable document structure. I simply use 'dynamodb-marshaler' to decode this part in my application code. I did it this way since I didn't want to bring in another dependancy to 'dynamodb-marshaler' into dynamoose and I couldn't see a better way of decoding 'unknown' document structures with the code already in the project. Perhaps if you have any better ideas of how to handle Mixed elements of the document structure?

nealshail:
sorry I was using the old 2014 version of dynamodb_local in travis.. Travis CL should now be good.

brandongoode:
Don't worry about the documentation.  Your comments are good enough to be used directly (i.e. I can copy and paste with best).  

Good catch with the dynamodb_local.  I would have made the same mistake (again).

I used the object type to handle undefined stuff and the array type to handle arrays of undefined stuff. They uses `JSON. stringify` to store the objects as JSON strings, however; I can see where this would be an issue if you needed to preserve undefined/null values.

I plan to update my existing test applications with the pull request and see what fun errors occur.  It will take me a few days to have time to do this, but I hope that I'll be able to get to it this week.

brandongoode:
@nealshail,

I've been working on this pull request.  I've had to make some changes to enable support for existing data stored with the object/array JSON strings.   I've added an option that allows users to choose to use the map/list document types for a model.  I have to leave the default as object/array to keep the backwards compatibility.

I've had an issue with the namespace addition.  If you don't assign the namespace, "undefined" is added.  I've fixed this, but what was the purpose for adding the namespace vs. using the prefix?

Overall I think the changes add some good of functionality and will be merging them into the master branch soon.

nealshail:
Great to hear you've made progress.

The namespace was in response to the limitation posed by Dynamodb to one database per AWS account. If you have multiple applications in the same AWS account, then there is no way group separate tables/collections with that application. So the suggestion is to prefix all tables with a 'namespace'. see the following discussion. I did notice that the prefix could be added at a lower level. But i just wanted a way to set it once for the application and have it propagate throughout.
http://stackoverflow.com/questions/25019380/multiple-applications-and-dynamodb-ideal-way-to-configure
Great that you fixed the error case with no namespace is defined!

Yes I understand you need to consider backwards compatibility.

Will be keen to see when you are finished with the changes.!

brandongoode:
The prefix can also be set globally and overridden at the model level.  This was not documented (my bad).

*Global*
```js
dynamoose.setDefaults({ prefix: 'dev-' });
```

*Model Override*
```js
dynamoose.model('table', schema, { prefix: 'staging-' });
```

To reduce confusion (for me), would it be acceptable to use the prefix and remove the namespace?

nealshail:
This is great. Yes we should just use the one way..   I am happy to adjust my code to use the prefix.

brandongoode:
Cool -- I'll make the changes.  I'll try to push the updates to the master branch in the next 24 hours.

brandongoode:
When you get a chance please pull the master and see if everything is still working.  Once you confirm, I'll update the documentation and push a new version.

nealshail:
Thanks Brandon, Sorry, have been super busy..  Will pull it down now and have a look as soon as i can.

PR #14, opened 2015-07-30
implemented batchGet, batchPut and batchDelete

ricardopinto:
implemented batchGet, batchPut and batchDelete


ricardopinto:
batchGet was already implemented by @khamasaki

PR #21, opened 2015-09-10
Add regex attribute names to map keys.

benjcooley:
This change allows maps to store arbitrary sets of child elements as long as the property names of those elements match a regex named attribute.   If there are two or more regex named attributes in the map that are mutually exclusive, the map may contain two or more types of data.  Regex named attributes and ordinary named attributes can coexist in the same map allowing there to be single instances of some data, and multiple instances of other.

Example:

```
myMap: {
  type: 'map',
  map: {
    "/.*/" : {
      type: 'map',
      map: {
        name: String
      }
    }
  }
```

This matches the following JSON:

```
myMap: {
  Foo: { name : "Foo" },
  Bar: { name: "Bar" },
  Too: { name: "Too" }
} 
```

This essentially turns maps into dynamic sets, allowing you to store sets of multiple arbitrarily complex nested json items in a dynamodb set.

A typical use scenario is creating a 'map' in a schema that is keyed by either a record hashKey or some other unique id (uuid?) that stores some sort form of cached information for that item.  

Note that the dynamodb update() method can insert and remove items from these maps at any level in the hierarchy using the $PUT command.   To add you $PUT the data for a given map key, to remove you $PUT a null value for that key.

```
// Add item to map
MyModel.update({ id: "00012341234" }, { $PUT: { "myMap.Foo" : { name: "Foo" } } });

// Remove item from map
MyModel.update({ id: "00012341234" }, { $PUT: { "myMap.Foo" : null } });
```

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -335,8 +335,8 @@ Attribute.prototype.toDynamo = function(val, noSet, model) {
     for(var name in this.attributes) {
       var attr = this.attributes[name];
       attr.setDefault(model);
-      if (this.nameRegex) {
-        var regex = this.nameRegex;
+      if (attr.nameRegex) {
+        var regex = attr.nameRegex;
         for(var propName in val) {
           if (regex.test(propName)) {
             dynamoAttr = attr.toDynamo(val[propName], undefined, model);
@@ -431,7 +431,7 @@ Attribute.prototype.parseDynamo = function(json) {
         var childAttr = attr.attributes[attrName];
         if (childAttr.nameRegex) {
           var regex = childAttr.nameRegex;
-          for (propName in v) {
+          for (var propName in v) {
             if (regex.test(propName)) {
               attrVal = childAttr.parseDynamo(v[propName]);
               if(attrVal !== undefined && attrVal !== null){
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -18,7 +18,7 @@ function Attribute(schema, name, value) {
   this.name = name;
 
   // If name begins/ends with a slash it is a regex used to match properties in a 'map'
-  if (name && name.substr(0, 1) === "/" && name.substr(name.length - 1, 1) === "/") {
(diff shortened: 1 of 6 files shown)


benjcooley:
Could you let me sit on this for a bit while I add tests?  

Wanted to see if there was any interest before going to the trouble.

brandongoode:
I like the idea.  I'll wait untill you have the test cases.

Instead of (or in addtion to) using a regex as a string, could we use the regex? You could then check if `name instanceof RegExp` and set the name to the `RegExp.toString()`

Just a thought.

benjcooley:
Been very busy, but I have added some test for this in my own project repo.  Will try to separate them and move them into this repo, as well as just rebase this PR and resubmit.

Also now have full middleware support for redis and elasticsearch that I can possibly npm separately.  Works quite nicely and you can just drop it in with use() and select which models are cached and which fields are searchable, set search options, etc.  Doesn't yet support the 'callback' style calls yet, and also doesn't support batch operations.

benjcooley:
If I can get time to wrap it up, I'll call the middleware 'dynamoose-redis' and 'dynamoose-elasticsearch' if you don't mind.

brandongoode:
Great work.  Don't have time right now to look over all of it, but I'll try to carve out some time to look over it in detail soon.

Naming conviention is good with me (Need to make sure we link to them in the Readme).

brandongoode:
@benjcooley -- I need to apologize.  I let this PR completely slip my mind.  Since it has been so long, I wanted to make sure you didn't have any update prior to merging it.  Again, sorry for the 3 month delay.

brandongoode:
@benjcooley - Any updates on this pull request?  Again sorry for the delay.

benjcooley:
Sorry about the delay.  I'll try to move these to separate pulls.  

It would be nice to not have to merge constantly.

brandongoode:
Cool thanks

ngosala:
any ETA on this ?

PR #25, opened 2016-04-26
Documentation fix for scan lastKey

tuxmachine:
Documentation on the lastKey is probably from before the implementation of promises, anyway a small fix that could save another dev a couple minutes.


brandongoode:
Thanks - The Readme needs all the help it can get.

tuxmachine:
That's awfully fast, thanks!

PR #31, opened 2016-05-30
Swap parameter values for put example in Readme

andybee:
As far as I can tell from testing, the 2 parameters for the first update put example are the wrong way around if both put examples are attempting to set an age for the previously defined id 4 object.


brandongoode:
Thanks for the correction -- my dyslexia got the best of me.

PR #40, opened 2016-07-07
Remove reference to testing infrastructure to prevent warnings from javascript packagers

jmather:
Not sure how you will want to move the other stuff around to continue getting coverage, but having this set up makes WebPack throw errors while bundling.


brandongoode:
Thanks.  I'll look into better ways to run the test coverage.  That logic is over 2 years old, and I'm sure there is a better way.

PR #43, opened 2016-07-15
Closes #42 - loosen aws-sdk

### LINKED ISSUE #42: Use version ^2.1.0 instead of 2.1.* of AWS-sdk
Is there a particular reason you've locked it down to 2.1.* rather than just using ^2.1.0.

I would imagine a company such as Amazon would understand semver and therefore would not break this within the 2.* range. This impacts on other libraries that may use ^2.3.0 for example, in that they are using two individual AWS-sdk instances, rather than the same instance.

davidhouweling:
## Description
Loosen the AWS-sdk dependency so that duplications don't occur between other libraries that may also rely on AWS-sdk, or a local dependency.

## Motivation and Context
In a project I am working on, we depended on a more recent version of AWS-sdk, but this library was using its own version, so that it can use the same major version, we should loosen the dependency.

## Types of changes
- Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
   },
   "dependencies": {
     "debug": "*",
-    "aws-sdk": "2.1.*",
+    "aws-sdk": "^2.1.0",
     "q": "~1.0.1",
     "hooks": "0.3.2"
   }


brandongoode:
Thanks

PR #51, opened 2016-08-03
Adding option when updating to allow empty array

fishcharlie:
There is probably a better way to do this and I would love someone to test just to double check to make sure I'm not going crazy before merging into master.


brandongoode:
Thanks -- Looked good -- I updated the option name to better align with the other options styling.

PR #73, opened 2016-10-12
GSI Synchronization (should be ready for acceptance proofs)

jottinger:
This should be able to sync indexes from local to remote, given the correct update flag.


lib/Table.js
```diff
@@ -11,154 +11,299 @@ function Table(name, schema, options, base) {
...
+    ]
+  };
+  ddb.updateTable(params, function (err, data) {
+    debug('deleteIndex handler running');
+    if (err) {
+      deferred.reject(err);
+    }
+    else {
+      setTimeout(function () {
+        table.waitForActive();
```
brandongoode: You need to wait for the `waitForActive` promise to return before resolving the current `deleteIndex` promise.

```js
table.waitForActive().then(function () {
  deferred.resolve(data);
});
```

lib/Table.js
```diff
@@ -11,154 +11,299 @@ function Table(name, schema, options, base) {
...
-        debug('error initializing', err);
-        return deferred.reject(err);
-      }
-    });
+      .then(function (data) {
+        debug('table exist -- initialization done');
+        localTableReq = buildTableReq(table.name, table.schema);
+        var indexes = compareIndexes(localTableReq, data.Table);
+
+        debug('%s', JSON.stringify(indexes, null, 2));
```
brandongoode: Need to make sure to honor all the promises.  Something like this (untested):

```js
if (table.options.update) {
  return Q.all(indexes.delete.map(function (index) {
    return table.deleteIndex(index.IndexName);
  }))
  .then(function() {
    return Q.all(indexes.create.map(function (index) {
      return table.createIndex(index.IndexName);
    }));
  })
  .then(function() {
    table.initialized = true;
    return deferred.resolve();
  });
} else {

  if (indexes.delete.length > 0 || indexes.create.length > 0) {
    debug('indexes are not synchronized and update flag is set to false');
    return deferred.reject('indexes are not synchronized and update flag is set to false');
  }

  return table.waitForActive()
  .then(function () {
    table.initialized = true;
    return deferred.resolve();
  });
}
```

brandongoode:
Sorry for the delay -- Last week was a little hectic, and I just now had a chance to review the PR.   I made a couple of small tweaks.

jottinger:
No worries, and thank you!

PR #75, opened 2016-10-21
Fixed unhandled validation errors

MikePennington:
Issue: Attribute.toDynamo would throw an unhandled exception when a validation error was triggered, bypassing the promise handling.

Solution: try-catch calls to Attribute.toDynamo and handle them with deferred.reject(err).


brandongoode:
Thanks for helping to clean up the code.  Merging now.

PR #78, opened 2016-11-08
update readme

againer:
- replaced `New` with `new` keyword


brandongoode:
Thanks -- sorry for the delay.

PR #89, opened 2016-12-24
Handle thrown errors and pass them to callbacks and promises

joshuadutton:
Recently, there have been improvements on how validation errors have been handled, but I noticed that there were still a few cases that weren't handled: 

1. In `Model` there are calls to error throwing methods `Model.processCondition`, `Schema.parseDynamo`, `Schema.toDynamo`, and `Attribute.toDynamo` that aren't surrounded by `try/catch` with the error passed to the callback and promise.
2. `Query` and `Scan` building throw errors that need to be handled by the caller, even though there is an error mechanism in the callback or promise when `exec` is called.

Fixes:
- In `Model` I started by adding more `try/catch` statements at the call sites, but I realized that the logic could be simplified by doing `try/catch` statements at a higher level. So I added those and removed the others. 
- For `Query` and `Scan` building, I'm passing the first error encountered along so that it is sent to the callback and promise when `exec` is called.

Possible problems: 
- Passing an error along instead of throwing changes the stack trace when the error is handled. Since most users of this library probably don't care about the implementation details and just want to know what's wrong in the calling code, this shouldn't be much of an issue. 
- There is one thrown error that isn't handled, but it is when a model is initialized (see: https://github.com/automategreen/dynamoose/blob/master/lib/Model.js#L130), so there is no callback or promise mechanism. I assume that in this case, this is still the desired outcome.

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -561,7 +561,6 @@ Model.update = function(NewModel, key, update, options, next) {
 
       var model = new NewModel();
       model.$__.isNew = false;
-
       schema.parseDynamo(model, data.Attributes);
 
       debug('updateItem parsed model', model);
@@ -785,16 +784,12 @@ Model.batchGet = function(NewModel, keys, options, next) {
 
   getReq.Keys = keys.map(function (key) {
     var ret = {};
-    schema.parseDynamo(model, data.Attributes);
-    debug('deleteItem parsed model', model);
-
     ret[hashKeyName] = schema.hashKey.toDynamo(key[hashKeyName]);
 
     if(schema.rangeKey) {
       var rangeKeyName = schema.rangeKey.name;
       ret[rangeKeyName] = schema.rangeKey.toDynamo(key[rangeKeyName]);
     }
-
     return ret;
   });
 
@@ -824,8 +819,8 @@ Model.batchGet = function(NewModel, keys, options, next) {
       function toModel (item) {
         var model = new NewModel();
         model.$__.isNew = false;
-
         schema.parseDynamo(model, item);
+
         debug('batchGet parsed model', model);
 
         return model;
@@ -836,7 +831,6 @@ Model.batchGet = function(NewModel, keys, options, next) {
         // convert unprocessed keys back to dynamoose format
         models.unprocessed = data.UnprocessedKeys[newModel$.name].Keys.map(function (key) {
           var ret = {};
-
           ret[hashKeyName] = schema.hashKey.parseDynamo(key[hashKeyName]);
 
           if(schema.rangeKey) {
@@ -994,7 +988,6 @@ Model.batchDelete = function(NewModel, keys, options, next) {
 
   var batchRequests = toBatchChunks(newModel$.name, keys, MAX_BATCH_WRITE_SIZE, function(key) {
     var key_element = {};
-
     key_element[hashKeyName] = schema.hashKey.toDynamo(key[hashKeyName]);
 
     if(schema.rangeKey) {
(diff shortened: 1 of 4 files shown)


brandongoode:
Thanks for helping clean up the code.   I won't have much time over the next couple days to review it.  I'll need to understand what's causing the test failures.

```log
118 passing (11s)
  5 failing

  1) Model Deletes item with invalid key:
     ValidationError: Validation failed: id
    at Attribute.toDynamo (lib/Attribute.js:1281:7)
    at NewModel.Model.delete (lib/Model.js:2524:45)
    at Context.<anonymous> (test/Model.js:335:15)
  

  2) Scan Scan with not beginsWith (error):
     AssertionError: expected Function { name: '' } to throw exception with a message matching 'Invalid scan state: beginsWith() cannot follow not()', but got 'expected true to not exist'
    at Assertion.fail (node_modules/should/cjs/should.js:231:17)
    at Assertion.value (node_modules/should/cjs/should.js:305:19)
    at Context.<anonymous> (test/Scan.js:473:20)
  

  3) Scan Scan with not in (error):
     AssertionError: expected Function { name: '' } to throw exception with a message matching 'Invalid scan state: in() cannot follow not()', but got 'expected true to not exist'
    at Assertion.fail (node_modules/should/cjs/should.js:231:17)
    at Assertion.value (node_modules/should/cjs/should.js:305:19)
    at Context.<anonymous> (test/Scan.js:505:20)
  

  4) Scan Scan with not between (error):
     AssertionError: expected Function { name: '' } to throw exception with a message matching 'Invalid scan state: between() cannot follow not()', but got 'expected true to not exist'
    at Assertion.fail (node_modules/should/cjs/should.js:231:17)
    at Assertion.value (node_modules/should/cjs/should.js:305:19)
    at Context.<anonymous> (test/Scan.js:537:20)
  

  5) Scan Scan with ANDed filter with filter object (error):
     AssertionError: expected Function { name: '' } to throw exception with a message matching 'Invalid scan state; %s can only be used once', but got 'expected true to not exist'
    at Assertion.fail (node_modules/should/cjs/should.js:231:17)
    at Assertion.value (node_modules/should/cjs/should.js:305:19)
    at Context.<anonymous> (test/Scan.js:603:20)
```

Thanks again and Merry Christmas

joshuadutton:
Sorry I haven't been responsive. I've been out for the holidays.

- Failed test 1 is failing because of a bug in my changes. I addressed that, but I haven't pushed it up yet. I will do that early next week.
- Failed tests 2-5 are failing because they are expecting the functions they are calling to throw and they are not anymore. If we continue forward with my approach, those tests will need to change. 

Also, I've realized that one part of my approach is not ideal: since the callback could be one of two arguments, 'options' or 'next', and I am catching errors now at a higher level, I have to do that check before I pass the error on (see my helper function `sendErrorToCallback`). This now duplicates some of the logic that is already done in each of the functions that are being called within the `try` blocks. I want your feedback if you prefer this approach or if I should go back to putting more `try/catch` blocks at a lower level just around the specific functions that throw (`Model.processCondition`, `Schema.parseDynamo`, `Schema.toDynamo`, and `Attribute.toDynamo`).

Thank you, and Happy New Year!

brandongoode:
My thoughts, for errors 2-5 is the promise should not resolve successfully, but should resolve with an error (i.e. reject).

joshuadutton:
Great. That's how I've implemented it, so I will change the tests to reflect that.

brandongoode:
@joshuadutton - I'm planning on reviewing and merging sometime this weekend, but I wanted first to check and make sure you didn't have any more updates.  Thanks!

joshuadutton:
@brandongoode I don't have any other planned updates. Thanks.

brandongoode:
Sorry for the delay -- looks great. Thanks.

Issue #85, opened 2016-12-08
Model.update does not create 'required' fields, create 'createAt' or create 'updateAt' timestamps.


aloof-ruf:
`Model.update()` works well when the desired item already exists in the database, but does not do the following when the item does not exist:

1. create any attributes set as 'required' (nor generate defaults)
2. create 'createdAt' or 'updatedAt' attributes/timestamps

For example:

**apple.js** (define schema)
```
var dynamoose = require('dynamoose');

dynamoose.AWS.config.update({
  region: 'whatever'
});

dynamoose.local();

var appleSchema = new dynamoose.Schema({
  id: {
    hashKey: true,
    type: Number,
    required: true
  },
  appleType: {
    type: String,
    default: 'spartan',
    required: true,
    trim: true
  },
  status: {
    type: String,
    default: 'crisp',
    required: true,
    trim: true,
    validate: function (value) { return ['rotten', 'bruised', 'crisp'].indexOf(value) !== -1; }
  }
},
{
  timestamps: true
});

var apple = dynamoose.model('Apple', appleSchema);

module.exports = { 
   apple: apple
};
```

**index.js** (perform database operations)
```
var appleModel = require('./apple.js').apple;

console.log("Apples Test");

function createNewWithUpdate() {
  appleModel.update({id: 4}, {appleType: 'honeygold'}, function (error, result) {
    if (error) return console.error(error);
    console.dir(result);
  });
}

function updateApple1 () {
  appleModel.update({id: 1}, {appleType: 'gala'}, function (error, result) {
    if (error) return console.error(error);
    console.dir(result);

    createNewWithUpdate();
  });
}

appleModel.batchPut([
  {id: 1, appleType: 'granny smith'},
  {id: 2, appleType: 'spartan'},
  {id: 3, appleType: 'red delicious'}
], function (error, apples) {
  if (error) return console.error(error);
  console.dir(apples);

  setTimeout(updateApple1, 5000);
});
```

Then, the result when I run: `aws dynamodb scan --table-name Apple --endpoint-url http://localhost:8000` is:
```
{
    "Count": 4, 
    "Items": [
        {
            "status": {
                "S": "crisp"
            }, 
            "appleType": {
                "S": "spartan"
            }, 
            "id": {
                "N": "2"
            }, 
            "createdAt": {
                "N": "1481235232940"
            }, 
            "updatedAt": {
                "N": "1481235232940"
            }
        }, 
        {
            "status": {
                "S": "crisp"
[...]

PR #96, opened 2017-01-13
Issue 85 create required attributes on update

aloof-ruf:
This pull request is meant to address Issue #85. There is a lot going on here...

`Model.update` now has a new touch. If you pass a `null` or `undefined` key, then there is a check for key defaults. If the key attributes have defaults, those will be used. If they don't, _then_ an error is thrown. This check is currently not in place.

`timestamps` (as defined in the schema) are now updated by default. You can turn this behaviour off by:

1. Not having `timestamps`
2. setting the `updateTimestamps` option to `false`

But, most importantly, this PR now has the update _also_ conditionally update required attributes. There is one catch though: **required attributes that do not have a default value will throw an error on update if no value is provided for them**. 

I also added some tests for `Module.update`

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -528,7 +528,7 @@ If true, the attribute can be updated to an empty array. If falsey, empty arrays
 
 **createRequired**: boolean
 
-If true, required attributes will be filled with their default values on update (regardless of you specifying them for the update). Defaults to true.
+If true, required attributes will be filled with their default values on update (regardless of you specifying them for the update). Defaults to false.
 
 **updateTimestamps**: boolean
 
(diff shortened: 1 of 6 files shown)


brandongoode:
First -- Sorry for the delayed response. 

It looks great.

The one thing I'm unsure about is `createRequired` defaulting to `true`, because it changes the behavior of existing code.  I don't want to change to break a bunch of peoples existing code or worse overwrite their data.

What are your thoughts on changing the `createRequired` default to `false`?  What about creating global option to set the default behavior?

aloof-ruf:
Setting it to `false` is fine with me. I agree with you, these changes should _allow_ for the feature instead of imposing it on existing code. Looks like I have merge conflicts to resolve anyway

brandongoode:
Thanks a bunch for this PR.

PR #98, opened 2017-01-17
Fix bug when `next` callback is null

joshuadutton:
Missed a bug that occurs when promises are used. 

The tests don't have any tests for promises. I will add some in a future PR.


brandongoode:
Thanks -- Promises test cases would be great!

PR #101, opened 2017-01-18
Added .count() for Query and Scan

strygin:
Added .count(), .counts(), for Query and Scan using `Select`:
http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-Select 
Also some small docs cleanups.


strygin:
Fixed that.

PR #109, opened 2017-02-15
fix Attribute.prototype.toDynamo method to support Native Booleans

zarov:
Hi, just saw that the native booleans have been added, but it was missing a check when "dynamofying" the attribute. So here is a quick fix.

---
This change is [Reviewable](https://reviewable.io/reviews/automategreen/dynamoose/109)

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -374,7 +374,10 @@ Attribute.prototype.toDynamo = function(val, noSet, model) {
       val = type.dynamofy(val);
     }
 
-    val = val.toString();
+    if(type.dynamo !== 'BOOL') {
+      val = val.toString();
+    }
+
     if(type.dynamo === 'S') {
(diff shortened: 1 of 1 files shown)


brandongoode:
Thanks

luanpablo-luizalabs:
Hi, this really helps me and I'm currently installing the lib from master

any ideas of when a new release will be launched?

thanks a lot \o

zarov:
@luanpablo-luizalabs don't know about the release, but you can add the package with a SHA1 as the version, as seen here https://docs.npmjs.com/files/package.json#github-urls (it is the solution I used)

luanpablo-luizalabs:
@zarov oh, awesome, I'll do that until the version is available.

thankss

brandongoode:
@luanpablo-luizalabs - sorry for the slowness in releasing the next version.

PR #118, opened 2017-03-20
fix Model.prototype.create cannot create new item on DynamoDB if hash…

mooyoul:
…Key is one of reserved keyword.

See: http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ReservedWords.html


brandongoode:
Great addition -- Thanks.

PR #121, opened 2017-04-01
Fix critical bug with deleting and creating global secondary index

webwin0:
Dynamoose previously removed index and tried to create new. But that time was more than timeout and it's crashed app.
We shouldn't set ProvisionedThroughput to previous value after new code deploy.
ProvisionedThroughput is the dynamic value that can be managed with auto scaling script.

### DIFF
--- a/lib/Table.js
+++ b/lib/Table.js
@@ -38,11 +38,8 @@ var compareIndexes = function compareIndexes(local, remote) {
       if (remoteIndexes[j].IndexName === localIndexes[i].IndexName) {
         // let's see if the core data matches. if it doesn't,
         // we may need to delete the remote GSI and rebuild.
-        var localIndex = _.pick(localIndexes[i], 'IndexName', 'KeySchema', 'Projection', 'ProvisionedThroughput');
-        var remoteIndex = _.pick(remoteIndexes[j], 'IndexName', 'KeySchema', 'Projection', 'ProvisionedThroughput');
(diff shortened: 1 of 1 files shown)


lib/Table.js
```diff
@@ -38,8 +38,8 @@ var compareIndexes = function compareIndexes(local, remote) {
       if (remoteIndexes[j].IndexName === localIndexes[i].IndexName) {
         // let's see if the core data matches. if it doesn't,
         // we may need to delete the remote GSI and rebuild.
-        var localIndex = _.pick(localIndexes[i], 'IndexName', 'KeySchema', 'Projection', 'ProvisionedThroughput');
-        var remoteIndex = _.pick(remoteIndexes[j], 'IndexName', 'KeySchema', 'Projection', 'ProvisionedThroughput');
+        var localIndex = _.pick(localIndexes[i], 'IndexName', 'KeySchema', 'Projection');
+        var remoteIndex = _.pick(remoteIndexes[j], 'IndexName', 'KeySchema', 'Projection');
         if (remoteIndex.hasOwnProperty('ProvisionedThroughput')) {
```
brandongoode: This if statement isn't needed now.

brandongoode:
Good point. Can you also remove the if statement that follows because it's not needed?  Then I'll merge.  Thanks.

webwin0:
Updated pr.
Thanks for dynamoose :)

brandongoode:
Awesome -- Thanks

PR #123, opened 2017-04-11
Add typescript typings

breath103:
it supports most of features dynamoose has now. 
one possible future risk would be we might not maintain d.ts so that it goes out of sync with JS file,
but it seems like you guys don't really have plan to change "interfaces" of this library for awhile, so i thought it would be pretty safe to just put this in the repository


breath103:
Thanks! I'll do my best to keep up the tsd file haha

nihsmik:
👍

PR #125, opened 2017-04-11
Allow for unnamed attributes

cbarrese:
This is an attempt to address issue https://github.com/automategreen/dynamoose/issues/107

I pulled the code out of the conversation, modified slightly to get it to work, and wrote a couple of tests against it.  I'm looking for some validation that this is on the right track and what else you'd like to see in terms of test coverage and documentation in order to accept this pull request.

### DIFF
--- a/lib/Schema.js
+++ b/lib/Schema.js
@@ -87,13 +87,24 @@ function Schema(obj, options) {
 Schema.prototype.toDynamo = function(model) {
 
   var dynamoObj = {};
-  //debug('toDynamo with schema attributes', this.attributes);
-  for(var name in this.attributes) {
+
+  for(var name in model) {
+    if(!model.hasOwnProperty(name)){
+      continue;
+    }
     var attr = this.attributes[name];
-    attr.setDefault(model);
-    var dynamoAttr = attr.toDynamo(model[name], undefined, model);
-    if(dynamoAttr) {
-      dynamoObj[attr.name] = dynamoAttr;
+
+    if(!attr & this.options.saveUnknown) {
+      attr = Attribute.create(this, name, typeof model[name]);
+      this.attributes[name] = attr;
+    }
+
+    if(attr) {
+      attr.setDefault(model);
+      var dynamoAttr = attr.toDynamo(model[name], undefined, model);
+      if(dynamoAttr) {
+        dynamoObj[attr.name] = dynamoAttr;
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


lib/Schema.js
```diff
@@ -87,13 +87,24 @@ function Schema(obj, options) {
 Schema.prototype.toDynamo = function(model) {
 
   var dynamoObj = {};
-  //debug('toDynamo with schema attributes', this.attributes);
-  for(var name in this.attributes) {
+
+  for(var name in model) {
+    if(!model.hasOwnProperty(name)){
+      continue;
+    }
```
cbarrese: Needed to add this continue in order to avoid trying to turn schema, options, ..., etcetera into `Attribute` objects that would attempt to be persisted.

lib/Schema.js
```diff
@@ -87,13 +87,24 @@ function Schema(obj, options) {
...
     var attr = this.attributes[name];
-    attr.setDefault(model);
-    var dynamoAttr = attr.toDynamo(model[name], undefined, model);
-    if(dynamoAttr) {
-      dynamoObj[attr.name] = dynamoAttr;
+
+    if(!attr & this.options.saveUnknown) {
+      attr = Attribute.create(this, name, typeof model[name]);
```
cbarrese: I think the `typeof` here works better than always assuming `Object`.  `JSON.stringify` doesn't always give you what you'd want. 
```
> JSON.stringify("astring")
'"astring"'
```

lib/Schema.js
sdrioux: think you can just use Object.getOwnPropertyNames(),

cbarrese: tried that and had a bunch of test failures that didn't seem worth tracking down since the solution above worked.

test/Query.js
```diff
@@ -254,7 +254,7 @@ describe('Query', function (){
     Dog.query('breed').eq('unknown')
     .filter('color').not().null().exec()
     .then(function (dogs) {
-      dogs.length.should.eql(5);
+      dogs.length.should.eql(4);
```
cbarrese: based on the test data setup, this assertion was wrong

test/Query.js
```diff
@@ -268,7 +268,7 @@ describe('Query', function (){
     .filter('color').not().eq('Brown')
     .exec()
     .then(function (dogs) {
-      dogs.length.should.eql(1);
+      dogs.length.should.eql(2);
```
cbarrese: based on the test data setup, this assertion was wrong

brandongoode:
Thanks for the PR.  I know several people will be happy to get this.

The code and test cases look good (I wonder how the Query tests have been passing - thanks for fixing that).

Could you add a few quick line to the schema options section of the README? -
 https://github.com/automategreen/dynamoose#options

brandongoode:
Thanks a bunch.

PR #127, opened 2017-04-13
Default function enhancements

### LINKED ISSUE #124: Generated attributes: Pass current model to attribute default function.
It would be awesome if we could specify a default function for an attribute and get the current instance of the model as a parameter. That would allow us to populate the secondary index in a better way.
```javascript
const schema = {
  _id: String,
  country: String,
  state: String,
  surName: String,
  surNameIndex: {
    index: {
      default: model => {
        return `${model.country}_${model.state}_${model.surName}`;
      }
    }
  }
}
```

Seems passing the model here would do the trick:
https://github.com/automategreen/dynamoose/blob/108fcaa91fc1785dc4d3623231f6b7a16f877dc4/lib/Attribute.js#L280

More generically however, having a forceDefault boolean for the attribute to ensure the default is applied every time would also be useful. This would ensure the secondary index data is properly updated when the data is persisted.

vangorra:
Making the model get passed to attribute's default function.
Adding flag forceDefault to force the default to be used all the time.
Adding dev dependency to automatically download local dynamo.
Adding grunt task to start/stop local dynamo instance.
Adding tests to support changes.

closes #124

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,3 @@ node_js:
 env: DEBUG=dynamoose*
 before_script:
   - npm install -g grunt-cli
-  - wget http://dynamodb-local.s3-website-us-west-2.amazonaws.com/dynamodb_local_latest.tar.gz
-  - tar xfz dynamodb_local_latest.tar.gz
-  - java -Djava.library.path=./DynamoDBLocal_lib -jar DynamoDBLocal.jar -inMemory &
-
--- a/Readme.md
+++ b/Readme.md
@@ -252,9 +252,19 @@ Defines the attribute as a local or global secondary index. Index can either be 
 
 Applies a default to the attribute's value when saving, if the values is null or undefined.
 
-If default is a function, the function is called, and the response is assigned to the attribute's value.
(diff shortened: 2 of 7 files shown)


brandongoode:
Awesome!   Thanks for adding the test case. Can you update the README and then I'll merge. 

I also like the addition to the gruntfile.  We can use this to replace the travisci logic.

vangorra:
I'll update the readme and travis file.

I wanted to gauge your reaction to the changes to the grunt file as I was considering making the local dynamo tasks automatic when running the tests. That way it's even simpler for a dev to start working on this project.

brandongoode:
Haven't had time to really dig in to the travis ci failure, but we may need to add the `-inMemory` option to `.launch`.  This would also keep local testing clean. 

```js
.launch(8000, null, ['-inMemory'])
```

vangorra:
Need to take some time to reproduce these failures in various versions of node.

brandongoode:
I fixed some test case errors and now this PR is passing.  Are you good with me merging it?

vangorra:
I'm good with merging. I was seeing some odd errors but it looks like you handled it.

PR #128, opened 2017-04-14
Support GSI with same hashkey as Table

cbarrese:
This PR addresses the issue described here
https://github.com/automategreen/dynamoose/issues/86

I read through the comments in the issue, and after looking at the code it looks like I could make a change to support fixing the issue by adding the rangeKey to the check to determine whether to try to use a global secondary index, and also providing the `query.using` functionality to explicitly use a specific index if multiple are configured for the hashkey attribute.

If this solution works for you I will update the Query doc to describe the new feature.

### DIFF
--- a/lib/Query.js
+++ b/lib/Query.js
@@ -87,7 +87,7 @@ Query.prototype.exec = function (next) {
   // the hash and range key on the primary table.  If they don't match then we
   // can look for a secondary index to query.
   if(schema.hashKey.name !== this.query.hashKey.name ||
-    (this.query.rangeKey && schema.rangeKey.name !== this.query.rangeKey.name)) {
+    (this.query.rangeKey && schema.rangeKey && schema.rangeKey.name !== this.query.rangeKey.name)) {
     debug('query is on global secondary index');
     for(indexName in schema.indexes.global) {
       index = schema.indexes.global[indexName];
--- a/lib/Query.js
+++ b/lib/Query.js
@@ -83,7 +83,11 @@ Query.prototype.exec = function (next) {
   };
 
   var indexName, index;
-  if(schema.hashKey.name !== this.query.hashKey.name) {
+  // Check both hash key and range key in the query to see if they do not match
+  // the hash and range key on the primary table.  If they don't match then we
+  // can look for a secondary index to query.
+  if(schema.hashKey.name !== this.query.hashKey.name ||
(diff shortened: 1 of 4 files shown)


lib/Query.js
```diff
@@ -83,7 +83,11 @@ Query.prototype.exec = function (next) {
   };
 
   var indexName, index;
-  if(schema.hashKey.name !== this.query.hashKey.name) {
+  // Check both hash key and range key in the query to see if they do not match
+  // the hash and range key on the primary table.  If they don't match then we
+  // can look for a secondary index to query.
+  if(schema.hashKey.name !== this.query.hashKey.name ||
+    (this.query.rangeKey && schema.rangeKey.name !== this.query.rangeKey.name)) {
```
brandongoode: schema may not have a rangeKey.  

```js
(this.query.rangeKey && schema.rangeKey && schema.rangeKey.name !== this.query.rangeKey.name)
```

brandongoode:
Looks good.  I only saw the one possible issue.  I'll merge when you have the documentation ready.  Thanks for the help.

lib/Query.js
cbarrese: yep, good catch.  i'll fix that and update the docs.  thanks for the review.

cbarrese:
Made the fix to check for schema rangeKey and updated documentation.

lib/Query.js
```diff
@@ -83,7 +83,11 @@ Query.prototype.exec = function (next) {
   };
 
   var indexName, index;
-  if(schema.hashKey.name !== this.query.hashKey.name) {
+  // Check both hash key and range key in the query to see if they do not match
+  // the hash and range key on the primary table.  If they don't match then we
+  // can look for a secondary index to query.
+  if(schema.hashKey.name !== this.query.hashKey.name ||
```
brandongoode: Just thought of an issue -- we need to add a check to skip this if `.IndexName` is already defined. Else it will overwrite what was configured with `.using`.

cbarrese: So `.using` sets an option property, then I set `.IndexName` with what was specified in `.using` after all the logic to try to figure out what it should be.  This way the `.using` should always take precedence.

https://github.com/automategreen/dynamoose/pull/128/files#diff-c7f251522c2a751e0d457dee5ffb83f0R146

brandongoode: Good point.  Sorry that I missed that.

PR #135, opened 2017-05-05
Added support for nested attributes if the type is 'object'

miguelduarte42:
I was having an issue with validating complex, nested Attributes, since their sub-attributes were not saved. I have added a case, similar to `map` and `list`, which can be found in the constructor of the Attribute module.

I ran the tests and it did not seem to have any side-effects.

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -37,6 +37,20 @@ function Attribute(schema, name, value) {
   }
 
   this.attributes = {};
+  
+  if (this.type.name === 'object'){
+
+    for (var objSubattrName in value) {
+      if (this.attributes[objSubattrName]) {
+        throw new errors.SchemaError('Duplicate attribute: ' + objSubattrName + ' in ' + this.name);
+      }
+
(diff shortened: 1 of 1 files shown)


brandongoode:
Thanks for the PR! 

Would you mind adding a test cast that has a nested 'object' type?  That way we can make sure nothing breaks the behavior in the future.

miguelduarte42:
Actually, playing around a bit more with it, I think I fundamentally misunderstood how objects should be stored in DynamoDB. I guess I should have used a Map, and dynamoose already accounts for Maps (and even Lists). I'm still making progress, and if I see that this PR does not make any sense, I will close it.

brandongoode:
Thanks for the update

miguelduarte42:
I just confirmed that my changes are unnecessary. The schema of complex JSON objects should be defined using the Map and Array properties, like the DynamoDB documentation specifies. Using a plain "object" like I was suggesting ends up stringifying the contents, which makes it difficult (if not impossible) to query or scan.

Issue #137, opened 2017-05-09
Implement the populate methods from Mongoose.


lucianopf:
# Reason
Since this is heavily inspired at Mongoose I think we should also implement the populate methods inspired by them.

This is their ref: [Official docs](http://mongoosejs.com/docs/populate.html)

Also that feature makes coding way faster without the need of implementing consecutive queries to get the full result.

# Proposal
Since I'm a new contributor I don't understand every aspect of this library yet, so I'll purpose small and incremental changes until we have a very similar structure of mongoose's populate method. 

By that I'll list goals here so we can keep track and if anyone else wants to get involved follow up the development.

- [x] Implement the populate method
- [x] Attach the method to the Model definition
- [x] Create tests to test this initial feature
- [x] Implement the ObjectId + ref options in an attribute
- [x] Implement a better API, such as `Dog.get(1).populate('father')`
- [x] Implement a better API, such as `Dog.get(1).populate('father.father')`
- [ ] Improve the documentation

brandongoode:
Sounds great!

Thanks for breaking the change up into smaller parts.  It makes it much easier to review and merge.

chops318:
This will be a great addition to dynamoose. What's the status on populate, if you'd like help, I'd love to try and implement. Thanks for the work!

lucianopf:
Yay, glad you like the proposal @chops318 ! 😁 

Go for it buddy, I'm kindda stuck with a lot of work to be done and I might take longer than expected to finish this feature, so, help is always welcome!! =D

troyswanson:
Would this eliminate the need for nested queries, e.g.:

```javascript
Product.get(id, (err, product) => {
  if (!product) { /* no product found */ }
  else {
    Category.get(id, (err, category) => {
      if (!category) { /* no category found */ }
      else {
        product.category = category
      }
    })
  }
})
```

/ref https://github.com/automategreen/dynamoose/issues/120#issuecomment-293621445

lucianopf:
@troyswanson Yeah yeah, definitely!!

My goal is to make:

`Product.get('id').populate('category')`

troyswanson:
AHH! I love it! Thank you!

chops318:
That's my intention aswell. I'm pretty much going to model the API as close to mongoose as possible

brandongoode:
@chops318 Thanks for offering to help.  I'm looking forward to your PR.

brandongoode:
[...]

PR #138, opened 2017-05-09
#137 - Enhancement/implementing populate

lucianopf:
Issue ref: #137

# Reason
Since this is heavily inspired at Mongoose I think we should also implement the populate methods inspired by them.

This is their ref: [Official docs](http://mongoosejs.com/docs/populate.html)

Also that feature makes coding way faster without the need of implementing consecutive queries to get the full result.

# Proposal
Since I'm a new contributor I don't understand every aspect of this library yet, so I'll purpose small and incremental changes until we have a very similar structure of mongoose's populate method. 

By that I'll list goals here so we can keep track and if anyone else wants to get involved follow up the development.

- [x] Implement the populate method
- [x] Attach the method to the Model definition
- [x] Create tests to test this initial feature
- [ ] Implement the ObjectId + ref options in an attribute
- [ ] Implement a better API, such as `Dog.get(1).populate('father')`
- [ ] Implement a better API, such as `Dog.get(1).populate('father.father')`
- [ ] Improve the documentation

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -7,6 +7,7 @@ var Query = require('./Query');
 var Scan = require('./Scan');
 var errors = require('./errors');
 var reservedKeywords = require('./reserved-keywords');
+var deepSet = require('lodash').set;
 
 //var MAX_BATCH_READ_SIZE   = 100;
 var MAX_BATCH_WRITE_SIZE  = 25;
@@ -77,6 +78,15 @@ Model.compile = function compile (name, schema, options, base) {
       return Model.get(NewModel, key, options, next);
     } catch(err) {
       sendErrorToCallback(err, options, next);
+      return Q.reject(err);
+    }
+  };
+
+  NewModel.populate = function (options, resultObj, fillPath) {
+    try {
+      return Model.populate(options, resultObj, fillPath);
+    } catch(err) {
+      sendErrorToCallback(err, options);
       return Q.reject(err);
     }
   };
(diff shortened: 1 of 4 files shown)


lucianopf:
🤔 

```
nvm install 5 
npm test

143 passing (8s)
Done.
```

Passing with node versions:
- 6.10.3
- 5.12.0
- 4.8.3
- 0.12.18

Something seems wrong at Travis tests

brandongoode:
Just a thought that I wanted to capture before I forgot -- Need to check to make sure it works for a Schema with hashkey.  I think it will, but want to double check.

brandongoode:
One thing I would like to change is the need to manually prepend the model name with the prefix.

lib/Model.js
```diff
@@ -398,6 +406,40 @@ Model.get = function(NewModel, key, options, next) {
...
+  var deferred = Q.defer();
+  var self = this;
+  var Models = this.$__.table.base.models;
+  if (!resultObj) {
+    resultObj = self;
+  }
+  if (!options.path || !options.model) {
+    throw new Error('Invalid parameters provided to the populate method');
+  }
+  try {
```
brandongoode: Loop through models and compare `Model.$__.name === Model.$__.options.prefix + options.model` -- or somethink like that

lucianopf: Awesome!!!

That'll avoid manually prefixing such as `'test-Dog6'`

I'll do it on the weekend

lucianopf:
Now it'll accept both the name of the Model or it's Prefix-ModelName

brandongoode:
Looks great.  Thanks

Issue #93, opened 2016-12-31
New feature: scan.all()


brandongoode:
Add support for scan to recusivly scan while `lastKey` exists.

```js
scan.all([delay[, max]])
```

`delay` is the time between recursive scans.  Default: 1sec

`max` is the maximum number of recursive scans. Default: 0 - unlimited

strygin:
Why do we need delay? ;-\

brandongoode:
The thought behind the delay is to allow the request to be throttled.

fishcharlie:
@brandongoode Close this issue? 😃 👍

PR #140, opened 2017-05-14
#93 - Adding support for scan.all()

fishcharlie:
This pull request attempts to add support for scan.all() as described in #93. I tried to integrate this with as minimal code changes as possible but there might be a better way to integrate this more cleanly and with less disruption to the code base.

Feel free to discuss and let me know if there is anything else I can do to get this merged in.


brandongoode:
Thanks for the PR.  It's a great addition.  Give me a few days to look it over.

fishcharlie:
Thanks Brandon. Let me know if I can make any adjustments or updates.

I'm not sure why the tests are failing. I ran the tests locally and they all passed.

I also just realized if for whatever reason options.all gets overwritten to null that will cause problems.

I'm currently on my phone but in a little bit I'll try to look into both isssues and repush if I'm able to come up with fixes.

fishcharlie:
@brandongoode That should fix the undefined issue. Still not sure why tests are failing. Any insight into this would be great. Let me know what other changes or adjustments you need from me when you have a chance to look it over.

brandongoode:
@fishcharlie,  It looks good.  The failure was just a timing issue that happens with TravisCI.  I reran the test, and it's passing now.

Could you add the feature to the Scan section in docs/index.md and some test cases?

fishcharlie:
@brandongoode Just added the documentation. Tried to add 2 unit tests. I'm not the best/most experienced with unit tests. So not sure if they are good tests or anything like that. If someone else wants to write better unit tests feel free since I'm not the best at that haha.

brandongoode:
Thanks.  I'll pull it down and add a few more checks to the test cases.

PR #145, opened 2017-05-23
added schema parsing overrides

gtg092x:
Added per field dynamo parsing overrides in addition to schema-wide parsing overrides.

The per field overrides are a convenience, but the schema-wide overrides can help users manage unknown properties.

From the Docs:

```js
var schema = new Schema({...}, {
  attributeFromDynamo: function(name, json, defaultParser) {
    switch(name) {
        case 'specialAttribute':
            return specialParser(json);
        default:
            return defaultParser(json);
    }
  }
});
```
--- or ---

```js
var schema = new Schema({...}, {
  attributeFromDynamo: function(name, json, defaultParser) {
    if (appearsToBeASpecialType(json)) {
      return specialParser(json);
    }
    return defaultParser(json);
  }
});
```

### DIFF
--- a/docs/_docs/schema.md
+++ b/docs/_docs/schema.md
@@ -143,6 +143,14 @@ Adds a setter function that will be used to transform the value before writing t
 
 Adds a getter function that will be used to transform the value returned from the DB.
 
+**toDynamo**: function
+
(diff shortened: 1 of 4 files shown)


brandongoode:
Thanks -- Nice additional options.

gtg092x:
Thank you!

PR #147, opened 2017-05-24
Schema Method Documentation

lengk:
Added some documentation for the Schema methods including examples about how to use it


brandongoode:
Good catch -- Thanks.

PR #148, opened 2017-05-24
Found typo in Static Methods docs

troyswanson:
s/commpiled/compiled

No idea why there is a diff for line 326...


brandongoode:
Thanks

Issue #141, opened 2017-05-15
how to achieve nested filters query?


anjnaharsora:
Friends.scan({
        or: [
            {and: [{userRef: {eq: {id: param.user}}, friend: {eq: {id: param.friend}}}]},
            {and:[{userRef: {eq: {id: param.friend}}, friend: {eq: {id: param.user}}}]}
        ]
    }
Example.

brandongoode:
Currently, this isn't supported in Dynamooose.  While DynamoDB does support this logic, it requires the use of a FilterExpression:

http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html#DDB-Scan-request-FilterExpression

Dynamoose uses the legacy ScanFilter which doesn't support nesting:

http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html#DDB-Scan-request-ScanFilter


This is something I would like to implement int he future, but haven't had the time to do.

See #65

anjnaharsora:
thanks @brandongoode

lrcry:
Hi all

I think it would be a nice idea for now that dynamoose allows raw filter in scan i.e. `Model.scan(filter, options, callback)` where a new option `useRawAwsFilter` could be added, so that we could use what is currently supported in aws-sdk.

```
const filter = {
  TableName: 'TestTable',
  FilterExpression: 'test.title = :title',
  ExpressionAttributeValues: {
    ':title': 'dynamotest'
  }
};
TestTable.scan(filter,{ 'useRawAwsFilter': true }).exec().then(...).catch(...)
```

I would like to try to fork one to see how it goes.

Regards

Hans

brandongoode:
@lrcry That great.  I was thinking about this some over the past week and wanted to share my thoughts.  

I don't think the `useRawAwsFilter` flag is needed. We could just assume if the `filterExpression` option is present then it should take priority.  We need to make sure to account for attribute names.  Also, in addition to passing in a filter option, I think having the scan building function too would be great.

Something like this:

```js
GameResults
  .scan()
  .filterExpression('#s > :score')
  .expressionAttributeValues({ score : 100 })
  .expressionAttributeNames({ s : 'score'})
```

or

```js
GameResults
  .scan({
    filterExpression: '#s > :score',
    expressionAttributeValues: { score : 100 },
    expressionAttributeNames: { s : 'score'}
  })
```

lrcry:
[...]

PR #149, opened 2017-05-25
Nested scan

lrcry:
For [#141](https://github.com/automategreen/dynamoose/issues/141)

Option for scan to use raw AWS filter or not.

If the option is true, use the filter as it is, and use the DocumentClient from aws-sdk.

Otherwise, do it in the default way.


lrcry:
Will test and fix, close at the moment

PR #158, opened 2017-06-04
Use AWS SDK DynamoDB Document Client to support nested scan

lrcry:
Finally I've got one workaround for now to support nested scan via aws-sdk dynamodb document client...

To use it (as a promise), just compose your own filter ~~~and specify the option {useRawAwsFilter: true}~~~ when scanning:
```javascript
    var Dog = dynamoose.model('Dog');
    // this should be composed, ref: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html
    var filter = {
      FilterExpression: 'details.timeWakeUp = :wakeUp',
      ExpressionAttributeValues: {
        ':wakeUp': '8am'
      }
    };

    Dog.scan(filter).exec()
      .then(function(dogs) {
        console.log(dogs);
      })
      .catch(function(err) {
        console.error(err);
      });
```
It is not necessary to provide `TableName` in the filter as the library will use the model name as table name by default, if a `TableName` is not explicitly specified.
I didn't touch anything which used the default `DynamoDB` client. I think it would be a good idea to modify dynamoose to use the `DocumentClient` to support more features, although it means an amount of work. :)

Limitations:
* When defining schemas, please use `{ useDocumentTypes: true }` if you would like to use nested scan as this is only supported for documents (I think it is a feature of DocumentClient, I googled that before but I could not remember the link, sorry for that)
* It seems that aws-sdk document client has not supported usage of attribute name placeholders yet (when I tried to use the plain aws-sdk in my code, it seemed never to work), which means that something as follows (`#timeWakeUp`) will get error:
```json
{
  "FilterExpression": "#timeWakeUp = :wakeUp",
  "ExpressionAttributeNames": {
    "#timeWakeUp": "details.timeWakeUp"
  },
  "ExpressionAttributeValues": {
    ":wakeUp": "8am"
  }
}
```
Instead, use the name directly (`details.timeWakeUp`):
```json
{
  "FilterExpression": "details.timeWakeUp = :wakeUp",
  "ExpressionAttributeValues": {
    ":wakeUp": "8am"
  }
}
```

### DIFF
--- a/lib/Scan.js
+++ b/lib/Scan.js
@@ -21,13 +21,12 @@ function Scan (Model, filter, options) {
   this.filters = {};
   this.buildState = false;
   this.validationError = null;
-
   if (typeof filter === 'string') {
     this.buildState = filter;
     this.filters[filter] = {name: filter};
   } else if (typeof filter === 'object'){
-    if (options && options.useRawAwsFilter) {
-      // if use raw, just assign the filter
+    if (typeof filter.FilterExpression === 'string') {
+      // if filter expression is given, just assign the filter
       this.filters = filter;
     } else {
       this.parseFilterObject(filter);
@@ -52,14 +51,10 @@ Scan.prototype.exec = function (next) {
   var options = this.options;
 
   var scanReq = { };
-  var dbClient = null;
 
-/**/
     function scanByRawFilter() {
       var deferred = Q.defer();
-      var AWS = Model.$__.base.AWS;
-      AWS.config.update({ endpoint: 'http://localhost:8000' });
-      dbClient = new AWS.DynamoDB.DocumentClient();
+      var dbClient = Model.$__.base.documentClient();
       dbClient.scan(scanReq, function(err, data) {
         if (err) {
           return deferred.reject(err);
@@ -74,7 +69,7 @@ Scan.prototype.exec = function (next) {
       return deferred.promise.nodeify(next);
     }
 
-  if (options.useRawAwsFilter) {
+  if (this.filters && typeof this.filters.FilterExpression === 'string') {
     // use the raw aws filter, which needs to be composed by the developer
     scanReq = this.filters;
     if (!scanReq) {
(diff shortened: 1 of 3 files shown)


lib/Scan.js
```diff
@@ -26,7 +26,12 @@ function Scan (Model, filter, options) {
     this.buildState = filter;
     this.filters[filter] = {name: filter};
   } else if (typeof filter === 'object'){
-    this.parseFilterObject(filter);
+    if (options && options.useRawAwsFilter) {
```
brandongoode: I think it would be easier for other if we just check for FilterExpression in the filter.

```js
if(type filter.FilterExpression === 'string')
```

lib/Scan.js
```diff
@@ -46,68 +51,111 @@ Scan.prototype.exec = function (next) {
...
-            filterAttr.toDynamo(val, true)
-          );
+  var scanReq = { };
+  var dbClient = null;
+
+/**/
+    function scanByRawFilter() {
+      var deferred = Q.defer();
+      var AWS = Model.$__.base.AWS;
+      AWS.config.update({ endpoint: 'http://localhost:8000' });
```
brandongoode: I think this line should be removed

lib/Scan.js
```diff
@@ -46,68 +51,111 @@ Scan.prototype.exec = function (next) {
...
+            return deferred.resolve([]);
+          }
+          return deferred.resolve(data.Items);
+        }
+      });
+
+      return deferred.promise.nodeify(next);
+    }
+
+  if (options.useRawAwsFilter) {
```
brandongoode:
```js
if(this.filter && type this.filter.FilterExpression === 'string')
```

lib/Scan.js
lrcry: Thanks. Let me check this to see what is going on at this point. I remembered I've already moved the AWS things into the model to keep consistency...

lrcry:
@brandongoode  Can you please review my latest change as per the last review you've posted? Thanks a lot.

brandongoode:
Looks great! Could you add a section in the Scan documentation and I'll merge it: https://github.com/automategreen/dynamoose/blob/master/docs/_docs/scan.md

lrcry:
@brandongoode 
I've done the document update. Please view it at #167

PR #161, opened 2017-06-08
Add catch on the waitForActive

asue:
Do the reject on the promise if an exception is thrown from the waitForActive function


brandongoode:
Thanks for helping to improve the code.

Issue #153, opened 2017-06-01
Release 0.8 Planning


brandongoode:
It's been a while since the last release, so 0.8 will have a lot of updates.

- [x] Finalaize new features for 0.8 
- [x] Create change log
- [x] Update depencancies
- [x] Update version and publish

## Change Log Notes (let me know if something is missing)

- useNativeBooleans #55 
- saveUnknown #125 
- Support for multiple indexes defined on the hashkey attribute of the table 
- scan.all() #93 #140
- scan.parallel d7f7f77
- TTL support 92994f1 
- added schema parsing overrides #145
- populate #137 
- Added consistent() to scan.  #15 #142
- Default function enhancements #127
- Create required attributes on update #96
- Add typescript typings #123
- Added .count() for Query and Scan #101

fishcharlie:
Just #61, #81, and #91 to go?

brandongoode:
@fishcharlie #61, #81, and #91 **may** get dropped from 0.8.  It really depends on if someone implements them before #141 & #137.  I going to try and get 0.8 published in the next week or so.

fishcharlie:
@brandongoode Looking forward to it. Looks like there is a lot of great features in this update.

fishcharlie:
@brandongoode Any updates on this?

brandongoode:
@fishcharlie I've had a few high priority things that consumed 100% of my time over the past couple weeks.  I'm going to try to get this done this week.

lrcry:
Just wondering if nested scan is with v0.8 as well? #141

brandongoode:
@lrcry  Yes nested scan will be in it -- I'll add it to the change log

PR #169, opened 2017-06-19
Update Typescript definition (typings)

mooyoul:
HI!
Recently, We've updated our typescript definitions to follow changes that described on #153.
This is the result of our job.

# What's changed on typescript definitions
- Added bunch of new feature definitions (e.g. `populate`, `count`, Global Secondary Index support, TTL support)
- Fixed wrong definition. return value of `Model#queryOne` was defined as wrong type (`Model[]`), queryOne should be return single model, not array. now `Model#queryOne` should return single model value.

Cheers 🍻


brandongoode:
Great! Thanks for keeping the definitions updated.

PR #171, opened 2017-06-26
update typescript definition file

mooyoul:
Recently, We had to access AWS configuration via `dynamoose.AWS.config`. 
but there's no type definition of that re-exported `aws-sdk` module.

so. we added type support of `aws-sdk` module.

Please see attached screenshots :)

Cheers.


brandongoode:
Great - Thanks

PR #174, opened 2017-07-01
Cast model values to their respectable schema types

jutaz:
This adds support for proper typecasting of the models. Currently, if `dynamoose.Model` receives an `Object`, it does not try to validate and parse it, which, at times, can cause model to not strictly adhere to types. Take a look at this example:
```js
import dynamoose, {Schema} from 'dynamoose';

dynamoose.AWS.config.update({
  region: 'us-east-1'
});
// Local Dynamo instance
dynamoose.local('http://localhost:4569');

const bugSchema = new Schema({
  id: {
    type: Number,
    required: true
  },
  message: {
    type: String,
    required: true
  }
});

const bugModel = dynamoose.model('Bug', bugSchema);

bugModel.create({
  // id is a String.
  id: '1234',
  message: 'This is a bug',
}).then((model) => {
  // Logs: string
  console.log(typeof model.id);
});
```

In the above example dynamoose would not do typecasting, that it is able to do, even before sending data to DynamoDB. This causes type problems, because we expect that `model.id` to be a `Number`, not `String`. While in JS-land that's not a big deal, it might be in others.

While I'm not overly familiar with the codebase, I do think that this patch should be enough to make dynamoose work 👌 and properly typecast.

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -60,6 +60,7 @@ Model.compile = function compile (name, schema, options, base) {
   function NewModel (obj) {
     Model.call(this, obj);
     applyVirtuals(this, schema);
+    schema.parseDynamo(this, obj);
   }
 
   util.inherits(NewModel, Model);
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -60,7 +60,7 @@ Model.compile = function compile (name, schema, options, base) {
   function NewModel (obj) {
     Model.call(this, obj);
     applyVirtuals(this, schema);
-    schema.parseDynamo(this, obj);
+    schema.parseDynamo(this, schema.toDynamo(this));
   }
 
   util.inherits(NewModel, Model);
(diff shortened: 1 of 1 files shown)


jutaz:
Well, I should have ran local tests before submitting the PR.

However I'm noticing, that this approach seems to upset dynamoose - in some cases, if part of the model is missing, it will error out.

What would be the guidance on implementing this? I can do the heavy-lifting, but would love some guidance on best approach here.

lib/Model.js
```diff
@@ -60,6 +60,7 @@ Model.compile = function compile (name, schema, options, base) {
   function NewModel (obj) {
     Model.call(this, obj);
     applyVirtuals(this, schema);
+    schema.parseDynamo(this, schema.toDynamo(this));
```
brandongoode: The issue is the Object constructor is used in a lot of places and may not be a full object.  The partial objects passed to  `schema.toDynamo` cause the errors.

During a save (or create), the object is already passed to `schema.toDynamo`.  What needs to happen is that Dynamo Object needs to be parsed.  

Try adding `schema.parseDynamo(this, item.Item)` around line 278.

jutaz: Thanks for this! it seems that this does indeed fix the bug

PR #175, opened 2017-07-03
Add proper support for 'CONTAINS' filter for lists on scans

zookatron:
[DynamoDB fully supports running 'contain' filters on Lists](http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Condition.html), but currently trying to perform a 'contains' filter on a list in a scan with Dynamoose improperly gives the error "'Values must be array in a `list`", due to it being converted to the Dynamo value with the parent list Attribute instead of the list sub-attribute. 

Eg. `Model.scan({listAttribute: {contains: 'value'}}).exec()` fails when it should succeed by returning all results that have a 'value' entry in the listAttribute List.

### DIFF
--- a/lib/Scan.js
+++ b/lib/Scan.js
@@ -98,11 +98,14 @@ Scan.prototype.exec = function (next) {
           ComparisonOperator: filter.comparison
         };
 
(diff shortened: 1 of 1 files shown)


brandongoode:
Thanks for the fix.

PR #179, opened 2017-07-17
Fix usage of next callback without checking for existence

zookatron:
At the top of the `exec()` function, it checks for the existence of the `next` callback, and returns a promise if it does not exist:
```
  if (this.validationError) {
    if (next) {
      next(this.validationError);
    }
    return Q.reject(this.validationError);
  }
```
However at the bottom, if the waitForActive option is set, it does not perform this check and simply calls `next()`, which causes the function to throw an error when waitForActive is set and `next` is not provided. This commit fixes this issue, so it will always check and return a promise if `next` is not set.


brandongoode:
Thanks for the fix -- sorry for the slow response.

PR #181, opened 2017-07-19
Bugfix - Cast raw scans to models too

jutaz:
This is a simple fix to fix a case such as:

```js
import Model from '../../models/models';

var filters = {
  FilterExpression: 'attr = :attr',
  ExpressionAttributeValues: {
    ':attr': attr
  }
};

return Model
  .scan(filters)
  .exec()
  .then((results) => {
    return results[0]; // Plain Object
  });
```

Since we are still performing a scan on a known table of a known model, it makes sense to also cast all of the objects to that model.

### DIFF
--- a/lib/Scan.js
+++ b/lib/Scan.js
@@ -52,13 +52,19 @@ Scan.prototype.exec = function (next) {
 
   var scanReq = { };
 
-  function toModel (item) {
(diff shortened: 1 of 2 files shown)


brandongoode:
Thanks for the fix -- sorry for the slow response

PR #185, opened 2017-07-27
Add validation for callback before call it

jibrang:
## What does this PR do?
- Add a validation to see if callback exists before call it, because when you call ```scan.get()``` without a callback and the server is not retrieving data, it crashes.

## Where should the reviewer start?
:checkered_flag:
lib/Scan.js

## Screenshots
```TypeError: next is not a function```
```node_modules/dynamoose/lib/Scan.js:290:7```


brandongoode:
Thanks for the fix.  This was also fixed by #179 which has been merged.

PR #187, opened 2017-08-03
Fixed type definition of SchemaOptions

f-barth:
Made `attributeFromDynamo` and `attributeToDynamo` optional in type definition of `SchemaOptions`. In current state api consumer would need to implement these operations in order to set scalar schema options.


brandongoode:
Thanks.

PR #194, opened 2017-08-21
Remove underscore in favour of lodash

ro-savage:
I noticed dynamoose was using both `lodash` and `underscore`. So I removed `underscore`, and now dynamoose only uses lodash.

One less dependancy.


brandongoode:
Thanks -- Sorry for taking so long to respond

PR #200, opened 2017-09-02
Bug fix: In Model.put() / Model.save(), the returned promise is not rejected on error

AndreySoftware:
For example, if doing a conditional save that fails, the returned promise object is resolved instead of being rejected.
This happens because the following code returns the wrong promise:
if(model$.options.waitForActive) { return model$.table.waitForActive().then(putItem).catch(deferred.reject); }


brandongoode:
Great! Thanks for the fix.  (and sorry for the delay in merging).

PR #201, opened 2017-09-04
Query - Add ability to use contains on a list

ro-savage:
DynamoDB supports `contains` when using a `query` on a list.

This adds that functionality. It is the same code as can be found in `Scan` + a new test. 

Thanks @zookatron.


brandongoode:
Thanks -- sorry for the delay

PR #202, opened 2017-09-07
Enable http keep alive functionality

zookatron:
In our use case (AWS lambda functions) this improves the time it takes to make a typical DynamoDB request from 100-200ms to 10-20ms. It's possible this should be a configuration setting or something instead of always enabled, but I wanted to submit this first and get feedback on whether that is necessary and how it should be done.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,6 +2,7 @@
 
 var Schema = require('./Schema');
 var Model = require('./Model');
+var https = require('https');
 
 var debug = require('debug')('dynamoose');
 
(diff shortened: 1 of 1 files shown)


lrcry:
This sounds like a good idea if it really helps reduce the time, but IMO this would be better if we just set it if needed instead of enforcing it in Dynamoose, or at least provide an option to toggle it.

While when I do local development, I would not wish to have Dynamoose HTTPS keep-alive enabled most of the time as I might connect to a local running DynamoDB container, which would probably use HTTP.

E.g. if I wish to use this feature, I could just parse this as DynamoDB configs when initialising Dynamoose / defining a Dynamoose model.

BTW I think it is a better idea to add this setting to the documents FYI, as perhaps some of the users (e.g. me) do not even know about this setting...

Just would like to know what @brandongoode thinks on this.

lib/index.js
```diff
@@ -78,10 +79,25 @@ Dynamoose.prototype.ddb = function () {
   }
   if(this.endpointURL) {
     debug('Setting DynamoDB to %s', this.endpointURL);
-    this.dynamoDB = new this.AWS.DynamoDB({ endpoint: new this.AWS.Endpoint(this.endpointURL) });
+    this.dynamoDB = new this.AWS.DynamoDB({
```
brandongoode: Don't default this to https (local needs to be http)

brandongoode:
I like the change. 

In general, I like to lean towards making the default behavior to be the "best" configuration for a none DynamoDB expert.  However, for anyone that passes in any manual settings, it should default back to the AWS default (i.e., for local testing).  By only defaulting to the https keep alive for the default endpoints (which are https), the update will preserve backward compatibility.

We also need to document how to get back to the AWS default.

brandongoode:
btw -- sorry for the slow response

ro-savage:
@zookatron - Can you explain a little how this works? I thought because each time Lambda fired it was a 'new' server, so it couldn't be kept alive?

(Also, 200ms seems very long. None of my queries take over 100ms for the lambda to run and finish, often taken closer to 50ms)

softmarshmallow:
Question.: Is keep alive automatically being enabled if possible?

fishcharlie:
@softmarshmallow No. This code was part of v1. You can implement something similar yourself in v2. However it is not the default.

PR #203, opened 2017-09-09
Updated docs to show conditionals for Model.update()

ro-savage:
The docs were missing the fact that conditionals can be used on `Model.update()`.

Updated the docs with the same opts that were in `Model.put()`


brandongoode:
Thanks again

PR #213, opened 2017-10-10
AWS dynamodb expects milliseconds, not seconds for TTL/Expires.

creo-verus:
Should be the same format as the timestamps in the preceding lines.


creo-verus:
Reference: http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/howitworks-ttl.html

creo-verus:
I was so totally incorrect here.

Issue #197, opened 2017-08-24
In Model.put() / Model.save(), the returned promise is not rejected on error


AndreySoftware:
For example, if doing a conditional save that fails, the returned promise object is resolved instead of being rejected.
This happens because the following code returns the wrong promise:
`
if(model$.options.waitForActive) {
      return model$.table.waitForActive().then(putItem).catch(deferred.reject);
    }`

Suggested fix:
return `deferred.promise.nodeify(next)` instead.

brandongoode:
Would you be willing to correct it in a PR?

lennym:
I've just hit this same issue with `Model.create`. I'll put together a PR.

lennym:
Ugh, just spent the last hour trying to recreate the issue with a failing test before I spotted that this has already been fixed at #200 :-(

Any chance of publishing a new version?

jkrusinski:
A new version for this bug would be of great help for me as well!

brandongoode:
@lennym  Sorry for not publishing an updated release and sorry for not getting back any sooner.  I've been slammed lately and some of my sided projects (i.e. this one have suffered).  I'll try to push out an updated version this week.

brandongoode:
The fix is in 0.8.6

PR #215, opened 2017-10-19
Enable unit test code coverage reporting

lovell:
Hi Brandon, I was bitten by #197 and noticed that the accepted fix in #200 did not include a test that could have caught the problem.

I tried to generate the unit test code coverage report for this module with `grunt coverage` but received a `"html-cov" reporter not found` error.

It appears mocha dropped support for `html-cov` reporting in v3.0.0.

Given this module does not appear to require any of the features of Mocha v3, the quickest way to generate a coverage report was to downgrade Mocha to the most recent v2.x version.

With the small change in this PR it is now possible to use `npm run coverage` (or `yarn coverage`) to generate full code coverage reports.

Whilst "downgrading" Mocha may feel like a step backwards, the ability to generate a code coverage report provides a big step forwards in terms of confidence-building in this useful module.

If there's interest, I could take a look at modernising the test tooling to use something like `nyc`, which provides more useful statistics around branch coverage.

Cheers,
Lovell

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
   },
   "main": "index.js",
   "scripts": {
-    "test": "grunt test"
+    "test": "grunt test",
+    "coverage": "grunt coverage"
   },
   "keywords": [
     "dynamodb",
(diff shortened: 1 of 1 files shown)


brandongoode:
@lovell  -- Thanks for the offer to modernize the testing.  I would gladly take you up on the offer.  One of my other repos has already been moved over to istanbul and uses https://coveralls.io for coverage tracking.

PR #217, opened 2017-10-23
Bugfix - Provide model when converting keys .toDynamo.

jutaz:
This allows schema to run validation and defaults. Take the following schema for example:
```js
new Schema({
  id: {
    type: String,
    index: {
      global: true
    },
    required: true,
    default: model => `${model.properties.ownerType}_${model.properties.ownerId}`,
    validate: (value, model) => value === `${model.properties.ownerType}_${model.properties.ownerId}`
  },
  properties: {
    type: 'map',
    map: {
      ownerType: {
        type: String,
        validate: value => ownerTypes.includes(value),
        required: true
      },
      ownerId: {
        type: Number,
        validate: value => typeof value === 'number',
        required: true
      },
      organizationId: {
        type: Number,
        validate: value => typeof value === 'number',
        required: true
      }
    }
  }
});
```

Now suppose I wanted to delete a model of that schema by doing the following:
```js
AModel.delete();
```

There would be a JS exception raised due to `model` being `undefined` in `validate` method, because `schema.rangeKey.toDynamo` and/or `schema.hashKey.toDynamo` were not given the actual model as a context.

This PR fixes this particular edge case. However, I noticed that there are quite a few occurrences where `.toDynamo` is being called without providing a model - should those callsites also provide a model?

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -851,11 +851,11 @@ Model.prototype.delete = function(options, next) {
   };
 
   try {
-    getDelete.Key[hashKeyName] = schema.hashKey.toDynamo(this[hashKeyName]);
+    getDelete.Key[hashKeyName] = schema.hashKey.toDynamo(this[hashKeyName], undefined, this);
 
     if(schema.rangeKey) {
       var rangeKeyName = schema.rangeKey.name;
-      getDelete.Key[rangeKeyName] = schema.rangeKey.toDynamo(this[rangeKeyName]);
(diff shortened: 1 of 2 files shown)


brandongoode:
Good catch.  It may need to be updated in the others too.  Originally the validate didn't get passed the model so it wasn't needed, but it now may be needed for all calls.

Would you mind adding your example to the test cases to make sure it's not missed with any future changes?

jutaz:
> Would you mind adding your example to the test cases to make sure it's not missed with any future changes?

yup, will sure do!

jutaz:
Updated with tests

brandongoode:
Thanks again for the fix and test case

PR #219, opened 2017-10-24
Switch to nyc (istanbul) for unit test coverage

lovell:
Hi Brandon, this is as discussed in #215

This PR uses current percentages as bare minimum threshold values, which can gradually be increased in line with coverage.

It also bumps mocha back to latest version that still provides support for Node 0.12.


brandongoode:
Thanks for the help.

PR #223, opened 2017-10-31
Adding support for query.all()

takehilo:
This pull request adds support for query.all(). It works similarly to the scan.all().


brandongoode:
Thanks for the additional functionality

fishcharlie:
@brandongoode Any ideas when this will make it into a release?

PR #225, opened 2017-11-01
Fix range key on queries

zookatron:
Currently range keys don't work on Queries when constructed all at once via the constructor because it uses the `value` key, while the `exec` function looks for the `values` key. This is the simplest possible commit that allows range keys to work via the Query constructor, though I'm not sure if it's the best solution, please let me know if there's a better way of fixing this issue.


brandongoode:
Looks good to me.  Thanks

Issue #226, opened 2017-11-01
Cannot query with updatedAt as range key


nbarnett:
Context: I wish to query a global secondary index using `updatedAt` as the range key. 

Example:
```
import dynamoose from 'dynamoose';

// dynamoose config here

const Schema = dynamoose.Schema;

// Create the model
const Record = dynamoose.model('Record', new Schema({
  recordID: {
    type: String,
    hashKey: true
  },
  tableID: {
    type: String,
    required: true,
    index: {
      global: true,
      rangeKey: 'updatedAt'
    }
  }
}, {
  timestamps: true
}));

// Add some data to the table (in seperate requests so that the updated time is different)
new Record({
  recordID: 'ca4cba8c-d7e7-4845-b554-cb8ee6859642',
  tableID: '78d650c5-9b7f-404e-9873-03068f06b51d'
}).save();

new Record({
  recordID: '534d980b-3cf5-455d-9a79-b31e68db4bca',
  tableID: '78d650c5-9b7f-404e-9873-03068f06b51d'
}).save();

new Record({
  recordID: 'c00a9bfb-3450-4356-b94e-94c02887e649',
  tableID: '78d650c5-9b7f-404e-9873-03068f06b51d'
}).save();

new Record({
  recordID: '1a2415c6-d3cc-412d-a945-bb1ef398c7f7',
  tableID: '78d650c5-9b7f-404e-9873-03068f06b51d'
}).save();

new Record({
  recordID: 'd2d1512f-627e-4199-9eb2-2ad52ebe07fd',
  tableID: '78d650c5-9b7f-404e-9873-03068f06b51d'
}).save();

// Query the GSI
const startTime = 1509514835544; // Pick a time somewhere in the middle to see if it works.

Record.query('tableID').eq('78d650c5-9b7f-404e-9873-03068f06b51d')
  .where('updatedAt').ge(startTime)
  .exec((err, res) => {
    console.log(res);
  });
```
In this example `ge` and `gt` return no results, while `lt` and `le` just return all of the records.

There is no issue doing this with `createdAt` (to test, replace `updatedAt` with `createdAt` in the schema and query) or any other range keys, just with `updatedAt`.

nbarnett:
So I did some digging and think I've found the source of the problem. In `Schema.js` line 59, `updatedAt` is defined with a setter:
```
obj[updatedAt] = { type: Date, default: Date.now, set: function() { return Date.now(); } };
```
[...]

PR #228, opened 2017-11-03
Fix issue where querying on updatedAt rangeKey was broken

nbarnett:
The issue that this fixes is discussed in https://github.com/automategreen/dynamoose/issues/226

This change prevents the `updatedAt` setter from overwriting user input, allowing queries on a secondary index that uses `updatedAt` for the rangeKey.

### DIFF
--- a/lib/Schema.js
+++ b/lib/Schema.js
@@ -56,7 +56,13 @@ function Schema(obj, options) {
     }
 
     obj[createdAt] = { type: Date, default: Date.now };
-    obj[updatedAt] = { type: Date, default: Date.now, set: function() { return Date.now(); } };
+    obj[updatedAt] = { type: Date, default: Date.now, set: function(val) {
+      if (val === undefined) {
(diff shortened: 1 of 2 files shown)


nbarnett:
As mentioned in https://github.com/automategreen/dynamoose/issues/226, I realised that this fix didn't work for the intended functionality of updating `updatedAt`. I have reverted that change and written two unit tests: one to ensure that the `updatedAt` field actually updates, and the other (which currently fails) ensures that querying on an `updatedAt` range key works.

brandongoode:
@nbarnett Thanks for the additional test cases.  It appears the test case only fails on node version 0.12.  What version of node are you using locally?

brandongoode:
Also - sorry for overlooking this PR and issue #226 for so long.

fishcharlie:
@nbarnett You still interested in getting this in?

fishcharlie:

PR #240, opened 2017-11-21
Fix to allow unknown types to be populated as native dynamo types

stang-tgs:
Fix to allow unknown types to be populated as native dynamo types when useDocumentTypes=true

I separated out the basic `Attribute.setType` from `Attribute.setTypeFromRawValue` instead of conditionals inside `Attribute.setType` to clearly differentiate when what should be called and the purpose of each method.  I've also commented out a piece of code in `Attribute.js` since it would break the toDynamo on a List (unknown Attribute).  I'm not sure what the guard is there for (possible for defined schema so there are no heterogenous lists?).  I will leave it at your discretion to modify.

I've also added a Test for this particular case.  All tests still pass.

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -16,8 +16,11 @@ function Attribute(schema, name, value) {
   this.schema = schema;
 
   this.name = name;
-
-  this.setType(value);
+  if (this.schema.options.saveUnknown) {
+    this.setTypeFromRawValue(value);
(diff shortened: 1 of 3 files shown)


brandongoode:
Looks good.  Thanks for the fix and additional test cases.

I 'think' the guard was to prevent multiple types in a List but it really doesn't add much value.

PR #245, opened 2017-11-25
Fixed types

koladilip:
- Made option fields optional as JS code handles them when there are not present.
- Added overloaded create function to allow passing options which is missed out.


brandongoode:
Thanks for the update.

PR #254, opened 2017-12-07
Refactoring Errors File

tylerbrownhenry:
Just noticed a bunch of redundant code, and out of habit refactored it. Love this project.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const util = require('util');
+var util = require('util');
 
 function ErrorType(defaultMessage,errorName){
   let newError = (message)=>{
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -1,59 +1,24 @@
 'use strict';
 
-var util = require('util');
+const util = require('util');
 
-function SchemaError(message) {
-  Error.call(this); //super constructor
-  Error.captureStackTrace(this, this.constructor); //super helper method to include stack trace in error object
+function ErrorType(defaultMessage,errorName){
+  let newError = (message)=>{
 
-  this.name = this.constructor.name; //set our function’s name as error name.
-  this.message = message || 'Error with schema';
+    Error.call(this); //super constructor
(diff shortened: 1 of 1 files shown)


brandongoode:
@tylerbrownhenry - thanks for helping to clean up the code.  Could you change the `const` back to `var`.  We're still supporting node 0.12 (that will change in the near future).  

Also -- Sorry for the slow response -- took some time off.

tylerbrownhenry:
Changed.

brandongoode:
Thanks for the quick fix!  

I wasn't paying enough attention. That was the only the first of several ES6 syntax things: 

https://travis-ci.org/automategreen/dynamoose/jobs/324274242#L583

billouboq:
@brandongoode Hi brandon, when will you drop support for node 0.12 ?

lucianopf:
I don't think we need to drop support for 0.12, instead we can transpile that if needed, what do you think @brandongoode ?

billouboq:
@lucianopf version 0.12 is not used a lot actually, even lower than version 5

https://semaphoreci.com/blog/2017/11/22/nodejs-versions-used-in-commercial-projects-in-2017.html

lucianopf:
We can see that in two ways... 😂 
1 - 0.12 is useless and we should drop support
2 - 0.10 is still very useful and we should start supporting, and by supporting 0.10 it makes sense to also support 0.12 

😂 
(I usually don't support lower versions than 4.x, but I understand the goal of it, for example [here)](https://github.com/filipedeschamps/cep-promise)

fishcharlie:
I **believe** (take this with a grain of salt) the current plan is to drop Node 0.12 support with Dynamoose version 1.0.

fishcharlie:
So yes, as of right now the plan is to drop Node.js 0.12 support in version 1.0. We will plan on getting this landed with version 1.0.

fishcharlie:
Closing due to PR #414 resolving this PR.

PR #257, opened 2017-12-08
pass model instance directly to schema::attributeFromDynamo

AlexeyKhristov:


brandongoode:
Thanks - sorry for the slow respone.

PR #258, opened 2017-12-09
Fix typings

hassankhan:
A schema with nested properties would throw errors, despite Dynamoose  allowing for it.

What now works:
```
{
  userId: {
    type: String,
    hashKey: true,
    required: true,
  },
  profile: {
    bio: {
      type: String,
    },
    location: {
      type: String,
    },
  },
  avatar: {
    type: String,
  },
};
```

### DIFF
--- a/dynamoose.d.ts
+++ b/dynamoose.d.ts
@@ -33,6 +33,10 @@ declare module "dynamoose" {
     virtualpath(name: string): any;
   }
 
+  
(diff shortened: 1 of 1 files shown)


hassankhan:
I'm using v0.8.6, I haven't actually checked against `master`

brandongoode:
Thanks and sorry for the slow response.

PR #274, opened 2018-01-15
Feature/add enum attribute

### LINKED ISSUE #239: Support to enum attribute
I'm new to dynamoDB, I'm trying to migrate from mongoDB (mongoose) to dynamoDB (dynamoose).
I came across some difficulties, one of them was the use of enum.
I did not find support in the dynamoose documentation.
Is there an alternative?

```javascript
//Example Schema in mongoose
var breakfastSchema = new Schema({
      drink: {
        type: String,
        enum: ['Coffee', 'Tea']
      }
})
```

billouboq:
Add enum feature like mongoose one with tests and documentation.
Don't hesitate to tell me if there is something to change or if something is missing.

Fixes #239


docs/_docs/schema.md
```diff
@@ -122,6 +126,10 @@ function(model) {
 }
 ```
 
+**Enum: Array of string**
```
fishcharlie: This should probably be `enum` not `Enum`, correct? And probably array of strings.

test/Schema.js
```diff
@@ -835,4 +836,98 @@ describe('Schema tests', function (){
...
+    var Dog = dynamoose.model('Dog' + Date.now(), schema);
+    var oscar = new Dog();
+
+    oscar.name = 'oscar';
+    oscar.weight = 100;
+
+    oscar.save(function(err) {
+      should(err).be.null();
+
+      Dog.$__.table.delete(function () {
```
fishcharlie: What's the reason for the `table.delete` here?

lib/Attribute.js
```diff
@@ -427,6 +427,11 @@ Attribute.prototype.toDynamo = function(val, noSet, model) {
     }
 
     if(type.dynamo === 'S') {
+      if (this.options.enum) {
```
fishcharlie: This looks to be in the `toDynamo` function. Do we think it would be beneficial to add it to the `fromDynamo` function as well? Not sure if it makes sense, what are your thoughts on this?

test/Schema.js
fishcharlie: After looking at this closer this now makes sense.

docs/_docs/schema.md
fishcharlie: Fixed this in my commit to this PR

lib/Attribute.js
fishcharlie: @brandongoode can you take a look at this and let me know your thoughts? Do you think we should add this to `fromDynamo` as well? Should we throw an error in the `fromDynamo` if it's not correct in the enum, or?

PR #277, opened 2018-01-18
Remove JSON.stringify for model in Debug message

koladilip:
This is causing 'Converting circular structure to JSON' error.

If the model is complex enough then this will cause the above error so removed JSON.stringify


fishcharlie:
Thanks for this @koladilip

PR #280, opened 2018-01-24
adds `ddb` interface to typescript module

ralucas:
This adds the `ddb` method to the typescript definitions.


stewartjarod:
Think we can pull this in? Or maybe pull in https://gist.github.com/rvaidya/27017a6495fedc27282e7df9e7727529 which also includes `ddb`

fishcharlie:
Thanks for this @ralucas

PR #289, opened 2018-02-19
Fixing issue where some scan tests weren't being run correctly

fishcharlie:
These tests were not being run due to async and function problems. For example if you changed the error message to something different than what it should be, the test would **never** fail.

I changed it to remove the function part and ensure the test doesn't finish until it's actually done. I also tested to ensure the test would fail when I change the `err.message.should.eql` to something different than what it should be.

### DIFF
--- a/test/Scan.js
+++ b/test/Scan.js
@@ -543,7 +543,7 @@ describe('Scan', function (){
   });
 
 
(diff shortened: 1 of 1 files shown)


fishcharlie:
@brandongoode Any ideas why this test is failing? I didn't touch that section of the code.

PR #306, opened 2018-03-11
Adding support for SSESpecification

### LINKED ISSUE #283: Support for "SSESpecification" Schema option
DynamoDB [recently announced server-side encryption](https://aws.amazon.com/blogs/aws/new-encryption-at-rest-for-dynamodb), and it would be great if we could enable it when creating a table via the Dynamoose Schema options.. or at the very least, supply a way for users to pass-through additional options that aren't parsed by the Dynamoose API.

fishcharlie:
Closes #283

**NOTICE** this change also upgrades the AWS SDK to version `2.207.0`.


fishcharlie:
For some reason the following test case is failing.

```
  it('Should enable server side encryption', function(done) {
    var Model = dynamoose.model('TestTable', { id: Number, name: String }, { serverSideEncryption: true });
    setTimeout(function () {
      Model.$__.table.describe(function(err, data) {
        console.log(err, data);
        data.Table.SSEDescription.Status.should.equal("ENABLED");
        done();
      });
    }, 2000);
  });
```

`Uncaught TypeError: Cannot read property 'Status' of undefined`

```
{ Table: 
   { AttributeDefinitions: [ [Object] ],
     TableName: 'test-TestTable',
     KeySchema: [ [Object] ],
     TableStatus: 'ACTIVE',
     CreationDateTime: 2018-03-11T01:56:05.340Z,
     ProvisionedThroughput: 
      { LastIncreaseDateTime: 1970-01-01T00:00:00.000Z,
        LastDecreaseDateTime: 1970-01-01T00:00:00.000Z,
        NumberOfDecreasesToday: 0,
        ReadCapacityUnits: 1,
        WriteCapacityUnits: 1 },
     TableSizeBytes: 0,
     ItemCount: 0,
     TableArn: 'arn:aws:dynamodb:ddblocal:000000000000:table/test-TestTable' } }
```

Maybe DynamoDB local doesn't return a `SSEDescription` property when describing the table? I'm going to remove this test case for now, but if anyone has any ideas on this I would love to hear some ideas on how to ensure we are testing this properly as we should be trying to increase our code coverage not lower it.

docs/_docs/config.md
```diff
@@ -37,6 +38,16 @@ Basic example:
 var Cat = dynamoose.model('Cat', { id: Number, name: String });
 ```
 
+serverSideEncryption: boolean
+
+Indicates whether server-side encryption is enabled (true) or disabled (false) on the table. This boolean will be passed into the `SSESpecification.Enabled` option property when creating the table. Currently (when feature was implemented) DynamoDB doesn't support updating a table to add or remove server-side encryption, therefor this option will only be respected on creation of table, if table already exists in DynamoDB when using Dynamoose this value will be ignored.
```
ecwyne: Smallest thing. `therefore` is misspelled

khamasaki:
Hi - I've been trying to use this with a recent version of dynamoose from master and it doesn't seem to be working. I think what might have happened is somehow commit 367637c was lost in the merge. It still passes in the table but the buildTableReq function no longer uses it.

Issue #216, opened 2017-10-22
Support for native promises (async/await)


fishcharlie:
It would be awesome to see support for native promises, so that Dynamoose can support the new `async`/`await` syntax. It looks like according to [this post](https://medium.com/@joshuawcomeau/using-async-await-with-traditional-packages-10316d6cc4ee) one option is to wrap existing code in a native promise to make it super modular.

I know there are certain features that 3rd party promise libraries offer compared to native promises, not sure how much Dynamoose uses these features and what the pros and cons would be to implementing this. But the `async`/`await` syntax is super useful and would love to see support with Dynamoose.

brandongoode:
Completely agree.  If you look at my first goal for 1.0 release (https://github.com/automategreen/dynamoose#release-10) it's to update to ES6.  Part of that would include native promises.

That being said, I need to find some spare time (something I'm in short supply of lately) to get around to the 0.9 list first.

lengk:
I've been able to use most functions with the syntax

Aedalus:
Would you be open to a PR if I tried to tackle this? I'd like to try to tackle this to remove the dependency on Q, ideally using native async/await when possible and the native util.promisify otherwise.

fishcharlie:
@Aedalus, @mogwai is correct, `async`/`await` work currently. I'm 100% open to a PR for that tho. I would say to just keep the PR at just removing the Q dependency to start, and using the native util.promisify. I think doing a rewrite to use `async`/`await` internally would be more massive. That being said, I'm open to that PR as well, I just would like to keep PR's as small as possible so they are easier to review. If that makes sense.

Aedalus:
So I believe I'm making good progress. Refactored a few of the `Q.defer/deferred.promise.nodeify(next)`, and all unit tests are still passing. 

However I had one gotcha that I'm still trying to figure out how I want to approach. Node throws a `UnhandledPromiseRejectionWarning` whenever a native promise isn't caught. (Same happens with Bluebird I believe). This results in `console.errors` whenever a rejected promise is created. 

Seems like there are a few ways to handle this, none of which I love.

Option 1: Explicitly add a catch handler for each thrown promise. Way too verbose imo, and seems kinda hacky
```
 function yourExportedFunction() {
     const p = promiseThatMightRejectFn(); 
     p.catch(() => {}); // add an empty catch handler
     return p;
 }
```

[...]

PR #315, opened 2018-03-15
batchPut() not working with async/await keywords

mogwai:
Hey I noticed the `batchPut` method doesn't work with async/await syntax:

I wrote a test for it but the linter wasn't sure how to configure the test runner to run this one in a higher version of node.
I get this back when using:
```js
let results = await Model.batchPut(models)
console.log(results)
//"{"Responses":{},"UnprocessedItems":{}}"

```

### DIFF
--- a/test/Model.js
+++ b/test/Model.js
@@ -1385,6 +1385,30 @@ describe('Model', function (){
         done();
       });
     });
+    
+    it('Use async await to put', function (done) {
+      var cats = [];
+
+      for (var i=0 ; i<100 ; ++i) {
+        cats.push(new Cats.Cat({id: 100+i, name: 'Tom_'+i}));
+      }
+      var asyncPut = function(){
+        done();
+      };
+      var es6code = 
+        'asyncPut = function () {'+
+          'var result = await Cats.Cat.batchPut(cats) '+
+          'should.exist(result)'+
+          'result.length.should.eql(cats.length)'+
+        '}'+
+        'asyncPut()'; 
+      eval(es6code);
+      asyncPut().catch(function(err){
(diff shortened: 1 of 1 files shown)


fishcharlie:
@lengk Looks like this is failing. And I'm not completely sure what the objective of this PR is. Is this to report a problem with Dynamoose by writing a failing test? Or is this just meant to extend the test coverage of Dynamoose?

lengk:
Reporting a problem. The batchPut method doesn't returned the saved models when using async await syntax

fishcharlie:
@lengk Have you gotten `async`/`await` to work with Dynamoose? I haven't played around with it **too** much yet. Have you also seen #216?

fishcharlie:
@lengk Good to know, I thought since Dynamoose is using [`q`](https://www.npmjs.com/package/q) instead of native Javascript promises it wouldn't work. I'll look into this when I get a chance.

lengk:
This test can't be dealt with until the library has been updated, as mentioned in #216, so probably v1.0

fishcharlie:
@lengk Yeah, and we probably should get rid of the `eval`. I'll still look into this bug when I get a chance and see if I can come up with a solution.

mogwai:
Sorry I've realised that I was expecting the `batchPut()` operation to behave the same as `save()`.

When you use `model.save()` you are returned the new parsed value. This is not the case for `batchPut()`

fishcharlie:
@mogwai Is this still something you need me to take a look at?

mogwai:
@fishcharlie No I'm saying I don't think that this pull request is valid.

PR #317, opened 2018-03-17
Adding updateTimestamps option to Model.save

### LINKED ISSUE #227: updatedAt Timestamp Only/Don't Change Under Certain Conditions
It would be awesome to add a feature where the updatedAt property only updates or doesn't update under certain conditions. I have an attribute that often changes and I would like to enable an option for users that when this one attribute changes they can either choose to have it change the updatedAt property or not. Since I sort by the updatedAt property later on this would be super helpful.

Looks like we would have to change the following lines a bit to maybe pass in more data and enable that option.

https://github.com/automategreen/dynamoose/blob/71bf82a1103779337248ddcdf337f6d3eda32fb3/lib/Schema.js#L59
https://github.com/automategreen/dynamoose/blob/dc95809fdd409587f16f9981549d53b71738c02d/lib/Attribute.js#L314

fishcharlie:
Closes #227

### DIFF
--- a/docs/_docs/model.md
+++ b/docs/_docs/model.md
@@ -44,6 +44,10 @@ odie.save({
 });
 ```
 
+Options:
+  - overwrite: should overwrite the existing item in DynamoDB (default: true)
+  - updateTimestamps: should update the updatedAt timestamp if exists (default: true)
+
 ### Model.batchPut(items, options, callback)
 
 Puts multiple items in the table. Will overwrite existing items.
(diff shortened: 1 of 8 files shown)


fishcharlie:
This might be better to implement in plugins #325

fishcharlie:
I actually think this is useful to integrate into Dynamoose itself since timestamps are. If we ever move timestamps into a separate plugin then we can reconsider moving this PR into a separate plugin. But until then, let's move ahead with getting this PR in.

PR #319, opened 2018-03-20
allow update partial data

cncolder:


fishcharlie:
Thanks for this @cncolder

PR #320, opened 2018-03-20
Normalise Scan response shape when using "RawAWSFilter"

DanielBoa:
## Problem
When using a raw AWS filter to Scan the returned Item shape is not consistent with a regular Scan.

Instead of:
```
NewModel {
    name: 'Snoopy',
    details: { timeWakeUp: '8am', timeSleep: '8pm' },
    ownerId: 19,
    color:  ['black', 'white'],
    cartoon: 'true',
    breed: 'Beagle'
}
```
We get:
```
NewModel {
    name: 'Snoopy',
    details: { timeWakeUp: '8am', timeSleep: '8pm' },
    ownerId: 19,
    color: Set { values: ['black', 'white'], type: 'String' },
    cartoon: 'true',
    breed: 'Beagle'
}
```
This is happening because `AWS.DynamoDB.DocumentClient` creates `Set` instances for attributes of type StringSet etc. These aren't native instances of `Set` but rather a custom AWS Class which isn't exposed.

## Solution
When mapping over the returned objects within `Scan.js` I iterate over each items properties and check for such instances to convert them to regular array instances.

### DIFF
--- a/lib/Scan.js
+++ b/lib/Scan.js
@@ -65,7 +65,7 @@ Scan.prototype.exec = function (next) {
     function scanByRawFilter() {
       var deferred = Q.defer();
       var dbClient = Model.$__.base.documentClient();
-      var AWSSet = dbClient.createSet([1, 2, 3]).constructor;
+      var DynamoDBSet = dbClient.createSet([1, 2, 3]).constructor;
 
       dbClient.scan(scanReq, function(err, data) {
         if (err) {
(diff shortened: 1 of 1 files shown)


fishcharlie:
@brandongoode I created an issue #322, to track maybe adding tests for this PR in the future. For now we can approve it and add them later on.

fishcharlie:
Thanks for this @DanielBoa!!

PR #321, opened 2018-03-21
Suffix Option

mogwai:
I've been "overriding" the model method to do this. Thought it might be handy if was a default option. 

I've commented in the Model.js file as I'm not quite sure whether we need to change some code in there too


fishcharlie:
Thanks for this @mogwai, I just made a few changes to add some tests and update the docs to reflect this change. I'm going to try to have @brandongoode take a look at this before merging in. We'll try to get this in for v0.8.7.

Issue #159, opened 2017-06-07
Recording Update History


mogwai:
## Summary
I want the version to show the changes that have happened to an entry, saving previous versions in a new row. Or, these oldVersions could be saved in the object itself if there is already a convention that uses the __v or _version field #16.

When compiling the model: 

```javascript
var Model = dynamoose.model('Models', Model, {versioning:true})
```

This would result in the schema having a __v field which specifies 

```javascript
{
    name: {
        type: String,
        hashKey: true,
    },
    __v: {
        type: Number,
        default: 1
    },
    createdAt: { // Will now be when this update occured
        type: Date,
    },
    updatedAt: { // When this was changed
        type: Date,
    }
}
```

## Requirements

- Queries to the database will automatically get the latest version of an item 
- Putting and Saving items will automatically increment the __version number and create a new item 

## Results
- Will allow developers to implement reverting back to previous versions

troyswanson:
This is a cool idea! 👍 

IMO, storing the previous version within the same object makes the most sense, in case you are using a unique ID as the hash key.

Maybe you could specify which fields need to be versioned? This can be done in the schema instead of the model creation.

```json
{
  "id": "ABC",
  "name": "Version 3 -- The second update",
  "__v": {
    "number": 3,
    "history": [{
      "name": "Version 1 -- The original string",
      "createdAt": "<timestamp_1>"
    }, {
      "name": "Version 2 -- The first update",
      "createdAt": "<timestamp_2>"
    }]
  },
  "createdAt": "<timestamp_1>",
  "updatedAt": "<timestamp_3>"
}
```

Thoughts?

lengk:
> Maybe you could specify which fields need to be versioned? This can be done in the schema instead of the model creation.

Yes but if you do that, will all the fields selected be saved in the history or just what changed? 
When querying the database this could increase bandwidth as a database gets older. Is there a way to query dynamo asking for a certain part of the object to not be included:

```javascript
User.get({
        username:"user"
    }, 
    {
        __v: false
    })
```


Agree with remove of updatedAt in the history field entries.

troyswanson:
> When querying the database this could increase bandwidth as a database gets older.

This is a good point, and your method of dealing with that makes the most sense. I don't think what you're asking for is possible, however.

[...]

PR #325, opened 2018-03-23
Plugin Support - BETA

### LINKED ISSUE #18: Support for plugins
MongoDB mongoose supports replicating updates to an elastic search cluster using https://github.com/taterbase/mongoosastic, which is implemented as a plugin.   

Might be nice to be able to do this with dynamoose as well.

Again, can implement and potentially submit as a PR.

fishcharlie:
# Plugin Support - BETA

**WARNING: PLUGINS IS CURRENTLY IN BETA. THIS FUNCTIONALITY MIGHT CHANGE AT ANYTIME WITHOUT WARNING. DO NOT CONSIDER THIS FEATURE TO BE STABLE.**

## Summary

It's **finally** here (or coming soon, if this isn't merged in yet). Plugin Support is coming to Dynamoose. We are currently aiming to release this with Dynamoose version 1.2.0.

I'm creating this PR *way* too early. The reason for this is I'm trying to build this feature in the open to receive feedback and suggestions so the community can help build this and make it the best it can be.

## Ideas/Structure

- Plugins will be linked to models (possibly Dynamoose defaults as well in the future to link a plugin to all models, but will be integrated heavily into Dynamoose models)
- It will be possible to register multiple plugins to one model
- Plugins will watch for events to be "sent" or "emitted" from Dynamoose and can manipulate or change data
- If `*` is passed into action, in the `.on` method it will match all emitted types. Or maybe we can make both `action`/`type` and `stage` optional.

## Ideas we are exploring

- Add support for plugins requiring a certain version of Dynamoose (and/or Node.js?)
- Add feature to [dynamoosejs.com](https://dynamoosejs.com) to search plugins

Items marked with `***` means that we are currently unsure how to implement this feature. Although it would be a cool feature, the details of the idea haven't been thought through to a level that makes implementation possible.

### Possible Plugin Implementation

```js
module.exports = function(obj, options) {
	obj.setName(“My Plugin”); // required
	obj.setDescription(“”); // optional
	obj.on(‘init’, function() { // this will handle all stages related to init
		console.log(“Plugin registered”);
	});
	obj.on(‘scan’, ‘preRequest’, function(objB) { // this will handle only preRequest stages on the scan type
		console.log(“About to make request to DynamoDB”);
	});
	obj.on(‘scan’, ‘postRequest’, function(objB) { // this will handle only postRequest stages on the scan type, and will wait for promise to resolve before moving on
		return new Promise(function(resolve, reject) {
			resolve();
		});
	});
	return obj;
}
```

The `obj.on` method will add listeners that will be called when Dynamoose emits certain actions. It will accept 3 properties (action, [stage ,] func). `action` is a higher level stage that Dynamoose will emit. `stage` is a lower level stage that gives detail about where in the process Dynamoose is (for example right before AWS request, right after AWS request) all within a higher level action such as (.get, .scan, etc). If `stage` is not passed in, the function will be called for all emitted actions that match.

### Adding Plugin to Model

```js
var MyPlugin = require('ThePluginPackage');
var MyPluginB = require('ThePluginPackageB');
var Model = dynamoose.model('Puppy', {
    id: {
        type: Number,
        validate: function(v) {
            return v > 0;
        }
    },
    name: String,
    owner: String,
    age: {
        type: Number
    }
});
Model.plugin(MyPlugin); // this plugin will always take priority over the plugin below and be run first
Model.plugin(MyPluginB, {username: 'test', password: 'test'}); // this plugin will always take priority second and will be run after the first plugin, this plugin also passes options into the plugin

module.exports = Model;
```

## Potential actions/types to emit

- Plugin Register (“plugin:register”) (**)
- AWS Table creation (“table:create”)
- Scan (“model:scan”) (**)
- Query (“model:query”) (**)
- Save (“newmodel:save”) (this looks to be `put`)
- Put (“model:put”) (**)
- BatchPut (“model:batchput”) (might consider later)
- Create (“model:create”) (might consider later)
- Get (“model:get”) (**)
- Populate (“model:populate”) (might consider later)
- New Model Item (“newmodel:new”) (might consider later)
- Update ("model:update")

`**` indicates an action that has been fully implemented.

## Potential stages to emit

Stages (examples, depends on method):

- Right when call is first made (right when .save call is made)
- After Dynamoose has run the Dynamoify function
- Right before call will be made to 

## Use Cases we need to support

- Issue #18 
- Issue #159 (might need to add original object received from DynamoDB to instance)
- [Passport.js Plugin Support](https://www.npmjs.com/package/passport-local-mongoose)
- PR #317 (maybe we don't need to support this not)

## Todo Items

- Build some example/test plugins
- Add support for adding global plugins (to all models)

## Questions

- Is plugin compatibility a concern? And if so, should/can we do anything about it?
- Is having too many plugins/listeners on a model a concern? Any ideas on how to increase performance or is this a non issue?

## Other Notes

- I will be updating this PR with more detail as I work on it more
- Feel free to comment or [contact me](https://blog.charlie.fish/contact/) if you have suggestions, questions, ideas, etc.
- Close #18


lib/Scan.js
```diff
@@ -185,9 +187,10 @@ Scan.prototype.exec = function (next) {
     }
     scanOne();
     function scanOne() {
-
+      Model._emit('model:scan', 'request:pre', {event: {scan: this, callback: next, scanReq: scanOneReq}, actions: {updateScanReq: function(req) {scanOneReq = req;}}});
       debug('scan request', scanOneReq);
       Model.$__.base.ddb().scan(scanOneReq, function(err, data) {
+        Model._emit('model:scan', 'request:post', {event: {scan: this, callback: next, data: data, error: err}, actions: {updateError: function(error) {err = error;}, updateData: function(myData) {data = myData}}});
```
fishcharlie: Concerned about async issues with some of these action functions

mogwai:
During an event emition will the code of that event lock the next part of the process from happening?

```js
plugin.on('scan', 'postRequest', function(obj) { // this will handle only postRequest stages on the scan type, and will wait for promise to resolve before moving on (NOT SURE IF WE WILL SUPPORT THIS)
	return new Promise(function(resolve, reject) {
		resolve();
	});
});
```

I think that we should support this: 

```js
plugin.on('scan', 'postRequest', async function(obj) { 
  await somePromise();
});
```

Otherwise the next part of the process (Say to scan) will be executed prematurely. Otherwise the function should take a callback function that can be called when the logic is finished.

fishcharlie:
@mogwai The only clean way I can think of how to do this is to use the `async`/`await` syntax. I think adding in this support might have to come in version 1.0. Or do you feel like it's a big enough feature that needs to be released alongside plugin support?

mogwai:
Yes you mention it in the comments for the Listners. This would also mean awaiting the emitions throughout the codebase though for it to work properly. 

I think this is an important issue. Some plugins will just wish to respond to certain events while others will want to override the outcome of that function. 

For example, for caching, I need to sometimes prevent the continued execution of the scan function.

mogwai:
>  Really don't want to make breaking changes until v1.0

Fair enough :) 

> I wonder if we can just pass in the promise resolve and reject functions i

I'm not sure how this would work?

Perhaps would could write try catch blocks so the listener can throw a silent error to stop the operation from being completed?

You do give access to the model so for caching plugin I could just overwrite the scan.exec operation. This doesn't really feel like I'm using the plugin structure though.

fishcharlie:
@mogwai Although I don’t think that was the intention of the try catch block that actually sounds like a super slick solution to limit the amount of code we have to write. Although it won’t really work until we implement the await syntax right? I can’t think of a easy way to do it without the await syntax off the top of my head.

mogwai:
Neither can I. Are you going to use the try catch blocks then?

mogwai:
Yeah I can't really see how I'm going to do this without promise support to be honest. I think this architecture mainly focuses on the idea of reacting to dyanmoose operation lifecycle events. The plugin I want to make is going to need to rewrite and prevent them.

mogwai:
Would be cool to setup plugins for all the models?
```js
dynamoose.setDefaults({
   plugins: []
})
```

mogwai:
Also it would be good to be able to set default options for the plugins in the setDefaults method?

fishcharlie:
@mogwai In v1.0 we will be dropping support for all Node.js versions below v8. I'm not sure exactly how to adjust this to support your use case, but I know the first step was to support newer versions of Node.js. Thoughts?

mogwai:
Nice ok that is a good idea I think

Aedalus:
How are pre-event hooks meant to work? Are they just fire and forget, or if they fail, will they prevent the even from firing as well?

fishcharlie:
@Aedalus Not sure I completely understand your question.

fishcharlie:
For the request that @mogwai had regarding async support and exiting early. Please see [this branch](https://github.com/fishcharlie/dynamoose/tree/pluginsupport-async).

Things still to do on that branch:

- Add tests
- Write documentation

fishcharlie:
@mogwai You want to pull down the branch I mentioned in https://github.com/dynamoosejs/dynamoose/pull/325#issuecomment-435481704 and see if you can create the plugin you were looking at and give me some more feedback?

mogwai:
Yeah @fishcharlie I'll give it a go this week

mogwai:
Hi Charlie, I'm looking into your plugins-async branch. I can see that a caching plugin could now be created using the deferredMain promise and preventing the continued execution by returning false to continue. 

My concern is that there is a possibility of Q.defer being removed soon? If so it might affect the way the plugin will work?

fishcharlie:
@mogwai PR #456 should not be a breaking change. If you return or use a promise in your plugin, the code I wrote should just `await` on that promise. I don't think that should effect how promises work.

If you want to use `Q` in your plugin, that should probably be fine so long as you include it in your `package.json` as a dependancy. From my understanding `Q` works with the same stuff as JavaScript promises. Like `async`/`await`. So if you return a `Q` promise, our code should be able to handle it. Or of course you could just use native JavaScript promises.

mogwai:
Well thats another thing on its own. I don't think the library should force the user to use a promise. It should check whether the return value is a promise and await it if so. Other plugins may not need to return a promise and will be forced to return `Promise.resolve({})` or the library will throw an error. What are you thoughts?

I suggest changing the Plugin.js to:
```js
let result = listener.emit(type, stage, obj);
result = result.then ? await result : result;
```
In the `scanOne()` function, you call `..._emit('model:scan', 'request:pre'...` listeners and you pass the deferredMain. I'm guessing you've done this so a plugin can return something to the user, overwriting the methods original functionality. Is this correct?

fishcharlie:
@mogwai If a plug-in returns a non promise that should be perfectly fine. I’m pretty sure await will handle that gracefully without any additional logic on our end.

I could be TOTALLY wrong tho.

mogwai:
Here is an example of failure: 

```js
async function main() {
    const x = await asda();
    console.log(x)
}

function asda(){
  return 1;
}
```

mogwai:
Your right :) ignore that comment

fishcharlie:
I have updated the documentation in my [pluginsupport-async](https://github.com/fishcharlie/dynamoose/tree/pluginsupport-async) branch. I will be merging that branch into this PR soon (might wait for @mogwai's input first tho).

I'm aiming to get this PR merged in before the next version (#464). Keep in mind this will just be a beta version and not considered stable for that release. The goal will be to gain more input and feedback from developers to solidify it in a future release.

fishcharlie:
I have just merged the async branch into this PR.

PR #331, opened 2018-04-03
Adding better error handling for parsing invalid data from DynamoDB

### LINKED ISSUE #281: Improving error handling
Its quite common for someone in my team to run into an error like

```
err SyntaxError: Unexpected token N in JSON at position 0
    at parse (<anonymous>)
    at dedynamofy (/Users/Ro/www/passphere/lambda-api/node_modules/dynamoose/lib/Attribute.js:430:14)
 ...
```
In this case it comes from using JSON.parse on NaN (when it was expecting a number).

It would be great to throw a more useful error such as
```
err ModelError {
  name: 'ModelError',
  message: 'Attribute "total" of type "N" has an invalid value of "NaN"' }
```

I got this by wrapping `dedynamofy` in `Attribute` with a `try catch`.

There might be a better place to catch these errors, I didn't dive to deep.

Are you interested in a PR for this? Where would you like to catch the error and is `ModelError` the correct error to throw?

fishcharlie:
Close #281

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -453,23 +453,27 @@ Attribute.prototype.parseDynamo = function(json) {
   }
 
   function dedynamofy(type, isSet, json, transform, attr) {
-    if(!json){
-      return;
+    try {
+      if(!json){
+        return;
+      }
+      if(isSet) {
+        var set = json[type + 'S'];
+        return set.map(function (v) {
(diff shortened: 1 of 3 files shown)


ro-savage:
Looks good @fishcharlie .

I haven't tested it, but code review looks good.

PR #334, opened 2018-04-03
Version 0.9.0

fishcharlie:
## Version 0.9.0

Dynamoose version 0.9.0 will be skipped.

Please comment or [contact me](https://charlie.fish/contact) if you have any questions about this release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dynamoose",
-  "version": "0.8.7",
+  "version": "0.9.0",
   "description": "Dynamoose is a modeling tool for Amazon's DynamoDB (inspired by Mongoose)",
   "typings": "./dynamoose.d.ts",
   "homepage": "https://github.com/automategreen/dynamoose",


mogwai:
> In version 1.0 we will be removing support for older Node.js versions

@fishcharlie What node version will be supported?

fishcharlie:
@mogwai I’m not totally sure which Node.js versions we will be removing support for yet in version 1.0. I believe we should move to a system where we support active LTS Node.js versions. But I’m not sure if that is too aggressive of a strategy.

The new ES6 syntax is fun and I’d love to be able to move Dynamoose to using the new syntax much faster. Especially since it’s a lot easier to upgrade Node.js than front end JS.

fishcharlie:
@mogwai Any suggestions or thoughts on that front?

mogwai:
I think >v7 to let us use await/async in the library. 

What are your thoughts on using native promises?

fishcharlie:
@mogwai Native promises is definitely a goal for 1.0.

mogwai:
Looking at the npm page for the (package)[https://www.npmjs.com/package/dynamoose] we could be aggressive as its not crazy popular (yet ;) ) 

> The new ES6 syntax is fun and I’d love to be able to move Dynamoose to using the new syntax much faster. Especially since it’s a lot easier to upgrade Node.js than front end JS.

I agree, but its important to allow as many people to use this as possible. AWS lambda only supports  v4.3, v6.10 and v8.10.

mogwai:
I'm not really sure whats best but I would think we should consider our options based on what aws supports and being a bit agressive

fishcharlie:
@mogwai I am 100% open to only supporting Node.js version 8+ in Dynamoose version 1.0. Especially since AWS Lambda now supports Node.js v8.10. This would give the flexibility to use the new `async`/`await` syntax and truly go all in on the new syntax/features. Dynamoose version 1.0 I don't envision adding a lot of new features. It will be a lot of optimizations, refactoring, and updating syntax. So I think it makes sense to try to be as aggressive as possible right now so we don't have to do refactor/breaking updates as often.

That is just my initial thoughts tho. I always thought people were pretty aggressive at updating Node.js to the latest version and adapting the new syntax due to it being **so** easy to update. But I read an article a little bit ago (might have been on this repo, I forget) and a large portion of projects still use much older versions of Node.js. It makes a lot more sense why front end JS takes a lot longer to be adapted, but I'm not sure why Node.js isn't updated and adapted faster due to you only having to update your servers, not your users browsers.

I think this is also a discussion that @brandongoode should give some input on. Luckily we have a lot of time before we have to make this decision so it's good we are starting to think about it early, but we still have time, and we have to get 0.9 out before 1.0.

fishcharlie:
Version 0.9.0 will not be happening in favor of going straight to version 1.0.0. Let me know if you have any questions!

PR #336, opened 2018-04-11
Fix saveUnknown throwing errors when value where falsy

jonathonlui:
When a model has saveUnknown=true and trying to save the model with an unknown attribute with `0` or `false` values was causing `Attribute.setTypeFromRawValue` to throw a `SchemaError`.

I remove the value check from `setTypeFromRawValue`, which I assume was unnecessarily copied from `setType`. This allows `0` and `false` to be correctly converted to dynamo types.

Attributes with the other falsy values like are `null`, `undefined`, or  'NaN' are now excluded because it was causing `new Attribute` to throw `TypeError` at https://github.com/automategreen/dynamoose/blob/35deac2aaaa418f01b37d9bf8427626a22d142d7/lib/Attribute.js#L12

Also fix the NewModel name test to include missing suffix.

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -150,10 +150,6 @@ Attribute.prototype.types = {
 
 Attribute.prototype.setTypeFromRawValue = function(value) {
   //no type defined - assume this is not a type definition and we must grab type directly from value
-  if(!value) {
-    throw new errors.SchemaError('Invalid attribute value: ' + value);
-  }
-
   var type;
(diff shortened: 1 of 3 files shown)


fishcharlie:
Going to keep this open for a bit, just to make sure no one has any other thoughts, and to time this with the release of Dynamoose 0.9. We'll for sure try to get this into version 0.9.

And thank you so much for this!

Issue #301, opened 2018-03-03
Nested Schemas/Models


fishcharlie:
Dynamoose should support the following.

```
var productsSchema = new dynamoose.Schema({
	_id: String,
	name: String
});
module.exports = productsSchema;


var userSchema = new dynamoose.Schema({
	_id: String,
	firstname: String,
	lastname: String,
	email: String,
	password: String,
        products: [productsSchema]
}, {
	timestamps: true
});
module.exports = dynamoose.model('myProject_User', userSchema);
```

OR

```
var productsSchema = new dynamoose.Schema({
	_id: String,
	name: String
});
module.exports = productsSchema;


var userSchema = new dynamoose.Schema({
	_id: String,
	firstname: String,
	lastname: String,
	email: String,
	password: String,
        myProduct: productsSchema
}, {
	timestamps: true
});
module.exports = dynamoose.model('myProject_User', userSchema);
```

fishcharlie:
Then you could create it by doing the following.

```
var Product = require("...")
var User = require("...")

var myProduct = new Product({_id: "item1", name: "Computer"});
var myUser = new User({_id: "user1", product: [myProduct]});
```

OR

```
var Product = require("...")
var User = require("...")

var product = new Product({_id: "item1", name: "Computer"});
var myUser = new User({_id: "user1", myProduct: product});
```

fishcharlie:
I might take a stab at submitting a pull request to integrate this if I have time. Any tips, suggestions, pointers, things I need to watch out for, thoughts on how to integrate would be super helpful.

fishcharlie:
See PR #329 for this.

Euklidian-Space:
Hey @fishcharlie I would like to take a crack at this if that's cool.  Got a question though.  I would like to run just one test file.  I am trying to get familiar with the inner workings of model and schema by doing discovery testing, but it is time consuming to run the whole test suite.  

Thanks,
ES

fishcharlie:
@Euklidian-Space That'd be awesome!! I saw a schema like the following before, but I don't know if it'd work.

```
var UserSchema = new dynamoose.Schema({
	userId: {
		hashKey: true,
		type: String
	},
	creationTime: {
		rangeKey: true,
		type: Number
	},
	first_name: {
		type: String
	},
	last_name: {
		type: String
	},
	device_info: [{
		device_id: {
			type: String,
			default: null
		},
		access_token: {
			type: String,
			default: null
		}
	}],
[...]

PR #337, opened 2018-04-11
Allow attributes to be objects with a `type` prop when using saveUnknown

jonathonlui:
So when saveUnknown is true, trying to save a model with an attribute that is an object can cause problems if the object has properties that match dynamoose's attribute definition options (type, required, hashKey)

For example say I want to include information about a cat's collar:

```javascript
var schema = { name: { type: String, hashKey: true }};
var options = { useDocumentTypes: true, saveUnknown: true };
var Cat = dynamoose.model('Cat', schema, options);

var cat = new Cat({ name: 'Fluffy', collar: { color: 'red', type: 'flat' }});
cat.save();
```

In DynamoDB the cat would look like:
```javascript
{
  name:  {S: 'Fluffy'},
  collar: {S: '[object Object]'}
}
```
The collar attribute isn't the expected map type because it had a `type` property and the `Attribute.setTypeFromRawValue` ends up using `collar.type` ('flat', a string value) as the value to determine the dynamo type of the attribute.

Then `toDynamo` calls `.toString()` on the `collar` object which returns `[object Object]`.

This PR changes `Schema.toDynamo` so it calls `Attribute.create` with an extra flag (`ignoreType`) that will causes the attribute not use the `value.type` property when setting the type.

It solves my specific use case (storing nested attributes that are objects with a `type` property), but I'm not 100% sure if this is the best way to handle this. 

If this PR isn't a good idea then we should consider at least adding a note about this behavior to the docs for the `saveUnknown` option or maybe emitting a warning or throwing an error if a model is being saved with "type" as the nested attribute name.

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -5,11 +5,11 @@ var util = require('util');
 
 var errors = require('./errors');
 
-function Attribute(schema, name, value) {
+function Attribute(schema, name, value, ignoreType) {
   this.options = {};
 
   debug('Creating attribute %s %o', name, value);
-  if (value.type){
+  if (value.type && !ignoreType){
     this.options = value;
   }
 
@@ -17,7 +17,7 @@ function Attribute(schema, name, value) {
 
   this.name = name;
   if (this.schema.options.saveUnknown) {
-    this.setTypeFromRawValue(value);
+    this.setTypeFromRawValue(value, ignoreType);
   } else {
     this.setType(value);
   }
@@ -43,7 +43,7 @@ function Attribute(schema, name, value) {
 
   if (this.type.name === 'map'){
 
-    if(value.type) {
+    if(value.type && !ignoreType) {
       value = value.map;
     }
     for (var subattrName in value){
@@ -148,7 +148,7 @@ Attribute.prototype.types = {
   }
 };
 
-Attribute.prototype.setTypeFromRawValue = function(value) {
+Attribute.prototype.setTypeFromRawValue = function(value, ignoreType) {
   //no type defined - assume this is not a type definition and we must grab type directly from value
   if(!value) {
     throw new errors.SchemaError('Invalid attribute value: ' + value);
@@ -156,7 +156,7 @@ Attribute.prototype.setTypeFromRawValue = function(value) {
 
   var type;
   var typeVal = value;
-  if (value.type){
+  if (value.type && !ignoreType){
     typeVal = value.type;
   }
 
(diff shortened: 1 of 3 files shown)


fishcharlie:
@jonathonlui I am kinda concerned about possible side-effects of this PR. #301 and #329 are both things I'd like to get into Dynamoose at some point, and I wonder if this could cause conflict with those. Due to the fact that I haven't thought through #301 and #329 enough, it could be an issue, or it could be a non-issue.

I do this the goal of this PR makes sense and is something I'd love to get merged in. Do you have any thoughts about it?

I could be totally wrong about all of this. It's just first thoughts that come to mind when reviewing it.

jonathonlui:
Not sure what a good solution is.

I'm thinking of creating an `UndefinedAttribute` class that is purpose-built for these types of attributes instead of trying to update the regular `Attribute` class to work correctly with `saveUnknown`.

Another option is to separate the `Attribute` class's "value" parameter into "type definition" and  "value" arguments so it's easier to prevent the type definition and value from causing inadvertent effects on each other.

Both these options are larger changes, but it's something I'm open to work on when I have time.

Since I need a solution for my use case soon(-ish), for a client project I'm going to write a validator to check data  and maybe mangle/unmangle my data's attribute names so that dynamoose definition keywords aren't used in the data. I'll try to get it as a PR for dynamoose if can generalize it.

I also tried to solve my issue by using `saveUnknown=false` and defining the attribute with an "object" definition but the definitions "type" property still gets interpreted in a way that's not what I want.

For example:

```javascript
  var Cat = dynamoose.model('Cat',
    {
      id: {
        type:  Number,
        hashKey: true
      },
      collar: {
        color: String,
      }
    },
    {useDocumentTypes: true}
  );

  Cat.create({
    id: 2,
    collar: {
      color: 'red',
    },
  });
```

Works fine and creates this dynamodb item:
```javascript
{
  id:  {N: '2'},
  collar: {
    M: {
      collar: {S: 'red'},
    },
  },
}
```

Whereas this:

```javascript
  var Cat = dynamoose.model('Cat',
    {
      id: {
        type:  Number,
        hashKey: true
      },
      collar: {
        color: String,
        type: String,
      }
    },
    {useDocumentTypes: true}
  );

  Cat.create({
    id: 2,
    collar: {
      color: 'red',
      type: 'flat',
    },
  });
```

Should create this dynamodb item:
```javascript
{
  id:  {N: '2'},
  collar: {
    M: {
      collar: {S: 'red'},
      type: {S: 'flat'},
    },
  },
}
```
but creates this:
```javascript
{
  id:  {N: '2'},
  collar: {S: '[object Object]"'},
}
```

fishcharlie:
@jonathonlui I'm all for making some improvements in this area. But I'm leaning towards just throwing some type of warning about `type`. I don't think `type` is something that developers should be using as object properties due to the conflicts with Dynamoose.

If you feel strongly about actually fixing this on a deeper level than just throwing a warning please comment here to continue the discussion.

I'm closing this PR for now. Please feel free to create an issue or PR for the warnings part of this.

PR #338, opened 2018-04-13
model.originalItem()

fishcharlie:
Adding a `model.originalItem()` function.


fishcharlie:
The test seems to be failing on Node.js 0.12.

PR #339, opened 2018-04-13
Fix loading document type attributes from DynamoDB when when saveUnknown=true

jonathonlui:
This PR fixes an issue when using `saveUnknown` + `useDocumentTypes` and trying to retrieve  a model that hasn't yet been "used" for example:

```javascript
  var Cat = new Schema({
    id: Number,
  }, {
    useDocumentTypes: true,
    saveUnknown: true,
  });

  // Assume an item is in the Cat table:
  // {
  //   id: { N: '1' }
  //   petProperties: {
  //     M: {
  //       name: { S: 'Fluffy' },
  //       age: { N: 10 },
  //     }
  //   }
  // }

  Cat.get({ id: 1}};

  // throws SyntaxError: Unexpected token o in JSON at position 1
```

This is because since the Cat schema doesn't have `getProperties` attribute `Schema.parseDynamo` will create an new attribute with an `Object` type because `Attribute.lookupType` returns `Object` for Dynamo "'M" types.

This then causes the `Attribute.parseDynamo` to be parsed using `JSON.parse` instead of the correct `mapify`.

Interestingly this error doesn't happen if the model has been used to create (or put/save) before doing get/query/scan.

```javascript
  var Cat = new Schema({
    id: Number,
  }, {
    useDocumentTypes: true,
    saveUnknown: true,
  });

  Cat.create({
    id : 1,
    petProperties: {
      name: 'Fluffy',
      age: 10,
    },
  );

  // Cat's schema now has a petProperties attribute

  Cat.get({ id: 1}};

  // No error
```

This is because when `Cat.create` runs, it creates the `petProperties` attribute in the model's schema, and when `Cat.get` runs, `Schema.parseDynamo` see that is has `petProperties` and doesn't try to create a new, incorrectly-typed `Attribute`.

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -160,14 +160,17 @@ Attribute.prototype.setTypeFromRawValue = function(value) {
     typeVal = value.type;
   }
 
-  if (util.isArray(typeVal)){
+  if (util.isArray(typeVal) || typeVal === 'list'){
     type = 'List';
   } else if ( (util.isArray(typeVal) && typeVal.length === 1) || typeof typeVal === 'function') {
     this.isSet = util.isArray(typeVal);
     var regexFuncName = /^Function ([^(]+)\(/i;
     var found = typeVal.toString().match(regexFuncName);
     type = found[1];
-  } else if (typeof typeVal === 'object'){
+    if (type === 'Object') {
+      type = 'Map';
+    }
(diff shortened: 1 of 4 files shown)


fishcharlie:
Thanks so much for this @jonathonlui. I'm going to have @brandongoode review this at some point before we get it merged in. We will aim to get this in version 0.9. Thanks again!

PR #341, opened 2018-04-17
Fix scanByRawFilter containing - Select: 'COUNT'

bsimpson53:
Trying to do a Select: 'COUNT' on a raw filter scan like:

```
var Dog = dynamoose.model('Dog');
var filter = {
  FilterExpression: 'details.timeWakeUp = :wakeUp',
  ExpressionAttributeValues: {
    ':wakeUp': '8am'
  },
  Select: 'COUNT'
};
Dog.scan(filter, { useRawAwsFilter: true }).exec()
```

generates this error:

```
Uncaught TypeError: Cannot read property 'map' of undefined
      at Response.<anonymous> (node_modules/dynamoose/lib/Scan.js:82:45)
```

Intercept the case where a raw filter scan where data is defined but data.Items is unpopulated and therefore by elimination the filter must have contained ```Select: 'COUNT'```

### DIFF
--- a/lib/Scan.js
+++ b/lib/Scan.js
@@ -74,6 +74,10 @@ Scan.prototype.exec = function (next) {
           if (!data) {
             return deferred.resolve([]);
           }
+          if (!data.Items) {
+            var counts = { count: data.Count, scannedCount: data.ScannedCount };
+            return deferred.resolve(counts);
+          }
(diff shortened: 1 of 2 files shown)


fishcharlie:
Thanks so much for this @bsimpson53, we will for sure try to get this in for version 0.9! Thanks again!!

PR #350, opened 2018-04-25
added support for specifying the ReturnValues option in update

manast:
```Model.update``` always returns all the new values in every call. This can have a severe performance penalty in some scenarios. This PR provides a backwards compatible option ```returnValues``` when calling ```update``` so that the 5 different settings provided by Dynamo can be used:

- NONE - If ReturnValues is not specified, or if its value is NONE, then nothing is returned. (This setting is the default for ReturnValues.)
- ALL_OLD - Returns all of the attributes of the item, as they appeared before the UpdateItem operation.
- UPDATED_OLD - Returns only the updated attributes, as they appeared before the UpdateItem operation.
- ALL_NEW - Returns all of the attributes of the item, as they appear after the UpdateItem operation.
- UPDATED_NEW - Returns only the updated attributes, as they appear after the UpdateItem operation.

### DIFF
--- a/docs/_docs/model.md
+++ b/docs/_docs/model.md
@@ -386,6 +386,18 @@ A map of values for the condition expression. Note that in order for
 automatic object conversion to work, the keys in this object must
 match schema attribute names.
 
+**returnValues**: string
+
+From [the AWS documentation](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html)
+Use ReturnValues if you want to get the item attributes as they appear before or after they are updated. For UpdateItem, the valid values are:
+
+- NONE - If ReturnValues is not specified, or if its value is NONE, then nothing is returned. (This setting is the default for ReturnValues.)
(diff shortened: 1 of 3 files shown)


fishcharlie:
@manast This looks amazing. Thanks so much!! Is there any chance you could add documentation regarding this new feature? We should also probably confirm to make sure there is a test to ensure backwards compatibility, and if not we should add one to confirm the backwards compatibility.

We will try to get this in for version 0.9. Thanks again!!

fishcharlie:
Thanks again @manast! We'll get this in for version 0.9.

dobrynin:
Can we add a note to the documentation clearly stating that the Dynamoose default differs from the DynamoDB default? The current phrasing led me to believe that `NONE` was the default for Dynamoose.

Also, what do you think about adding an option `defaultReturnValues` so that you can globally override the current default of `ALL_NEW`?

fishcharlie:
@dobrynin Both ideas sound great! PR would be awesome.

PR #354, opened 2018-05-04
Recreate dynamo db instance on .local()

renanvalentin:
The current solution requires .local() to be called first in order to
use dynamo db locally.

Recreating dynamo db instance when calling .local() solves issues like
script import order.

Example:
```javascript

// evil-cat.js
import dynamoose from 'dynamoose';

export const EvilCat = dynamoose.model('Cat', { id: Number, name: String });

// evil-cat.test.js
import dynamoose from 'dynamoose';
import { EvilCat } from 'db/evil-cat';

beforeAll(() => {
  dynamoose.local(); // won't work because dynamoose was already initialized inside of evil-cat.js 
});
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,8 +3,15 @@
 var Schema = require('./Schema');
 var Model = require('./Model');
 var https = require('https');
+var AWS = require('aws-sdk');
 
 var debug = require('debug')('dynamoose');
+
+function createLocalDb(endpointURL) {
+  return new AWS.DynamoDB({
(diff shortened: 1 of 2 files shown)


fishcharlie:
@renanvalentin Looks like this could be valuable. But the tests are failing due to linter errors. Can you run the tests locally (`npm test`) and fix those? 🙂

renanvalentin:
@fishcharlie Fixed 🗡

fishcharlie:
We'll get this merged in for version 0.9. Thanks so much!

PR #356, opened 2018-05-14
Saying which table is effected by Error

mogwai:


lib/Table.js
```diff
@@ -162,8 +162,9 @@ Table.prototype.init = function (next) {
 
         } else {
           if (indexes.delete.length > 0 || indexes.create.length > 0) {
-            debug('indexes are not synchronized and update flag is set to false');
-            deferred.reject(new Error('indexes are not synchronized and update flag is set to false'));
+            var mess = table.name + 'indexes are not synchronized and update flag is set to false';
```
fishcharlie: Can we add a space before the `indexes` word so there is a space between the table name and indexes?

fishcharlie:
@mogwai For that last commit, isn't that pretty similar to #331?

mogwai:
Yup shall I remove it?

fishcharlie:
@mogwai Yeah that'd be great.

PR #357, opened 2018-05-15
Update the return type of ddb()

koladilip:
This is giving type error when we called the function and when we corrected the type to _AWS.DynamoDB instead of typeof _AWS.DynamoDB error is gone.


fishcharlie:
@koladilip We will get this in for version 0.9.

Issue #360, opened 2018-05-18
Using saveUnknown throws an error when trying to save the boolean value false


pscamman:
https://github.com/automategreen/dynamoose/blob/0bbae14aa47d6abe70757194a803f70ccbd872dc/lib/Attribute.js#L151-L155
The above if statement triggers on any falsy value. It is reached when saveUnknown is true:
https://github.com/automategreen/dynamoose/blob/0bbae14aa47d6abe70757194a803f70ccbd872dc/lib/Attribute.js#L19-L21

Unless I'm missing something, saving false (and null even) should be an option here, so this looks like unintended behavior to me.

This wasn't a problem back in 0.8.6.

fishcharlie:
@pscamman That does sound like unintended behavior. It looks like it has to do with PR #240. I'm not quite sure how to fix this currently. Any tips would be awesome. Otherwise I'll try to get around to taking a look at this soon.

pscamman:
@fishcharlie Cool. If I find some extra time I'll see about taking a deeper look myself as well.

ecwyne:
It seems to me that this is fixed if https://github.com/automategreen/dynamoose/blob/master/lib/Attribute.js#L154 is removed entirely.

This enables saving `false`, `''`, and `0`

The only other "falsey" values are `null`, `undefined`, and `NaN`.

`null` is a specific [type](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_AttributeValue.html) in dynamo
`undefined` should probably just be ignored (not saved)
(Not sure what to do with `NaN`)

ecwyne:
See PR #363 for my proposed fix

ecwyne:
@fishcharlie is right, there is already a PR that addresses this (#336)

You mentioned including this fix in `0.9`, is there any sense of how soon that will be? (This bug is preventing me from moving forward)

Thanks so much for the awesome library! It makes working with dynamo SO much more enjoyable!

fishcharlie:
@ecwyne I'm currently waiting to hear back from @brandongoode. Currently although I have quite a bit of access on this project, I do not have the ability to publish new versions on NPM. I'm going to try to work with @brandongoode to discuss ways to keep this project alive. If that doesn't happen, at some point I will consider creating a new version of this repo (fork) to ensure it gets maintained.

[...]

PR #363, opened 2018-06-11
Allow saving falsey values with saveUnknown

ecwyne:
Addresses #360 

Values of `false`, `''`, and `0` should all be saved without error.

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -150,7 +150,7 @@ Attribute.prototype.types = {
 
 Attribute.prototype.setTypeFromRawValue = function(value) {
   //no type defined - assume this is not a type definition and we must grab type directly from value
-  if(!value) {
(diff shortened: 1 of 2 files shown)


fishcharlie:
This looks to be a duplicate of #336

ecwyne:
Tests are only failing because of use of `const` keyword in `chalk` on `node v0.12`

fishcharlie:
@ecwyne It actually looks like Chalk is a major problem for other PRs... maybe skipping version 0.9 and going right to 1.0 and removing support for older Node.js version is the way to go. For your use cases what version(s) of Node.js do you require Dynamoose to support? There has been some discussion about what versions of Node.js to support. What are your thoughts on this?

ecwyne:
I'm using `v8.0` on lambda. No need to support older versions from me!

PR #371, opened 2018-06-19
Only use JSON.parse to check for Objects

ecwyne:
To Reproduce:

```
const db = dynamoose.model('db', new dynamoose.Schema({
  id: { type: String, hashKey: true}
}), {
  saveUnknown: true
});

db.create({
  id: 'foobar',
  unknown: '12345',
}).then(() => {
  db.get({ id: 'foobar' }).then(console.log);
});
```

This issue came up because I am storing a number as a string and `JSON.parse('12345')` does not throw. (Neither does `JSON.parse('true')` or `JSON.parse('false')`

In `v1.0` will dynamoose always use dynamo `Map`s and `BOOL`s? If so then we can always assume a string is a string.

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -682,8 +682,7 @@ module.exports.create = function(schema, name, obj) {
 module.exports.lookupType = function (dynamoObj) {
 if(dynamoObj.S !== null && dynamoObj.S !== undefined) {
   try {
-    JSON.parse(dynamoObj.S);
-    return Object;
+    return JSON.parse(dynamoObj.S) instanceof Object ? Object : String;
   } catch (err) {
     return String;
   }


fishcharlie:
@ecwyne Looks like the tests are failing for this. It'd probably be good to add tests cases for this too right?

In terms of Dynamo types. In `v1.0` the default will be switched to use Dynamo types by default, but this can be overwritten in the options object. I still need to find the code where I can change the default for this.

ecwyne:
The tests are failing because it expects `{ S: '1' }` to parse to `1` and not `'1'`.

I am of the opinion that if an unknown value comes in as `{ S: '1' }` the assumption should be made that it's a string (because if it had originally been created as a `Number` it would have been stored in dynamo as such)

@fishcharlie what do you think?

fishcharlie:
@ecwyne For unknown types I personally disagree. If it's an unknown type I think we should try to parse it as far as possible. Parsing it as `1` instead of `'1'` gives the developer more flexibility. Especially since it's the current behavior and I don't think changing it has enough positives and breaking the behavior doesn't seem to have enough advantages right now. What do you think?

ecwyne:
As a developer I'd expect to get back what I put in.

If I put in a `Boolean` I get back a `Boolean`
If I put in a `Number` I get back a `Number`
if I put in an `Object` I get back an `Object`
If I put in an `Array` I get back an `Array`
If I put in a `String` I get back a `String` (regardless of what is in the string)

If I wanted to get back a `Number` I would have put it in that way.

fishcharlie:
@ecwyne That's great, but if the model is using `saveUnknown` and not using Dynamo types, why shouldn't we try to parse that as far as we can? I understand what you are saying, but if we don't know the type the user meant it to be, why not parse that as far as we can?

ecwyne:
Let's say I'm storing [commit hashes](https://github.com/dynamoosejs/dynamoose/commits/master). I would expect them to all be strings, but I only have to go back ~20 to find `5982320` which would be returned to me as a `Number`?

Also, as currently written I was getting errors when reading string numbers. It tried to treat it like a `Map` and converted `'90210'` to the following:

```
{
  '0': '9',
  '1': '0',
  '2': '2',
  '3': '1',
  '4': '0'
}
```

fishcharlie:
@ecwyne That use case is totally legit. Although I think that is a one off use case. For every one of those cases there are many cases that would work better with the current behavior.

Dynamoose shouldn't interpret anything about the use case that the developer is trying to build. If you'd like it to be strict and parse it as a number then I believe enabling Dynamo types or adding a property to your schema should work.

If I get a lot of push back we might reconsider. But I think Dynamoose can't confidently interpret the type it should parse as far as it can. Especially since that is the current behavior and making this a breaking change doesn't seem worth it.

If anyone has more feedback on this I'd love to hear it.

ecwyne:
Is there any reason *not* to use Dynamo types (other than backwards compatibility)?

fishcharlie:
@ecwyne Not that I'm aware of. I think backwards compatibility is the biggest thing. I know I have tables using the old system. I'm still not quite sure how I can migrate those to the new system.

ecwyne:
@fishcharlie 
#373 is a reproduction of the issue I had originally (which prompted me to create this PR in the first place)

PR #374, opened 2018-06-24
Fix the ModelError type

xiananfan:
ModelError type seems to be copied from SchemaError. This fixes the type issue.


fishcharlie:
@xiananfan Thanks for this!

PR #376, opened 2018-07-08
Toggle useDocumentTypes and useNativeBooleans defaults to true

fishcharlie:
I'm pretty sure this is a **🚨 breaking change 🚨**, but haven't run enough tests to verify yet.

This PR sets the useDocumentTypes and useNativeBooleans default options to `true` instead of `false`. This is the final major PR of v1.0.

### DIFF
--- a/test/Schema.js
+++ b/test/Schema.js
@@ -994,6 +994,11 @@ describe('Schema tests', function (){
       name: { S: 'Fluffy' },
       anObject: { S: '{"a":"attribute"}' },
       numberString: { S: '1' },
+      anArray: { S: '[2,{"test2": "5","test": "1"},"value1"]' },
+      anObjectB: { M: {"a":{"S": "attribute"}} },
+      anArrayB: { L: [{N:'2'},{M: {'test2': {S: '5'},'test': {S: '1'}}},{S: "value1"}] },
+      aBoolean: { S: "true" },
+      aBooleanB: { BOOL: true },
     });
 
     model.should.eql({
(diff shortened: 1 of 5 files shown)


fishcharlie:
I **THINK** I recently ran into this problem just by using useDocumentTypes and useNativeBooleans on the current release version. This makes me wonder if it's just a bug in the code that we need to track down and find.

brandongoode:
Looking at the changes now.  Sorry for the delay.

test/Schema.js
```diff
@@ -154,9 +154,9 @@ describe('Schema tests', function (){
     schema.attributes.aObject.type.name.should.eql('object');
     should.exist(schema.attributes.aObject.default);
```
brandongoode: Instead of changing these tests, it may be more useful to set useDocumentTypes and useNativeBooleans to false.

fishcharlie:
@brandongoode Just addressed those comments. Looks like those tests I mentioned earlier are still failing. You able to take a look at those? I can't figure out why they are failing.

brandongoode:
@fishcharlie I'm pull down the PR and see if I can figure out what's failing.

fishcharlie:
@brandongoode Thanks so much!!

fishcharlie:
@brandongoode Does the branch `fishcharlie-dynamodbtypesdefaulttoggle` fix this? Or is it still in progress? Looks like that branch is passing the CI tests. Not sure if you are still working on it tho.

brandongoode:
@fishcharlie Yes.  I've modified a few of the test to resolve the errors you were having.  I also removed the logic to convert S types to Objects since with `useDocumentTypes: true` S should always be a String.

PR #383, opened 2018-07-13
Fix serverSideEncryption option

khamasaki:
There was a problem using the serverSideEncryption option due to the way buildTableReq was being called. The this parameter was being passed to the buildTableReq function but it was not being used, so the options from the table were being ignored. This PR calls the buildTableReq function as an instance function so the local this value is valid.


fishcharlie:
Thanks for this @khamasaki. We will get this in for version 1.0. Last major PR before version 1.0 is #376. If you want to provide any help on that, that would be greatly appreciated.

Issue #382, opened 2018-07-12
Model.update trying to update key attribute


soliveira-vouga:
Hi, I have the following model

```javascript  
new Schema({
  transactionId: { type: String, hashKey: true, required: true },
  userId: { type: String, required: true, index: { global: true, rangeKey: 'createdAt', throughput: 1 } },
  createdAt: { type: Number, rangeKey: true, required: true },
  amount: { type: Number, required: true}
}, {timestamps: true}))
````

When I try to do an update like this:

`Transaction.update({ transactionId, createdAt }, { status })`

I get the following error:

`One or more parameter values were invalid: Cannot update attribute createdAt. This attribute is part of the key`

Why is it trying to update the createdAt field? I'm supplying it as part of the key argument, not as the update argument

fishcharlie:
@soliveira-vouga I don't see any point in specifying the `createdAt` property unless you are wanting to update that as well. Specifying the `createdAt` property shouldn't narrow down the items that it will update because the hashKey `transactionId` is unique in DynamoDB. So I would try removing that property.

If that doesn't work please create a failing test case and create a PR for it and I'll look into it more.

soliveira-vouga:
`transactionId` is not unique for this particular table. `transactionId` + `createdAt` are you unique which is why transactionId as my hashKey and createdAt as my rangeKey constitute my primary key. 

The Model.update call, from my understanding, takes the key as the first argument and the attribute(s) to be updated as the second.

I'm passing in the key in the first as argument (transactionId and createdAt) and the second argument should be what I'm updating (amount).

This isn't much different from the API example where it's using `ownerId` and `ne` as hash and range keys respectedly, and `age` as the update argument.

The only difference I can see is that I'm using a let that has a bit of magic under the hood (createdAt is autopolated on creation as part of the timestamp option)

fishcharlie:
@soliveira-vouga Does this work if the rangeKey is something other than `createdAt`?

soliveira-vouga:
It does. I have just added a test to cover that scenario

soliveira-vouga:
Yes, there are workarounds. Another workaround is not using Model.update and passing the entire object into Model.save which is what I'm currently doing.

[...]

PR #384, opened 2018-07-13
Added test for updating a model with  createdAt as rangeKey

soliveira-vouga:
This PR is for https://github.com/dynamoosejs/dynamoose/issues/382

### DIFF
--- a/lib/Schema.js
+++ b/lib/Schema.js
@@ -44,8 +44,21 @@ function Schema(obj, options) {
       updatedAt = 'updatedAt';
     } else if (typeof this.options.timestamps === 'object') {
       if (this.options.timestamps.createdAt && this.options.timestamps.updatedAt) {
-        createdAt = this.options.timestamps.createdAt;
-        updatedAt = this.options.timestamps.updatedAt;
+    		if (typeof this.options.timestamps.createdAt === "string") {
+    			createdAt = this.options.timestamps.createdAt;
+    		} else if (typeof this.options.timestamps.createdAt === "object") {
+          createdAt = this.options.timestamps.createdAt.name;
+    		} else {
+          throw new errors.SchemaError('Invalid syntax for timestamp: ' + this.options.timestamps);
+        }
+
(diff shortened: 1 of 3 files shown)


fishcharlie:
In the master branch we have added new schema's so `Cat10` and maybe `Cat11` need to be renamed to something higher. So when someone fixes this merge conflict it needs to be renamed throughout the entire PR.

fishcharlie:
@soliveira-vouga I just wrote some basic logic for this. Tests are still failing tho. I feel like the logic I wrote should fix it by adding that range key. Are you sure that the only thing that we need to do to get this test to pass is add the range key? I haven't worked very much with range keys at all with DynamoDB, so I'm kinda a little bit in the dark here.

As well in order to merge this PR we still need to write tests for the `updatedAt` side of things, update the documentation to reflect this change, and update the TypeScript stuff.

fishcharlie:

dolsem:
@fishcharlie here's something interesting. I've just tried the test on the master branch of Dynamoose. It passes, but it passes because 'createdAt' field name gets replaced with '[object Object]':
```js
{
  name: 'Fluffy',
  '[object Object]': 2019-11-09T20:28:10.981Z,
  age: 2,
  updatedAt: 2019-11-09T20:28:13.270Z,
  [Symbol(originalItem)]: {
    name: 'Fluffy',
    '[object Object]': '2019-11-09T20:28:10.981Z',
    age: 2,
    updatedAt: '2019-11-09T20:28:13.270Z'
  }
}
```
This issue can be avoided if you use a slightly different but equivalent schema:
```js
const Cat20 = dynamoose.model('Cat20', {
    name: {
      type: String,
      hashKey: true
    },
    createdAt: {
      type: Date,
      rangeKey: true,
    },
    age: {
      type: Number
    }
  }, { timestamps: true });
```
Test fails with this schema because of 2 different issues. One of them is caused by the same code that causes #607, and so it should get fixed by #689. The other one is the original issue - #382, looks like it's still there.

PR #386, opened 2018-07-15
Fix data corruption when storing binary data

peecky:
Hello, I fixed a bug of storing corrupted binary data.

For example, following code did not properly store the value.

```javascript
    key: { type: String },
    value: { type: Buffer }
});
const Model = dynamoose.model('KVS', schema);

const s = Buffer.from([0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x13, 0xd3, 0x61, 0x60, 0x60]);
console.log(s); // <Buffer 1f 8b 08 00 00 00 00 00 00 13 d3 61 60 60>

await Model.batchPut([{ key: 'key1', value: s }]);
const c = (await Model.get({ key: 'key1' })).value;
console.log(c); // <Buffer 1f ef bf bd 08 00 00 00 00 00 00 13 ef bf bd 61 60 60>
```

Because not every binary data can be converted to UTF-8 string, also the AWS SDK supports both `Buffer` and `string` for type `B`, it is better not to convert the value to string when the type is `B` and the value is `Buffer` instance.

### DIFF
--- a/test/Model.js
+++ b/test/Model.js
@@ -18,6 +18,10 @@ var Cats = {};
 var ONE_YEAR = 365*24*60*60; // 1 years in seconds
 var NINE_YEARS = 9*ONE_YEAR; // 9 years in seconds
 
+var imageData = Buffer.from([0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x13, 0xd3, 0x61, 0x60, 0x60]);
+imageData.should.not.eql(Buffer.from(imageData.toString())); // The binary value should not be UTF-8 string for test.
+
+
 describe('Model', function (){
   this.timeout(15000);
   before(function(done) {
@@ -74,6 +78,7 @@ describe('Model', function (){
         vet:{name:'theVet', address:'12 somewhere'},
         ears:[{name:'left'}, {name:'right'}],
         legs: ['front right', 'front left', 'back right', 'back left'],
+        profileImage: imageData,
         more: {fovorites: {food: 'fish'}},
         array: [{one: '1'}],
         validated: 'valid'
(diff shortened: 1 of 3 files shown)


fishcharlie:
@peecky Shouldn't this include some tests as well?

peecky:
@fishcharlie Thanks. :)

test/Model.js
```diff
@@ -18,6 +18,10 @@ var Cats = {};
 var ONE_YEAR = 365*24*60*60; // 1 years in seconds
 var NINE_YEARS = 9*ONE_YEAR; // 9 years in seconds
 
+var imageData = Buffer.from([0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x13, 0xd3, 0x61, 0x60, 0x60]);
```
fishcharlie: These two lines should be included within a test and not globally scoped.

test/Model.js
```diff
@@ -319,6 +325,7 @@ describe('Model', function (){
       model.should.have.property('id', 1);
       model.should.have.property('name', 'Fluffy');
       model.should.have.property('vet', { address: '12 somewhere', name: 'theVet' });
+      model.should.have.property('profileImage', imageData);
```
fishcharlie: This whole section should probably be included in a separate test, not this one.

fishcharlie:
@peecky Thanks for this! Just submitted a review with some more changes :)

fishcharlie:
@peecky This looks good to me. My only other concern is that this could be a breaking change to some users. What are your thoughts on this?

peecky:
I can’t imagine any cases that the current version is useful or works well but new version breaks the working. 

Strictly speaking, however, it is a breaking change. It might be better to apply this change to the next major version. The bug is still avoidable on the current version by setting `toDynamo`.

```javascript
    key: { type: String },
    value: { type: Buffer, toDynamo: x => ({ B: x }) }
});
```

fishcharlie:
@peecky We will get this in for version 1.0. Thanks!

Issue #119, opened 2017-03-23
[Feature Request] Ability to provide specify format or generation of timestamps (ISO8601, specifically)


rex:
I looked through all the code involved in timestamps and the issues involving timestamps and was a little shocked to learn that there is no way to specify the format of the timestamps. The schema defaults to `Date.now` but it would be absolutely incredible if we could specify something different, like `Date.toISOString`. A long string of numbers has no at-a-glance value whatsoever unless you are armed with a conversion calculator. Having ISO8601-compliant timestamps would add tremendous at-a-glance value to the data and eliminate the need for conversion to/from that format when interacting with the database.

I am interested to hear your thoughts, maybe there is something I missed in the process of that feature getting added!

**Edit:** Just wanted to let the record show that I'm aware how trivial the conversion of Unix timestamps to ISO8601 compliant timestamps is (see [related SO post](http://stackoverflow.com/a/12868272/439499)) but I am talking here about avoiding the whole thing in principle by moving the responsibility of that conversion to the ORM. 

**Edit 2:** I'm struggling to find a clean way of doing this and my options are pretty much zero. Tried an instance method, it strictly adheres to the type `Date`. Looked at virtual properties, zero documentation. The words `method`, `static`, and `virtual` are completely absent from the entire readme. So, this is actually becoming a cry for help.

brandongoode:
First, sorry for the slow response (and sorry about the poor readme).

I think the ability to configure how dates are written and read would be a great add.  I originally chose the timestamp for its compactness and ease of sorting.  But it is a real pain to have to convert it every time I want to check something manually (I know I've done it a lot). 

I'll look into the option to configure it, but I am swamped and know it will be some time before I can do it. 

In the meantime, you could use the `set` and `get` options for an attribute to control how it is written to the DB.  

Something like:

```
  myDataAttr: {
    type: String,
    set: function(date) {
      if(date && date.toISOString) {
        return date.toISOString();
      }
    },
    get: function (val) {
      return new Date(val);
    }
  }
```

rex:
Hello @brandongoode! No worries on either count, I got things worked out and ended up just reading through the pertinent parts of the code (which, incidentally, is a pretty great substitute for a readme).

In a stroke of luck on my end, the current use case I (and my team) have for dynamoose ended up just being a single field `updatedAt`, which I defined as follows:

```javascript
updatedAt: {
    type: String,
    default: Date.toISOString
}
```

[...]

PR #402, opened 2018-08-02
Customize timestamp format

chunyenHuang:
#119 

Timestamp is good for index and query, but it's not really readable from database. In this PR, I provide the ability to overwrite the timestamp with Date String.

### DIFF
--- a/docs/_docs/schema.md
+++ b/docs/_docs/schema.md
@@ -239,7 +239,11 @@ var schema = new Schema({...}, {
   throughput: 5,
   timestamps: {
     createdAt: 'creationDate',
-    updatedAt: 'lastUpdateDate'
+    updatedAt: 'lastUpdateDate',
+    type: String, // default: Number for Date.now
+    default: function(){
+      return new Date().toLocaleString();
+    }
   }
 });
 ```
--- a/lib/Schema.js
+++ b/lib/Schema.js
@@ -40,6 +40,7 @@ function Schema(obj, options) {
   if (this.options.timestamps) {
     var createdAt = null;
     var updatedAt = null;
+    var timestampDefaultFunciton = Date.now;
 
     if (this.options.timestamps === true) {
       createdAt = 'createdAt';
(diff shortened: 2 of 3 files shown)


chunyenHuang:
cool, so based on the conversation in https://github.com/dynamoosejs/dynamoose/issues/382#issuecomment-405783548

If I understand correctly, the end goal will be 
```
{
    timestamps: {
        createdAt: {
            rangeKey: true, // this should add the index as well
            name: "createdAt",
            type: String,
            default: function (){ 
                return new Date().toLocaleString();
            }
        },
        updatedAt: true
    }
}
```
However, I think this just make the `timestamp` away from the schema. What about we just allow the user to overwrite the timestamp format through schema?

for example, 
```
    new Schema({
         started_at: {
            rangeKey: true, // this should add the index as well
            type: String,
            default: function (){ 
                return new Date().toLocaleString();
            }
        },
    }, {
        timestamp: {
            createdAt: 'started_at'
        }
    });
```

In `Schema.js`
```
    obj[createdAt] = obj[createdAt] || {};
    obj[createdAt].type = obj[createdAt].type || timestampType;
    obj[createdAt].default = obj[createdAt].default || timestampDefaultFunciton;
```

fishcharlie:
@chunyenHuang Sure that works. So long as in the schema none of the properties are required. For example if you don't set a type or default, it should just work as normal.

Maybe use a spread operator to set the default values and spread over the user defined options?

Only other concern is I think other parts of the application rely on the timestamps, so ensuring those work as expected is important too I think.

dolsem:
@chunyenHuang I'm very much interested in this as well, cause I'm getting the same error when I try to get an item from the database. It seems like what's failing is the test case you created for your feature/bug fix. After a quick look at it, might I suggest that you move your test code inside the already existing test 'Schema with timestamps options that are rangeKey'? Unless I'm missing something, I think it would make sense for this test to fail, because currently having timestamp as a rangeKey doesn't properly work. I hope you can get this pulled soon!

fishcharlie:

mgwidmann:
Is this feature possible? I can't seem to find it in the documentation or issues anywhere...

fishcharlie:
@mgwidmann This isn't implemented currently. Feel free to create a PR. You could also do this manually as part of your schema. The `Date` type in Dynamoose has a few different options for how to store the underlying date. Would be cool to add more options to that in the future, but haven't gotten around to that either yet.

PR #419, opened 2018-09-04
Change QueryKey and ScanKey type to any

cakedevp:
The method startAt requires an object with the same format as the one given by the lastKey value not a String

### Summary:
The type definition is incorrect for the types QueryKey and ScanKey since these two types are used for the startAt function.

The startAt key expects an object with the same format as the one given by the lastKey field when executing a query or scan and currently it was asking for a String which is causing a problem when trying to send the value.

There was a workaround for this using:

```
myquery.startAt({ id: { S: "someid" } } as any).exec();
```

But that's just not OK.

### Code sample:
This should work now:

```
myquery.startAt({ id: { S: "someid" } }).exec();
```

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/dynamoose.d.ts
+++ b/dynamoose.d.ts
@@ -252,7 +252,7 @@ declare module "dynamoose" {
   export interface QueryResult<T> extends Array<T> {
     lastKey?: QueryKey;
   }
-  type QueryKey = string;
+  type QueryKey = any;
 
 
   /**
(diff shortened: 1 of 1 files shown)


fishcharlie:
@cakedevp Thanks for this!

PR #420, opened 2018-09-05
Updates Expires attribute on Model Updates

devotox:
### Summary:
Allows the Expires attribute to be updated everytime the model updates.
This makes sense because you would want something to expire after it has not been touched for a certain amount of time not just from the time it was created

### Type (select 1):
- [ ] Bug fix
- [x] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:


fishcharlie:
@devotox Thanks for this! I think we do want to add documentation for this PR as well as write some tests for this.

I think we also need to confirm if this is a breaking change or not. We want to keep the existing behavior for the defaults as is I believe for this, and allow the user to pass in an option or something to overwrite the behavior.

devotox:
Documentation updated & default is false so should not be a breaking change

lib/Model.js
```diff
@@ -539,6 +539,11 @@ Model.update = function(NewModel, key, update, options, next) {
     options.updateTimestamps = true;
   }
 
+  // default updateExpires to true
```
fishcharlie: This should be `false`, correct? In this comment? Although it's just a comment, probably a good idea to ensure it is correct.

fishcharlie:
@devotox Great work on this!! I think this will be a very welcome feature. Just left a review with a few comments.

Have you confirmed that this feature is not already built into Dynamoose? I'm pretty confident it has not been built into Dynamoose yet, but I just want to verify.

fishcharlie:
@devotox This will be a very welcome feature and I know personally I will use it very extensively. So thank you for this!!

devotox:
@fishcharlie yes this has not been built into Dynamoose yet. Looking to get some time to write the tests. tried to quickly throw some together based on updateTimestamps but alas it failed lol. So gotta sit down and see why

devotox:
@fishcharlie Ive added tests but i noticed you are not testing updateTimestamps options in the schema tests. Shouldn't that be done and then also updateExpires.

At the moment you only test this on Model.save

fishcharlie:
@devotox If you have a strong argument for changing the default/current behavior feel free to make that argument. I'm all for changing default behaviors if it is justified. It also has to take into account the fact that upgrading Dynamoose then will cause problems for people who are relying on the current behaviors.

If we draw the conclusion that we should change the current behavior this will need to be part of v1.0 probably. I will try to hold off on pushing v1.0 until we determine that, but I'm in the very final stages of that deployment, and a lot of my projects are gonna rely on v1.0, so we do kinda need to figure this out. If that doesn't make it into v1.0 we can always bring it up again for v2.0.

lib/Model.js
```diff
@@ -298,11 +298,19 @@ Model.prototype.put = function(options, next) {
       options.overwrite = true;
     }
 
-    let toDynamoOptions = {updateTimestamps: true};
+    let toDynamoOptions = {
+      updateExpires: false,
```
fishcharlie: We don't need this, correct? It should just default to false by default without having this line.

lib/Model.js
```diff
@@ -539,6 +547,11 @@ Model.update = function(NewModel, key, update, options, next) {
     options.updateTimestamps = true;
   }
 
+  // default updateExpires to false
+  if (typeof options.updateExpires === 'undefined') {
+    options.updateExpires = false;
```
fishcharlie: Once again I don't think we need this section. 'undefined' is considered falsey so I don't think we need this part since the default will be undefined or false.

test/Model.js
```diff
@@ -493,6 +493,64 @@ describe('Model', function (){
...
+                });
+              });
+            }, 1000);
+          });
+        });
+
+
+        it('Save existing item without updating expires', function (done) {
+          var myCat = new Cats.Cat10({
+            id: 1,
```
fishcharlie: Let's make sure this `id` is unique to ensure we aren't getting conflicts with the test above.

test/Model.js
```diff
@@ -493,6 +493,64 @@ describe('Model', function (){
...
+
+          myCat.save(function(err, theSavedCat1) {
+            var expectedExpires = theSavedCat1.expires;
+
+            myCat.name = "FluffyB";
+            setTimeout(function() {
+              myCat.save({updateExpires: false}, function () {
+                Cats.Cat10.get(1, function(err, realCat) {
+                  realCat.name.should.eql("FluffyB");
+                  realCat.expires.should.eql(expectedExpires); // expires should be different than before
```
fishcharlie: Shouldn't the comment read `expires should be the same as before`? Or something along those lines?

lib/Model.js
devotox: I understand your point but isn't it better to have the default set so it is visible through the system that it is actually set to something?

fishcharlie: @devotox I'm pretty sure the only reason to have it there is to increase readability and ease of understanding for developers in the future. Anytime we are comparing that value to `false` or `true`, `undefined` will behave just like `false`. So I don't think removing this is a bad thing. Especially since it looks like you did and all the tests are passing. If you believe there is an edge case that your tests haven't accounted for that this would break please write a test and if that test fails then yes we should totally add it back in.

fishcharlie:
@devotox This looks really good. Do you think we are ready to get this merged in or is there still more work to do?

devotox:
@fishcharlie I just need to write the update tests. Yesterday when I tried couldn't get access to the updated expires attribute to compare them but will make another attempt this morning.

fishcharlie:
@devotox Let me know if I can help at all!

fishcharlie:
@devotox Looks like there are some merge conflicts as well. This is the last PR I'd like to get in before version 1.0. But if we can't get it done before version 1.0 we can hold off till version 1.1. I'm aiming on getting version 1.0 out hopefully this week or this weekend.

devotox:
Haaa looks like the merge broke something .... gonna look into this

fishcharlie:
@devotox Sounds good! And you still have one more test to write, correct?

devotox:
@fishcharlie Test added. Probably too late for the Release but here it is

devotox:
@fishcharlie Yes, i believe it can be merged. And I keep trying to think of a good argument or find something as proof but maybe I'm wrong. It just feels like that makes sense. "A document is expired a certain period from its last update"

devotox:
@fishcharlie You make a lot of sense with that. It should stay as close to the AWS SDK in that sense!

PR #422, opened 2018-09-08
Fix bad deserialization

### LINKED ISSUE #421: Incorrect response for Object type
### Summary:

Have a schema entry :  
```PayableAmount: {type: Map}```
have also tried
``` PayableAmount: {type: Object},```

#### Schema
```
const schema = {
Account: {type: String, hashKey: true}
IssueDate: {type: String}
PayableAmount: {type: Object}
}
```

#### Model
Model has no extra behavior apart from having a schema

#### General
```
Model.query({
      Account: {eq:"someID"},
      IssueDate: { le: "Date1", ge: "Date2},
    }).exect(callbackFunc);
```

Query works when I look in debugger and I get the expected return value from dynamo.  However, parsing the result for PayableAmount (type M) tries to read it as a type S and throws the following error:

SyntaxError: Unexpected token o in JSON at position 1
    at parse (<anonymous>)
    at dedynamofy (/development/server/node_modules/dynamoose/lib/Attribute.js:470:14)
    at Attribute.parseDynamo (/development/server/node_modules/dynamoose/lib/Attribute.js:540:13)
    at Schema.parseDynamo (/development/server/node_modules/dynamoose/lib/Schema.js:189:24)
    at toModel (/development/server/node_modules/dynamoose/lib/Query.js:240:18)
   

Should return the Map/Object

### Environment:

Operating System: OSX  
Operating System Version: 10.13.6
Node.js version (`node -v`): v10.8.0
NPM version: (`npm -v`): 6.4.0
Dynamoose version: "0.8.7"

### Type (select 1):
- [ ] Bug report
- [ ] Feature suggestion
- [ ] Question
- [ ] Other suggestion

### Other:

### PR DESCRIPTION
**Pull request by brennhill**
### Summary:
Fixes issue where a MAP is returned for an object type and we try to deserialize it as if it were a JSON string.  If we get a map back (like if it was saved by another process than dynamoose, which happens with microservices and the like), then we should respect that we have an object instead of a string and return the object.

Otherwise this can cause breakages for no good reason, where we receive an object under the circumstances where an object is expected, yet breaks anyway because we try to treat an object as a string, despite Dynamo helpfully letting us know what the type is ahead of time.

### GitHub linked issue:
Closes #421 

### Type (select 1):
- [X] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [X] No
- [ ] I'm not sure

- [X] Yes
- [ ] No
Up to you.

- [X] Yes
- [ ] No

### Other:
- [X] I have run `npm test` from the root of the project directory to ensure all tests continue to pass
- [X] I have confirmed that all my code changes are indented properly using 2 spaces

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -560,7 +560,11 @@ Attribute.prototype.parseDynamo = function(json) {
       }
       break;
     case 'array':
-    val = dedynamofy('S', this.isSet, json, JSON.parse, this);
+      if(json.L) {
+        val = dedynamofy('L', this.isSet, json, listify, this);
+      } else {
+        val = dedynamofy('S', this.isSet, json, JSON.parse);
+      }
     break;
     case 'map':
     val = dedynamofy('M', this.isSet, json, mapify, this);
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -553,8 +553,12 @@ Attribute.prototype.parseDynamo = function(json) {
     val = dedynamofy('N', this.isSet, json, datify, this);
     break;
     case 'object':
-    val = dedynamofy('S', this.isSet, json, JSON.parse, this);
-    break;
+      if(json.M) {
+        val = dedynamofy('M', this.isSet, json, mapify, this);
+      } else {
+        val = dedynamofy('S', this.isSet, json, JSON.parse);
+      }
+      break;
     case 'array':
     val = dedynamofy('S', this.isSet, json, JSON.parse, this);
     break;
--- a/test/MicroService.js
+++ b/test/MicroService.js
@@ -1,0 +1,140 @@
+'use strict';
+
+
+var dynamoose = require('../');
+
+dynamoose.AWS.config.update({
+  accessKeyId: 'AKID',
+  secretAccessKey: 'SECRET',
+  region: 'us-east-1',
+});
+dynamoose.local();
+var Schema = dynamoose.Schema;
+
+var should = require('should');
+var tableName = 'MicroDog';
+
+describe('Inner Map handling', function (){
+  before(function (done) {
+
+    function hookupDynamoose() {
+      dynamoose.setDefaults({ prefix: '', suffix: '' });
+
+      var microDogSchema  = new Schema({
+        ownerId: {
+          type: Number,
+          validate: function(v) { return v > 0; },
+          hashKey: true,
+        },
+        birthData: {
+          type: Object
+        },
+        breed: {
+          type: String,
+          required: true,
+          index: {
+            global: true,
+            rangeKey: 'ownerId',
+            name: 'BreedIndex',
+            project: true, // ProjectionType: ALL
+            throughput: 5 // read and write are both 5
+          }
+        },
+        name: {
+          type: String,
+          rangeKey: true,
+          index: true // name: nameLocalIndex, ProjectionType: ALL
+        },
+      });
+      dynamoose.model(tableName, microDogSchema);
+      done();
+    }
+
+    var microDog = {
+      Item: {
+        ownerId:{N:'1'},
+        birthData: {M:
+            {kennel:{S:"MyKennel"}, state:{S:"NC"}}
+        },
+        name: {'S':'Foxy Lady'},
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/Attribute.js [hunk 1]
```diff
@@ -553,8 +553,12 @@ Attribute.prototype.parseDynamo = function(json) {
     val = dedynamofy('N', this.isSet, json, datify, this);
     break;
     case 'object':
-    val = dedynamofy('S', this.isSet, json, JSON.parse, this);
-    break;
+      if(json.M) {
+        val = dedynamofy('M', this.isSet, json, mapify, this);
+      } else {
+        val = dedynamofy('S', this.isSet, json, JSON.parse);
```
fishcharlie: Is there a reason why the line that was removed included `this` as the last parameter but not this new version?

brennhill: 100% of tests pass with this change. I can guarantee some code fails without it since that's how I found it.  Let me see what I can do to prove it.

fishcharlie: @brennhill Yeah I'm just looking for new tests to be added. So in case this bug slips in again our tests will catch it before hand. If that makes sense.

brennhill: I went and looked at adding a test and it's going to take a lot of work to reproduce because dynamo stores objects as JSON strings intead of MAP.  Which means i'd have to fix how dynamo stores data, which would be a beaking change.  

As stated in my decription, this is a fix for circumstances where a process *other* than dynamoose is doing the saves to dynamo and is correctly saving Objects as Maps insted of JSON strings.

fishcharlie:
It’s still possible to write a test for this. You can use the AWS SDK to save an item directly to DynamoDB then replicate the issue. I think we are doing this somewhere else but I could be wrong.

I’m not able to look into this further right now but will try to make time to look into this further this weekend.

Based on the issue this is linked to tho I don’t think I have enough sample code to go on to write a test for this. So anymore code you can provide to replicate and test the issue would be very helpful. End to end would be best.

brennhill:
@fishcharlie I've gone ahead and done that and reproduced, I'll push a working test that will fail if you remove my code.

brennhill:
I've created some code to show the issue which is on my base branch.  However, the tests are SUPER brittle and everything breaks kinda randomly even though I've commented out the actual test and so on. In fact, the tests pass/fail somewhat randomly. I just can't fix the entire thing and I'm not even sure why this stuff is breaking at a distance.

@fishcharlie please take a look here: https://github.com/brennhill/dynamoose/commit/78abb16e264159c3af5ac32b9ea41943d3741206

You will see a simplified Dog schema (microdog) and a sample object.  You'll also see the dynamoDB table creation.  In this case, the 'birthData' type is an Object, and an Object (map) is saved to DynamoDB.

If Dynamoose reads this entry, even with Object type set in schema, it will FAIL to deseriealize the object because it expects a JSON string.  But a JSON string is NOT guaranteed.

brennhill:
Advice on how to get this working without the other tests breaking appreciated, I really don't know why this breaks.

brennhill:
@fishcharlie I found an identical issue with list handling, I've gone ahead and pushed the fix.  Arrays can return as type L and they also should be inspected to see if we already got an object back vs a JSON string.

fishcharlie:
@brennhill Awesome. I will try to get around to fixing up the tests as soon as I can. Do you have mock data for the `L` type that you can use to write another test that I can modify into the test structure as well? I just want to ensure we are testing as much as possible and as many edge cases as possible.

brennhill:
@fishcharlie I do not.  In my case it choked on even an empty list - []

BTW, I'm trying to query by ISO8601 dates (2018-12-01) and it's not working.  I've set local index on the date field.  Im trying to figure out if I have to specify index somehow in the query

fishcharlie:
@brennhill That sounds like a separate issue. If you could create a GitHub issue for that that'd be great and I can try to help further with that. I just like to keep things organized so that in case someone else has a similar problem they can easily find solutions.

fishcharlie:
Closing based on comment in #421

Issue #434, opened 2018-10-01
Maximum call stack size exceeded with 1.0.0


ThomasVictoria:
### Summary:

Hello, I updated Dynamoose to the latest version you pushed (1.0.0), and I'm getting an `Maximum call stack size exceeded` Error when I run a get on my model. I had to downgrade it to 0.8.7 to make it works like before. Also it looks like it depend of the power of the computer, some of us why a recent computer have been able to get a response but older and less powerful one were have the stack error.

#### Model
```
this.m = dynamoose.model(db_name, {
  id: {
    type: String,
    hashKey: true
  },
  name: String,
  retry: Number,
  module: Map,
  }, {useDocumentClient: true})
```

### General
```
let dynamoModule = new DynamoModule(bot.id)

let _module = await dynamoModule.m.get({ id: "intro" }).then(res => {
	return res
})
```
It get stuck there

### Current output and behavior:

`Maximum call stack size exceeded`

### Expected output and behavior:

I get my data and it works

### Environment:

Operating System: OSX
Operating System Version: 10.13.1
Node.js version (`node -v`): 8.11.2
NPM version: (`npm -v`): 5.6.0
Dynamoose version: 1.0.0

### Type (select 1):
- [x] Bug report
- [ ] Feature suggestion
- [ ] Question
- [ ] Other suggestion
- [ ] Something not listed here

### Other:
- [x] I have read through the Dynamoose documentation before posting this issue
- [x] I have searched through the GitHub issues (including closed issues) and pull requests to ensure this issue has not already been raised before
- [x] I have searched the internet and Stack Overflow to ensure this issue hasn't been raised or answered before
- [x] I have tested the code provided and am confident it doesn't work as intended
- [x] I have filled out all fields above
- [x] I am running the latest version of Dynamoose

fishcharlie:
@ThomasVictoria So this does work on 0.8.7 but not 1.0.0, correct? And you made no code changes between it working and not working?

The first thing that stands out to me, is I don't think you want a `.then` with an `await`, especially since your `.then` isn't really doing anything.

Can you try enabling the debug toggle and see if that gives any more useful information? You can do this by running `export DEBUG=dynamoose*` and reproducing the error. That will print a **lot** more information to the console. Posting the last chunk of that before the error might be really helpful.

Kinda sounds like it's getting into some type of loop that it shouldn't be.

fishcharlie:
[...]

Issue #343, opened 2018-04-20
Table index not selected correctly when querying


ricxsar:
I have a table `ScheduleLog` with 2 global indexes using the same primary key `orgId` and different range keys `createdAt`, `expectedArriveAt` . See schema below:
```
const schema = new Schema({
	id: {
		type    : String,
		hashKey : true,
		default : uuid.v1,
		required: true
	},
	orgId: {
		type    : String,
		index: [
		{
			global    : true,
			name      : 'OrganizationCreateAtIndex',
			rangeKey  : 'createdAt',
			throughput: 1
	       },
	       {
			global    : true,
			name      : 'OrganizationExpectedArriveAtIndex',
			rangeKey  : 'expectedArriveAt',
			throughput: 1
	       }],		
	       required: true,
	},
	assetId: {
		type    : String,
		required: true,
		index: {
			global    : true,
			name      : 'AssetIndex',
			rangeKey: 'createdAt',
			throughput: 1
	        },
	},
	scheduleId: {
		type    : String,
		required: true
	},
	status: {
		type    : String,
		default	: "QUEUED",
		required: true
	},
	origin: {
		type    : Object,
		required: true	
	},
	destination: {
		type    : Object,
		required: true
	},
	distance: {
		type    : Number
	},
	expectedDepartAt: {
		type    : Date,
		default : Date.now,
		required: true
	},
	actualDepartAt: {
		type    : Date
	},
	expectedArriveAt: {
		type    : Date,
		default : Date.now,
		required: true	
	},
	actualArriveAt: {
		type    : Date
	},
	time_minutes: {	// minutes
		type    : Number,
	},
	tracepoints: {
		type    : [String],
		default : [],
	},
	remarks: {
		type    : String
	}
},{
	throughput: 1,
	timestamps: true
});
```

When I query the table using the primary key `orgId` and range key `expectedArriveAt`, the `OrganizationCreateAtIndex` is used not the `OrganizationExpectedArriveAtIndex`. See complete query 
 and error below:
```
ScheduleLog.query('orgId').eq("b76e5930-5c79-11e7-917a-fd1a3417d363")
.where('expectedArriveAt').lt( new Date() )
.exec()
.then(function(res){
	console.log(res);
})
.catch(function(e){
	console.log(e)
})
```
```
{ [ValidationException: Query condition missed key schema element: createdAt]
  message: 'Query condition missed key schema element: createdAt',
  code: 'ValidationException',
[...]

PR #442, opened 2018-10-15
Allow querying local indexes and resolve problem with using updatedAt as range key in a query.

### LINKED ISSUE #439: Defining and querying a local index.
### Summary:
I have a schema for posts with timestamps that uses createdAt as the range key and updatedAt as a local index. The idea is to be able to query posts by a particular author by both createdAt and updatedAt.

#### Schema
```
const PostSchema = new dynamoose.Schema({
  author: {
    type: String,
    hashKey: true,
  },
  createdAt: {
    type: Number,
    rangeKey: true,
    default: Date.now,
  },
  updatedAt: {
    type: Number,
    default: Date.now,
    index: true,
  },
  text: String,
  deletedAt: {
    type: Number,
    index: {
      global: true,
    },
  },
}, {
  timestamps: true,
});
```

#### General
```
Post.query('author').eq(author).where('createdAt').lt(startAt).exec((err, posts) => {
// All posts created before the `startAt` timestamp.
}
Post.query('author').eq(author).where('updatedAt').gt(since).exec((err, posts) => {
// All posts updated (or created) after the `since` timestamp.
}
```

First query works as expected. Second query fails with:
```
Query condition missed key schema element: createdAt
```

Both queries work fine.

### Environment:

Operating System: Linux
Operating System Version: 4.15.0-36-generic
Node.js version (`node -v`): v9.11.2
NPM version: (`npm -v`): 6.4.1
Dynamoose version: 1.0.0

I've noticed that local index queries in Query tests in Dynamoose look differently. Instead of querying on the hash key of the table first and then specifying local index as the range key, it queries on the local index directly. This would make sense for a global index, but for local? If I understand DynamoDB correctly, local index allows you to choose an alternative range key for the table, while keeping the hash key. So by querying local index I should be able to choose the partition based on the author's name, then display all posts in that partition, sorted by the updatedAt values. Am I missing / not understanding something?

### Type (select 1):
- [ ] Bug report
- [ ] Feature suggestion
- [x] Question
- [ ] Other suggestion

### Other:

dolsem:
### Summary:

This PR fixes the issue discussed in #439, and one more bug:
`toDynamo` method of Attribute is called without `updateTimestamps` option set to `false` during `Query.exec()`, so when `updatedAt` timestamp is used as a range key in a query, supplied value gets substituted with current timestamp.

I also moved the debug statement for global secondary index, because local secondary index passes that outer check too.

### GitHub linked issue:

Closes #439 

### Other information:

I took the liberty of updating existing tests that were doing nothing meaningful instead of creating new ones. Let me know if you want me to resort to a more conservative option and just define my own tests. However, please note that test `Query with Local Global Index as range` (I assumed there's a typo in the name) currently queries a global index, not a local one, while test `Should allow multiple local indexes and query correctly` in its current state should fail as it doesn't even make sense conceptually (local indexes can't have their own range keys and it doesn't make sense for the same attribute to be the key of multiple local indexes). Moreover, the schema defined in that test most likely ends up being unused (which might be why it doesn't fail), because instead of creating a new table it tries to overwrite the one created by the previous test (`Log-1`) and likely fails silently, because you cannot add / remove local indexes for an existing table, you have to create a new one.

I marked this as a breaking change, however, if we assume that no one used the fact that updatedAt value was changed during a query as a "feature", it shouldn't break any code.

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [x] Testing improvement

- [x] 🚨 YES 🚨
- [ ] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/lib/Query.js
+++ b/lib/Query.js
@@ -88,10 +88,10 @@ Query.prototype.exec = function (next) {
   // the hash and range key on the primary table.  If they don't match then we
   // can look for a secondary index to query.
   if(schema.hashKey.name !== this.query.hashKey.name || (this.query.rangeKey && schema.rangeKey && schema.rangeKey.name !== this.query.rangeKey.name)) {
-    debug('query is on global secondary index');
     for(indexName in schema.indexes.global) {
       index = schema.indexes.global[indexName];
       if(index.name === this.query.hashKey.name && (!this.query.rangeKey || index.indexes[indexName].rangeKey === this.query.rangeKey.name)) {
+        debug('query is on global secondary index');
         debug('using index', indexName);
         queryReq.IndexName = indexName;
         break;
@@ -116,7 +116,7 @@ Query.prototype.exec = function (next) {
       debug('query is on local secondary index');
       for(indexName in schema.indexes.local) {
         index = schema.indexes.local[indexName];
-        if(index.name === rangeKey.name && index.indexes[indexName].rangeKey === this.query.rangeKey.name) {
+        if(index.name === rangeKey.name) {
           debug('using local index', indexName);
           queryReq.IndexName = indexName;
           break;
--- a/lib/Query.js
+++ b/lib/Query.js
@@ -135,7 +135,7 @@ Query.prototype.exec = function (next) {
       for (i = 0; i < rangeKey.values.length; i++) {
         val = rangeKey.values [i];
         keyConditions.AttributeValueList.push(
-          rangeAttr.toDynamo(val, true)
+          rangeAttr.toDynamo(val, true, Model, { updateTimestamps: false })
         );
       }
     }
--- a/test/Query.js
+++ b/test/Query.js
@@ -678,56 +678,50 @@ describe('Query', function (){
   	});
   });
 
-  it('Should allow multiple local indexes and query correctly', function (done) {
+  it('Should allow multiple local indexes and query correctly', async function () {
     var schema = new dynamoose.Schema({
       id: {
         type: String,
         hashKey: true,
-        required: true
       },
       orgId: {
         type: String,
-        index: [{
-          global    : false,
-          name      : 'OrganizationCreateAtIndex',
-          rangeKey  : 'createdAt',
-          throughput: 1
-      }, {
-          global    : false,
-          name      : 'OrganizationExpectedArriveAtIndex',
-          rangeKey  : 'expectedArriveAt',
-          throughput: 1
-      }],
-      required: true,
+        rangeKey: true,
       },
-      expectedArriveAt: Date
+      updatedAt: {
+        type: Date,
+        index: {
+          global: false,
+          name  : 'OrganizationUpdatedAtIndex'
+        }
+      },
+      expectedArriveAt: {
+        type: Date,
+        index: {
+          global: false,
+          name  : 'OrganizationExpectedArriveAtIndex'
+        }
+      }
     },{
       throughput: 1,
       timestamps: true
     });
-    var Log = dynamoose.model('Log-1', schema);
+    var Log = dynamoose.model('Log-2', schema);
 
     var log1 = new Log({id: "test1", orgId: "org1", expectedArriveAt: Date.now()});
-    log1.save(function() {
-      Log.query('orgId').eq("org1")
+    var log2 = new Log({id: "test1", orgId: "org2", expectedArriveAt: Date.now()});
+
+    await log1.save();
+    await log2.save();
+
+    var res = await Log.query('id').eq("test1")
       .where('expectedArriveAt').lt( new Date() )
       .exec()
-      .then(function(res){
-        res.length.should.eql(1);
-        Log.query('orgId').eq("org1")
-        .where('createdAt').lt( new Date() )
-        .exec()
-        .then(function(res){
-          res.length.should.eql(1);
-          done();
-        })
-        .catch(function(e){
-          done(e);
-        });
-      })
-      .catch(function(e){
-        done(e);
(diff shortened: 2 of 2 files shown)


lib/Query.js
```diff
@@ -135,7 +135,7 @@ Query.prototype.exec = function (next) {
       for (i = 0; i < rangeKey.values.length; i++) {
         val = rangeKey.values [i];
         keyConditions.AttributeValueList.push(
-          rangeAttr.toDynamo(val, true)
+          rangeAttr.toDynamo(val, true, Model, { updateTimestamps: false })
```
fishcharlie: Are we sure that ` { updateTimestamps: false }` is a valid option here? I'm assuming it is, but just want to verify.

test/Query.js
```diff
@@ -674,56 +678,50 @@ describe('Query', function (){
...
+  it('Should allow multiple local indexes and query correctly', async function () {
     var schema = new dynamoose.Schema({
       id: {
         type: String,
         hashKey: true,
-        required: true
       },
       orgId: {
         type: String,
-        index: [{
```
fishcharlie: Are we sure we want to remove these indexes? That was from issue #343.

test/Query.js
```diff
@@ -674,56 +678,50 @@ describe('Query', function (){
...
 
     var log1 = new Log({id: "test1", orgId: "org1", expectedArriveAt: Date.now()});
-    log1.save(function() {
-      Log.query('orgId').eq("org1")
+    var log2 = new Log({id: "test1", orgId: "org2", expectedArriveAt: Date.now()});
+
+    await log1.save();
+    await log2.save();
+
+    var res = await Log.query('id').eq("test1")
```
fishcharlie: Same thing here. Before it was `.query('orgId').eq("org1")`. I'm considered that this isn't testing #343 anymore.

test/Query.js
```diff
@@ -674,56 +678,50 @@ describe('Query', function (){
...
-          done(e);
-        });
-      })
-      .catch(function(e){
-        done(e);
-      });
-    });
+    res.length.should.eql(2);
+    
+    var res2 = await Log.query('id').eq("test1")
```
fishcharlie: And finally, same thing here related to previous comments.

fishcharlie:
@yokkoyokko Mind taking a look at this and submitting a GitHub review?

fishcharlie:
@dolsem In terms of the following.

> I marked this as a breaking change, however, if we assume that no one used the fact that updatedAt value was changed during a query as a "feature", it shouldn't break any code.

I don't think this is an issue or a breaking change. Did the `updatedAt` value really change during a query??? 😬 That is kinda bad. So yes, if that was happening and that is the only breaking change, I consider that a **major** bug. `updatedAt` to me, means that the item was actually updated and changed.

If you could, it'd be amazing if you could write a test to ensure that `updatedAt` won't be updated on a query, and if you want even a scan or get for that matter. I think based on what you said, that test should fail on the current version, and should pass on this PR. Just to ensure we are covering that so we don't break it in the future.

---

I **really** appreciate your work on this. Thank you **so** much!!!

lib/Query.js
dolsem: Yes, because this line is for converting the value of a range key in a query, `.where(rangeKey).gt(thisValue)`, to DynamoDB notation, and we really don't want to update it as we convert it.

dolsem:
@fishcharlie This last test you had questions about already checks for that (perhaps somewhat implicitly), that's why I use updatedAt as a local index. The test currently fails. If you change that to createdAt (which doesn't get updated with `updateTimestamps` set to `true`), nothing should change, as their values in this case are the same and set during creation, but the test will then pass. I could add an expectation that explicitly checks that the value of updatedAt stays the same after the query is formed, but I haven't seen a way to get the actual DynamoDB query after it gets generated by `Query.exec()` (other than output to debug), so I am not sure about the best way to go about it.

fishcharlie:
@dolsem Got it. Looks like there are still some comments from my review that I don't think have been addressed yet.

test/Query.js
dolsem: I took a look at #343 before making my changes. It's about querying multiple global indexes. There were two tests written for that issue though: one with global indexes, one with local. I didn't make changes to the one with global indexes, so #343 is still getting tested. As I explained above, the test that I updated never actually tested what it was supposed to test, instead it was likely acting as a duplicate, because it failed to overwrite the indexes on the table created by that first test. To see what I mean, try changing the name of the model from Log-1 to something else and re-running the test suite. This test will fail.

fishcharlie: @dolsem Got it

fishcharlie:
@dolsem What do you think? Think we are ready to merge in and get a release out?

dolsem:
@fishcharlie yes, I believe so.

fishcharlie:
@dolsem Awesome. I'll get that done and hopefully get a release out before the end of the day. Again thanks so much for your help! Was hoping #434 was going to be further along before we do a release, but looks like that won't happen. Thanks again!

fishcharlie:
@dolsem Thanks again for your work on this. My goal is since version 1.0.0 was such a massive update to provide bug updates ASAP for the next little bit for things that broke in version 1.0.0 that were working before version 1.0.0. Therefor version 1.0.1 has been **released** that includes this fix!!

Please keep filing issues if you notice any other problems and we will work to get them out as soon as possible.

dolsem:
No problem, glad I could help!

PR #443, opened 2018-10-15
Raise error if list or map not provided

dobrynin:
It seems to me that the existing conditional `(value === undefined && value[0] === undefined)` errors if value is indeed undefined. This change fixes that and adds a more descriptive error if the user forgets to provide the `list` key when using `type: 'list'`.

#### Schema
```
// Code here
```

#### Model
```
// Code here
```

#### General
```
// Code here
```

### GitHub linked issue:

#---

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No

### Other:
- [ ] I have run `npm test` from the root of the project directory to ensure all tests continue to pass

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -56,10 +56,11 @@ function Attribute(schema, name, value) {
   else if (this.type.name === 'list'){
 
     if(value.type) {
+      if (!value.list) throw new errors.SchemaError('No list schema provided for attribute:' + this.name );
       value = value.list;
     }
 
-    if (value === undefined && value[0] === undefined){
+    if (value === undefined || value[0] === undefined){
       throw new errors.SchemaError('No object given for attribute:' + this.name );
     }
     // stang: Don't know what this guard is for - had to remove because when parsing unknown attributes, this is legal?
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -42,6 +42,9 @@ function Attribute(schema, name, value) {
   if (this.type.name === 'map'){
 
     if(value.type) {
+      if (!value.map) {
+        throw new errors.SchemaError('No map schema given for attribute:' + this.name );
+      }
       value = value.map;
     }
     for (const subattrName in value){
@@ -56,13 +59,12 @@ function Attribute(schema, name, value) {
   else if (this.type.name === 'list'){
 
     if(value.type) {
-      if (!value.list) throw new errors.SchemaError('No list schema provided for attribute:' + this.name );
+      if (value.list === undefined || value.list[0] === undefined){
+        throw new errors.SchemaError('No list schema given for attribute:' + this.name );
+      }
       value = value.list;
     }
 
-    if (value === undefined || value[0] === undefined){
-      throw new errors.SchemaError('No object given for attribute:' + this.name );
-    }
     // stang: Don't know what this guard is for - had to remove because when parsing unknown attributes, this is legal?
     // if (value.length > 1){
     //   throw new errors.SchemaError('Only one object can be defined as a list type in ' + this.name );
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -43,7 +43,7 @@ function Attribute(schema, name, value) {
 
     if(value.type) {
       if (!value.map) {
-        throw new errors.SchemaError('No map schema given for attribute:' + this.name );
+        throw new errors.SchemaError(`No map schema given for attribute: ${this.name}`);
       }
       value = value.map;
     }
@@ -59,8 +59,8 @@ function Attribute(schema, name, value) {
   else if (this.type.name === 'list'){
 
     if(value.type) {
-      if (value.list === undefined || value.list[0] === undefined){
-        throw new errors.SchemaError('No list schema given for attribute:' + this.name );
+      if (!value.list || !value.list[0]){
+        throw new errors.SchemaError(`No list schema given for attribute: ${this.name}`);
       }
       value = value.list;
     }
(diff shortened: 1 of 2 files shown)


lib/Attribute.js
```diff
@@ -56,10 +56,11 @@ function Attribute(schema, name, value) {
   else if (this.type.name === 'list'){
 
     if(value.type) {
+      if (!value.list) throw new errors.SchemaError('No list schema provided for attribute:' + this.name );
```
fishcharlie: Just for formatting sake, can we change this to the following?

```
if (!value.list){
  throw new errors.SchemaError('No list schema provided for attribute:' + this.name);
}
```

lib/Attribute.js
```diff
@@ -56,10 +56,11 @@ function Attribute(schema, name, value) {
...
       value = value.list;
     }
 
-    if (value === undefined && value[0] === undefined){
+    if (value === undefined || value[0] === undefined){
```
fishcharlie: Since you said this was failing before, can we write a test that will fail in the current version, and pass with this PR to ensure this will work correctly in the future?

dobrynin: Will do.

dobrynin: I'm now running into an issue where if I fetch a record from dynamo that has a field set to an empty array, this error gets raised. Is that the intended behavior?

fishcharlie: @dobrynin Honestly I'm not sure. Are you running into that on the NPM version or only after this change? I believe this code should only be run when creating or setting up the schema/model. So I'm not sure why fetching a record from Dynamo would cause this code to be run unless you have `saveUnknown` set to `true`.

Can you try logging some things out such as `value` and see what that prints?

You can also run `export DEBUG=*` which will give you more console logs and details about what Dynamoose is doing behind the scenes.

dobrynin: You’re thinking was right, we do set `saveUnknown` to `true`.

fishcharlie:
@dobrynin You get a chance to work on this anymore? 😃

dobrynin:
Sorry, I’ve been preoccupied with other work and have neglected this. I’ll get that test going ASAP

fishcharlie:
@dobrynin Are you ready for this to be re-reviewed for the potential of getting it merged in?

dobrynin:
Yes, please do!

lib/Attribute.js
```diff
@@ -56,12 +59,12 @@ function Attribute(schema, name, value) {
   else if (this.type.name === 'list'){
 
     if(value.type) {
+      if (value.list === undefined || value.list[0] === undefined){
```
fishcharlie: Above we use `!value.map`, here we use `value.list === undefined` and `value.list[0] === undefined`. Is there a reason this line isn't just `(!value.list || !value.list[0])`

test/Schema.js
```diff
@@ -1190,4 +1190,46 @@ describe('Schema tests', function (){
...
+  it('Should throw error when type is map but no map is provided', function (done) {
+    try {
+      var schema = new Schema({
+        race: {
+          type: 'map',
+        }
+      });
+    } catch (err) {
+      err.should.be.instanceof(Error);
+      err.should.be.instanceof(errors.SchemaError);
```
fishcharlie: Not sure I'm a huge fan of this. If there is no error thrown, this test will still pass. I think it'd be better to create a variable above (set to nothing by default). Then inside the catch, set `err` to that variable we created above. Then right before `done` is called, run the assertions, to ensure that variable we created exists, ensure it's an instance of, etc.

test/Schema.js
```diff
@@ -1190,4 +1190,46 @@ describe('Schema tests', function (){
...
+    try {
+      var schema = new Schema({
+        race: {
+          type: 'list',
+          list: []
+        }
+      });
+    } catch (err) {
+      err.should.be.instanceof(Error);
+      err.should.be.instanceof(errors.SchemaError);
```
fishcharlie: Same thing here that I mentioned above

test/Schema.js
```diff
@@ -1190,4 +1190,46 @@ describe('Schema tests', function (){
...
+  it('Should throw error when type is list but no list is provided', function (done) {
+    try {
+      var schema = new Schema({
+        race: {
+          type: 'list',
+        }
+      });
+    } catch (err) {
+      err.should.be.instanceof(Error);
+      err.should.be.instanceof(errors.SchemaError);
```
fishcharlie: Same thing here that I mentioned above

lib/Attribute.js
```diff
@@ -56,12 +59,12 @@ function Attribute(schema, name, value) {
...
+        throw new errors.SchemaError('No list schema given for attribute:' + this.name );
```
fishcharlie:
```suggestion
        throw new errors.SchemaError(`No list schema given for attribute: ${this.name}`);
```

Super picky, I know. But PR #465 is doing this, so I think from now on we should do it throughout the code.

lib/Attribute.js
```diff
@@ -42,6 +42,9 @@ function Attribute(schema, name, value) {
   if (this.type.name === 'map'){
 
     if(value.type) {
+      if (!value.map) {
+        throw new errors.SchemaError('No map schema given for attribute:' + this.name );
```
fishcharlie:
```suggestion
        throw new errors.SchemaError(`No map schema given for attribute: ${this.name}`);
```

Super picky, I know. But PR #465 is doing this, so I think from now on we should do it throughout the code.

fishcharlie:
@dobrynin Any progress on this? I'd love to get it into version 1.3.0 if possible (#481), but also not going to hold up the release for this PR.

dobrynin:
@fishcharlie, sorry for the late reply, I've implemented your changes.

PR #456, opened 2018-11-06
Remove Q Dependency

### LINKED ISSUE #216: Support for native promises (async/await)

Aedalus:
### Summary:
This PR removes the dependency on Q while maintaining the current promise and callback based API.

#### Schema
```
// Code here
```

#### Model
```
// Code here
```

#### General
```
// Code here
```

### GitHub linked issue: #216 

Closes #216 

### Type (select 1):
- [ ] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [X] Something not listed here

- [ ] 🚨 YES 🚨
- [X] No
- [ ] I'm not sure

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### Other:
- [X] I have run `npm test` from the root of the project directory to ensure all tests continue to pass
- [X] I have confirmed that all my code changes are indented properly using 2 spaces

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -1,9 +1,5 @@
 'use strict';
 
-process.on('unhandledRejection', (reason, promise) => {
-  console.warn('Unhandled promise rejection:', promise, 'reason:', reason.stack || reason);
-});
-
 const util = require('util');
 const hooks = require('hooks');
 const Attribute = require('./Attribute');
@@ -143,7 +139,7 @@ Model.compile = function compile (name, schema, options, base) {
       return Model.query(NewModel, query, options, next);
     } catch(err) {
       sendErrorToCallback(err, options, next);
-      return Promise.reject(err)
+      return Promise.reject(err);
     }
   };
 
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -277,13 +277,12 @@ function validKeyValue(value) {
 }
 
 Model.prototype.put = function(options = {}, next) {
+  debug('put', this);
 
   if(typeof options === 'function') {
     next = options;
     options = {};
   }
-
-  debug('put', this);
 
   const result = new Promise((resolve, reject) => {
     let model, model$, item;
@@ -1253,6 +1252,7 @@ function batchWriteItems (NewModel, batchRequests) {
 }
 
 Model.batchPut = function(NewModel, items, options, next) {
+  debug('BatchPut %j', items);
 
   options = options || {};
   if(typeof options === 'function') {
@@ -1261,7 +1261,6 @@ Model.batchPut = function(NewModel, items, options, next) {
   }
 
   const result = new Promise((resolve,reject) => {
-    debug('BatchPut %j', items);
 
     if(!(items instanceof Array)) {
       reject(new errors.ModelError('batchPut requires items to be an array'));
@@ -1302,6 +1301,7 @@ Model.batchPut = function(NewModel, items, options, next) {
 };
 
 Model.batchDelete = function(NewModel, keys, options, next) {
+  debug('BatchDel %j', keys);
 
   options = options || {};
   if(typeof options === 'function') {
@@ -1310,7 +1310,6 @@ Model.batchDelete = function(NewModel, keys, options, next) {
   }
 
   const result = new Promise((resolve, reject) => {
-    debug('BatchDel %j', keys);
 
     if(!(keys instanceof Array)) {
       reject(new errors.ModelError('batchDelete requires keys to be an array'));
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -1290,6 +1290,7 @@ Model.batchPut = async function(NewModel, items, options, next) {
   if(next){
     this.batchPut(NewModel,items,options)
       .then(result => next(undefined, result), err => next(err))
+    return
   }
 
   return new Promise((resolve,reject) => {
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -9,6 +9,7 @@ const Scan = require('./Scan');
 const errors = require('./errors');
 const reservedKeywords = require('./reserved-keywords');
 const objectPath = require('object-path');
+const {linkPromiseToCallback} = require('./helpers');
 
 //const MAX_BATCH_READ_SIZE   = 100;
 const MAX_BATCH_WRITE_SIZE  = 25;
@@ -275,22 +276,16 @@ function validKeyValue(value) {
   return value !== undefined && value !== null && value !== '';
 }
 
-Model.prototype.put = async function(options = {}, next) {
+Model.prototype.put = function(options = {}, next) {
 
   if(typeof options === 'function') {
     next = options;
     options = {};
   }
 
-  if(next){
-    this.put(options)
-      .then( result => {next(null,result)}, err => {next(err) })
-    return
-  }
-
   debug('put', this);
 
-  return new Promise((resolve, reject) => {
+  const result = new Promise((resolve, reject) => {
     let model, model$, item;
 
     function putItem() {
@@ -352,6 +347,8 @@ Model.prototype.put = async function(options = {}, next) {
       reject(err);
     }
   })
+
+  return linkPromiseToCallback(result, next);
 };
 
 Model.prototype.save = Model.prototype.put;
@@ -372,7 +369,7 @@ Model.create = function(NewModel, obj, options, next) {
   return model.save(options, next);
 };
 
-Model.get = async function(NewModel, key, options, next) {
+Model.get = function(NewModel, key, options, next) {
   debug('Get %j', key);
 
   options = options || {};
@@ -381,13 +378,7 @@ Model.get = async function(NewModel, key, options, next) {
     options = {};
   }
 
-  if(next){
-    Model.get(NewModel, key, options)
-      .then(result => {next(undefined, result)}, err => {next(err)})
-    return
-  }
-
-  return new Promise((resolve, reject) => {
+  const result = new Promise((resolve, reject) => {
     if(key === null || key === undefined) {
       reject(new errors.ModelError('Key required to get item'));
     }
@@ -467,6 +458,8 @@ Model.get = async function(NewModel, key, options, next) {
       get();
     }
   })
+
+  return linkPromiseToCallback(result,next);
 };
 
 Model.prototype.populate = function (options, resultObj, fillPath) {
@@ -550,13 +543,7 @@ Model.update = function(NewModel, key, update, options, next) {
     options = {};
   }
 
-  if(next){
-    this.update(NewModel, key, update, options)
-      .then(result => {next(undefined,result)},err => {next(err)})
-    return
-  }
-
-  return new Promise((resolve, reject) => {
+  const result = new Promise((resolve, reject) => {
     // default createRequired to false
     if (typeof options.createRequired === 'undefined') {
       options.createRequired = false;
@@ -903,9 +890,11 @@ Model.update = function(NewModel, key, update, options, next) {
       updateItem();
     }
   })
+
+  return linkPromiseToCallback(result,next);
 };
 
-Model.delete = async function (NewModel, key, options, next) {
+Model.delete = function (NewModel, key, options, next) {
 
   options = options || {};
   if(typeof options === 'function') {
@@ -913,11 +902,6 @@ Model.delete = async function (NewModel, key, options, next) {
     options = {};
   }
 
-  if(next){
-    Model.delete(NewModel, key, options)
-      .then(result => {next(undefined, result)}, err => {next(err)})
-    return
-  }
   const schema = NewModel.$__.schema;
 
   const hashKeyName = schema.hashKey.name;
@@ -932,23 +916,18 @@ Model.delete = async function (NewModel, key, options, next) {
   }
 
   const model = new NewModel(key);
-  return model.delete(options);
+  const result = model.delete(options);
+  return linkPromiseToCallback(result,next);
 };
 
-Model.prototype.delete = async function(options, next) {
+Model.prototype.delete = function(options, next) {
   options = options || {};
   if(typeof options === 'function') {
     next = options;
     options = {};
   }
 
-  if(next){
-    this.delete(options)
-      .then(result => next(undefined, result), err => next(err))
-    return
-  }
-
-  return new Promise((resolve, reject) => {
+  const result = new Promise((resolve, reject) => {
     const schema = this.$__.schema;
 
     const hashKeyName = schema.hashKey.name;
@@ -1017,6 +996,8 @@ Model.prototype.delete = async function(options, next) {
       deleteItem();
     }
   })
+
+  return linkPromiseToCallback(result,next);
 };
 
 /*
@@ -1087,7 +1068,7 @@ Model.scan = function(NewModel, filter, options, next) {
   return scan;
 };
 
-Model.batchGet = async function(NewModel, keys, options, next) {
+Model.batchGet = function(NewModel, keys, options, next) {
 
   options = options || {};
   if(typeof options === 'function') {
@@ -1095,13 +1076,7 @@ Model.batchGet = async function(NewModel, keys, options, next) {
     options = {};
   }
 
-  if(next){
-    this.batchGet(NewModel, keys, options)
-      .then(result => next(undefined, result), err => next(err))
-    return;
-  }
-
-  return new Promise((resolve, reject) => {
+  const result = new Promise((resolve, reject) => {
     debug('BatchGet %j', keys);
     if(!(keys instanceof Array)) {
       reject(new errors.ModelError('batchGet requires keys to be an array'));
@@ -1201,6 +1176,8 @@ Model.batchGet = async function(NewModel, keys, options, next) {
       batchGet();
     }
   })
+
+  return linkPromiseToCallback(result, next);
 };
 
 function toBatchChunks(modelName, list, chunkSize, requestMaker) {
@@ -1275,7 +1252,7 @@ function batchWriteItems (NewModel, batchRequests) {
   });
 }
 
-Model.batchPut = async function(NewModel, items, options, next) {
+Model.batchPut = function(NewModel, items, options, next) {
 
   options = options || {};
   if(typeof options === 'function') {
@@ -1283,13 +1260,7 @@ Model.batchPut = async function(NewModel, items, options, next) {
     options = {};
   }
 
-  if(next){
-    this.batchPut(NewModel,items,options)
-      .then(result => next(undefined, result), err => next(err))
-    return
-  }
-
-  return new Promise((resolve,reject) => {
+  const result = new Promise((resolve,reject) => {
     debug('BatchPut %j', items);
 
     if(!(items instanceof Array)) {
@@ -1325,10 +1296,12 @@ Model.batchPut = async function(NewModel, items, options, next) {
     } else {
       batchPut();
     }
-  })
+  });
+
+  return linkPromiseToCallback(result,next);
 };
 
-Model.batchDelete = async function(NewModel, keys, options, next) {
+Model.batchDelete = function(NewModel, keys, options, next) {
 
   options = options || {};
   if(typeof options === 'function') {
@@ -1336,13 +1309,7 @@ Model.batchDelete = async function(NewModel, keys, options, next) {
     options = {};
   }
 
-  if(next){
-    this.batchDelete(NewModel,keys,options)
-      .then(result => next(undefined, result), err => next(err))
-    return
-  }
-
-  return new Promise((resolve, reject) => {
+  const result = new Promise((resolve, reject) => {
     debug('BatchDel %j', keys);
 
     if(!(keys instanceof Array)) {
@@ -1382,6 +1349,8 @@ Model.batchDelete = async function(NewModel, keys, options, next) {
       batchDelete();
     }
   })
+
+  return linkPromiseToCallback(result, next)
 };
 
 module.exports = Model;
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -1,7 +1,10 @@
 'use strict';
 
+process.on('unhandledRejection', (reason, promise) => {
+  console.warn('Unhandled promise rejection:', promise, 'reason:', reason.stack || reason);
+});
+
 const util = require('util');
-const Q = require('q');
 const hooks = require('hooks');
 const Attribute = require('./Attribute');
 const Table = require('./Table');
@@ -104,7 +107,7 @@ Model.compile = function compile (name, schema, options, base) {
       return Model.get(NewModel, key, options, next);
     } catch(err) {
       sendErrorToCallback(err, options, next);
-      return Q.reject(err);
+      return Promise.reject(err);
     }
   };
 
@@ -113,7 +116,7 @@ Model.compile = function compile (name, schema, options, base) {
       return Model.populate(options, resultObj, fillPath);
     } catch(err) {
       sendErrorToCallback(err, options);
-      return Q.reject(err);
+      return Promise.reject(err);
     }
   };
 
@@ -122,7 +125,7 @@ Model.compile = function compile (name, schema, options, base) {
       return Model.update(NewModel, key, update, options, next);
     } catch(err) {
       sendErrorToCallback(err, options, next);
-      return Q.reject(err);
+      return Promise.reject(err);
     }
   };
 
@@ -131,7 +134,7 @@ Model.compile = function compile (name, schema, options, base) {
       return Model.delete(NewModel, key, options, next);
     } catch(err) {
       sendErrorToCallback(err, options, next);
-      return Q.reject(err);
+      return Promise.reject(err);
     }
   };
 
@@ -140,7 +143,7 @@ Model.compile = function compile (name, schema, options, base) {
       return Model.query(NewModel, query, options, next);
     } catch(err) {
       sendErrorToCallback(err, options, next);
-      return Q.reject(err);
+      return Promise.reject(err)
     }
   };
 
@@ -149,7 +152,7 @@ Model.compile = function compile (name, schema, options, base) {
       return Model.queryOne(NewModel, query, options, next);
     } catch(err) {
       sendErrorToCallback(err, options, next);
-      return Q.reject(err);
+      return Promise.reject(err);
     }
   };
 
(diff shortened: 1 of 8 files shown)


lib/Model.js
```diff
@@ -276,75 +279,83 @@ function validKeyValue(value) {
...
-  try {
-    options = options || {};
-    if(typeof options === 'function') {
-      next = options;
-      options = {};
-    }
-    if(options.overwrite === null || options.overwrite === undefined) {
-      options.overwrite = true;
-    }
+  if(next){
```
fishcharlie: Instead of doing this. Why not create a helper function to use throughout the codebase? For example.

```
function promisify(promise, cb) {
    promise.then(result => { cb(null, result); }, err => { next(err); });
    return promise;
}
```

Then instead of this `if` statement. We can just set the promise to a variable instead of returning it. Then do something like:

```
return promisify(promise, next);
```

That way we don't have to write the `then` logic and all of that over and over.

You might have to play with that code. Probably do need a `if` statement within that function I created to ensure that `cb` exists. Not sure if the naming works and everything like that. So might need to play with it quite a bit.

With the current implementation I think someone trying to combine a callback with a promise would fail. So something like:

```
const item = Model.get("myitem", function(err, result) {
    console.log(result);
});
```

I think in that case `item` is undefined with your changes. But I think in the current version `item` would exist.

I would almost consider that a breaking change. I have no idea how people are using Dynamoose, and although I wouldn't recommend that, I can't be positive that people aren't writing code like that.

I **think** my suggestion above would address that.

lib/Model.js
```diff
@@ -276,75 +279,83 @@ function validKeyValue(value) {
   return value !== undefined && value !== null && value !== '';
 }
 
-Model.prototype.put = function(options, next) {
-  debug('put', this);
-  const deferred = Q.defer();
-  let model, model$, item;
+Model.prototype.put = async function(options = {}, next) {
```
fishcharlie: Why do we need `async` here? And why do we need to add `= {}` after `options`?

lib/Model.js
```diff
@@ -276,75 +279,83 @@ function validKeyValue(value) {
...
+  if(next){
+    this.put(options)
+      .then( result => {next(null,result)}, err => {next(err) })
+    return
+  }
 
-    let toDynamoOptions = {
-      updateTimestamps: true
-    };
+  debug('put', this);
```
fishcharlie: I think this should be at the top of the function.

lib/Model.js
```diff
@@ -365,99 +376,101 @@ Model.create = function(NewModel, obj, options, next) {
   return model.save(options, next);
 };
 
-Model.get = function(NewModel, key, options, next) {
+Model.get = async function(NewModel, key, options, next) {
```
fishcharlie: Same thing here, why do we need the `async`?

lib/Model.js
```diff
@@ -365,99 +376,101 @@ Model.create = function(NewModel, obj, options, next) {
...
   options = options || {};
   if(typeof options === 'function') {
     next = options;
     options = {};
   }
 
-  if(key === null || key === undefined) {
-    deferred.reject(new errors.ModelError('Key required to get item'));
-    return deferred.promise.nodeify(next);
+  if(next){
```
fishcharlie: See my above comment for this.

lib/Model.js
```diff
@@ -365,99 +376,101 @@ Model.create = function(NewModel, obj, options, next) {
...
-  }
+    let hashKeyName = schema.hashKey.name;
+    if(!validKeyValue(key[hashKeyName])) {
+      let keyVal = key;
+      key = {};
+      key[hashKeyName] = keyVal;
+    }
 
+    if(schema.rangeKey && !validKeyValue(key[schema.rangeKey.name])) {
+      reject( new errors.ModelError('Range key required: ' + schema.rangeKey.name) );
```
fishcharlie:
```suggestion
      reject(new errors.ModelError('Range key required: ' + schema.rangeKey.name));
```

lib/Model.js
```diff
@@ -365,99 +376,101 @@ Model.create = function(NewModel, obj, options, next) {
...
+          debug('cannot parse data', data);
+          return reject(e);
+        }
 
-      debug('getItem parsed model', model);
-      deferred.resolve(model);
-    });
-  }
+        if (schema.expires && schema.expires.returnExpiredItems === false && model[schema.expires.attribute] < new Date()) {
+          resolve();
```
fishcharlie: It looks like this was a carbon copy of the old version. Which is great. But I wonder if this is a bug. Shouldn't this be the following?

```suggestion
          return resolve();
```

I could be totally wrong tho... what do you think?

lib/Model.js
```diff
@@ -542,363 +554,374 @@ Model.update = function(NewModel, key, update, options, next) {
...
 
-  // default createRequired to false
-  if (typeof options.createRequired === 'undefined') {
-    options.createRequired = false;
-  }
-
-  // default updateTimestamps to true
-  if (typeof options.updateTimestamps === 'undefined') {
-    options.updateTimestamps = true;
+  if(next){
```
fishcharlie: See comment above about this.

Aedalus:
I'll look into the code coverage over the next couple days as I find some time.

lib/Model.js
```diff
@@ -542,363 +554,374 @@ Model.update = function(NewModel, key, update, options, next) {
...
+    if(newModel$.options.waitForActive) {
+      newModel$.table.waitForActive().then(updateItem).catch(reject);
+    } else {
+      updateItem();
+    }
+  })
 };
 
-Model.delete = function(NewModel, key, options, next) {
+Model.delete = async function (NewModel, key, options, next) {
```
fishcharlie: Why do we need the `async` here?

lib/Model.js
```diff
@@ -542,363 +554,374 @@ Model.update = function(NewModel, key, update, options, next) {
...
+
+  options = options || {};
```
fishcharlie: Why did this section get added? At first glance it looks good, but it wasn't in the old version. Was that a bug in the old version or?

lib/Model.js
```diff
@@ -542,363 +554,374 @@ Model.update = function(NewModel, key, update, options, next) {
...
+  if(typeof options === 'function') {
+    next = options;
+    options = {};
+  }
 
+  if(next){
```
fishcharlie: See my comment above about this.

lib/Model.js
```diff
@@ -909,97 +932,95 @@ Model.delete = function(NewModel, key, options, next) {
...
   if(schema.rangeKey && !key[schema.rangeKey.name]) {
-    const deferred = Q.defer();
-    deferred.reject(new errors.ModelError('Range key required: %s', schema.hashKey.name));
-    return deferred.promise.nodeify(next);
+    return Promise.reject(new errors.ModelError('Range key required: %s', schema.hashKey.name));
   }
 
   const model = new NewModel(key);
-  return model.delete(options, next);
+  return model.delete(options);
```
fishcharlie: Do we really want to remove the `, next` here?

lib/Model.js
```diff
@@ -909,97 +932,95 @@ Model.delete = function(NewModel, key, options, next) {
...
 };
 
-Model.prototype.delete = function(options, next) {
+Model.prototype.delete = async function(options, next) {
```
fishcharlie: Why do we need the `async` here?

lib/Model.js
```diff
@@ -1070,114 +1091,120 @@ Model.scan = function(NewModel, filter, options, next) {
...
+    this.batchGet(NewModel, keys, options)
+      .then(result => next(undefined, result), err => next(err))
+    return;
   }
 
-  const batchReq = {
-    RequestItems: {}
-  };
+  return new Promise((resolve, reject) => {
+    debug('BatchGet %j', keys);
```
fishcharlie: I think this should be at the top of the function

lib/Model.js
```diff
@@ -1070,114 +1091,120 @@ Model.scan = function(NewModel, filter, options, next) {
...
-    }
-    return key;
-  });
-
-  if(schema.rangeKey && !keys.every(function (key) { return validKeyValue(key[schema.rangeKey.name]); })) {
-    deferred.reject(
-      new errors.ModelError('Range key required: ' + schema.rangeKey.name)
-    );
-    return deferred.promise.nodeify(next);
+  if(next){
```
fishcharlie: See my comment above about this

lib/Model.js
```diff
@@ -1234,119 +1261,131 @@ function batchWriteItems (NewModel, batchRequests) {
...
-  return Q.all(batchList).then(function (resultList) {
+  return Promise.all(batchList).then(function (resultList) {
     return reduceBatchResult(resultList);
   });
 }
 
-Model.batchPut = function(NewModel, items, options, next) {
-  debug('BatchPut %j', items);
-  const deferred = Q.defer();
+Model.batchPut = async function(NewModel, items, options, next) {
```
fishcharlie: Why the `async` here?

lib/Model.js
```diff
@@ -1234,119 +1261,131 @@ function batchWriteItems (NewModel, batchRequests) {
...
-  }
   options = options || {};
   if(typeof options === 'function') {
     next = options;
     options = {};
   }
 
-  const schema = NewModel.$__.schema;
-  const newModel$ = NewModel.$__;
+  if(next){
```
fishcharlie: See my comment above about this.

lib/Model.js
```diff
@@ -1234,119 +1261,131 @@ function batchWriteItems (NewModel, batchRequests) {
...
+      .then(result => next(undefined, result), err => next(err))
+    return
+  }
 
-  const toDynamoOptions = {
-    updateTimestamps: options.updateTimestamps || false,
-    updateExpires: options.updateExpires || false
-  };
+  return new Promise((resolve,reject) => {
+    debug('BatchPut %j', items);
```
fishcharlie: I think this should be at the top.

lib/Model.js
```diff
@@ -1234,119 +1261,131 @@ function batchWriteItems (NewModel, batchRequests) {
...
 
-Model.batchDelete = function(NewModel, keys, options, next) {
-  debug('BatchDel %j', keys);
-  const deferred = Q.defer();
-
-  if(!(keys instanceof Array)) {
-    deferred.reject(new errors.ModelError('batchDelete requires keys to be an array'));
-    return deferred.promise.nodeify(next);
-  }
+Model.batchDelete = async function(NewModel, keys, options, next) {
```
fishcharlie: Why the `async`?

lib/Model.js
```diff
@@ -1234,119 +1261,131 @@ function batchWriteItems (NewModel, batchRequests) {
...
   options = options || {};
   if(typeof options === 'function') {
     next = options;
     options = {};
   }
 
-  const schema = NewModel.$__.schema;
-  const newModel$ = NewModel.$__;
-  const hashKeyName = schema.hashKey.name;
+  if(next){
```
fishcharlie: See my comment above about this.

lib/Model.js
```diff
@@ -1234,119 +1261,131 @@ function batchWriteItems (NewModel, batchRequests) {
...
+    this.batchDelete(NewModel,keys,options)
+      .then(result => next(undefined, result), err => next(err))
+    return
+  }
 
-  const batchRequests = toBatchChunks(newModel$.name, keys, MAX_BATCH_WRITE_SIZE, function(key) {
-    const key_element = {};
-    key_element[hashKeyName] = schema.hashKey.toDynamo(key[hashKeyName]), undefined, key;
+  return new Promise((resolve, reject) => {
+    debug('BatchDel %j', keys);
```
fishcharlie: I think this should be at the top.

lib/Query.js
```diff
@@ -62,13 +61,17 @@ function Query (Model, query, options) {
 
 
 
-Query.prototype.exec = function (next) {
+Query.prototype.exec = async function (next) {
```
fishcharlie: Why the `async`?

lib/Query.js
```diff
@@ -62,13 +61,17 @@ function Query (Model, query, options) {
...
   debug('exec query for ', this.query);
+
+  if(next){
```
fishcharlie: See my comment above about this.

lib/Scan.js
```diff
@@ -35,13 +34,17 @@ function Scan (Model, filter, options) {
 
 
 
-Scan.prototype.exec = function (next) {
+Scan.prototype.exec = async function (next) {
```
fishcharlie: Why do we need `async` here?

lib/Scan.js
```diff
@@ -35,13 +34,17 @@ function Scan (Model, filter, options) {
...
   debug('exec scan for ', this.scan);
+
+  if(next){
```
fishcharlie: See my comment above about this.

lib/Scan.js
```diff
@@ -61,40 +64,39 @@ Scan.prototype.exec = function (next) {
...
-        if (!data) {
-          return deferred.resolve([]);
-        }
-        if (!data.Items) {
-          let counts = { count: data.Count, scannedCount: data.ScannedCount };
-          return deferred.resolve(counts);
-        }
-        return deferred.resolve(data.Items.map(function (item) {
-          let model;
+  async function scanByRawFilter() {
```
fishcharlie: Why the `async` here?

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
+  })
 };
 
 Table.prototype.init = function (next) {
-  debug('initializing table, %s, %j', this.name, this.options);
-  const deferred = Q.defer();
 
-  let table = this;
-  let localTableReq;
+  if(next){
```
fishcharlie: See my comment above about this.

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
-          return indexes.delete.reduce(function(cur, next) {
-            return cur.then(function() { return table.deleteIndex(next.IndexName); });
-          }, Q());
-        })
-        .then(function() {
-          return indexes.create.reduce(function(cur, next) {
-            return cur.then(function() { return table.createIndex(localTableReq.AttributeDefinitions, next); });
-          }, Q());
+  return new Promise((resolve, reject) => {
+    debug('initializing table, %s, %j', this.name, this.options);
```
fishcharlie: I think this should be at the top

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
 };
 
 Table.prototype.waitForActive = function (timeout, next) {
-  debug('Waiting for Active table, %s, %j', this.name, this.options);
-  const deferred = Q.defer();
 
-  if (typeof timeout === 'function') {
-    next = timeout;
-    timeout = null;
+  if(next){
```
fishcharlie: See my comment above about this.

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
 
-  let params = {
-    TableName: this.name
-  };
-  ddb.describeTimeToLive(params, function(err, ttlDescription) {
-    if (err) {
-      return deferred.reject(err);
-    }
+  return new Promise((resolve, reject) => {
+    debug('Describing ttl for table, %s', this.name);
```
fishcharlie: I think this should be at the top

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
+    waitForActive();
+  })
 };
 
 Table.prototype.describeTTL = function (next) {
-  debug('Describing ttl for table, %s', this.name);
-  const deferred = Q.defer();
 
-  let ddb = this.base.ddb();
+  if(next){
```
fishcharlie: See my comment above about this

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
+    ddb.describeTimeToLive(params, function(err, ttlDescription) {
+      if (err) {
+        return reject(err);
+      }
 
-  return deferred.promise.nodeify(next);
+      return resolve(ttlDescription);
+    });
+
+    return promise.nodeify(next);
```
fishcharlie: Where is `promise` defined? And why are we using two different methods for handling the callbacks? We do a lot of the following:

```
this.describeTTL()
       .then(result => {next(undefined,result)}, err => {next(err)})
return
```

This kinda ties into my question at the very top. And maybe if this thing exists and works, we could replace a lot of that code with this better version.

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
+  })
 };
 
 Table.prototype.updateTTL = function (next) {
-  debug('Updating ttl for table, %s', this.name);
-  const deferred = Q.defer();
 
-  let table = this;
+  if(next){
```
fishcharlie: See comment above

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
+    this.updateTTL()
+      .then(result => {next(undefined,result)}, err => {next(err)})
+    return
+  }
 
-  if(this.schema.expires && !this.base.endpointURL) {
+  return new Promise((resolve, reject) => {
+    debug('Updating ttl for table, %s', this.name);
```
fishcharlie: I think this should be at the top

lib/Table.js
```diff
@@ -75,277 +74,307 @@ const compareIndexes = function compareIndexes(local, remote) {
...
+        });
+    } else {
+      resolve();
+    }
+  })
 };
 
 Table.prototype.describe = function (next) {
+
+  if(next){
```
fishcharlie: See comment above about this

lib/Table.js
```diff
@@ -515,56 +544,71 @@ const buildTableReq = Table.prototype.buildTableReq = function buildTableReq(nam
 };
 
 Table.prototype.create = function (next) {
-  let ddb = this.base.ddb();
-  let schema = this.schema;
-  let createTableReq = this.buildTableReq(this.name, schema);
 
-  debug('ddb.createTable request:', createTableReq);
+  if(next){
```
fishcharlie: See comment above about this

lib/Table.js
```diff
@@ -515,56 +544,71 @@ const buildTableReq = Table.prototype.buildTableReq = function buildTableReq(nam
...
-  ddb.createTable(createTableReq, function (err, data) {
-    if (err) {
-      debug('error creating table', err);
-      return deferred.reject(err);
-    }
-    debug('table created', data);
-    deferred.resolve(data);
-  });
-  return deferred.promise.nodeify(next);
+    debug('ddb.createTable request:', createTableReq);
```
fishcharlie: I think this should be at the top

lib/Table.js
```diff
@@ -515,56 +544,71 @@ const buildTableReq = Table.prototype.buildTableReq = function buildTableReq(nam
...
 
 Table.prototype.delete = function (next) {
   let deleteTableReq = {
     TableName: this.name
   };
 
-  debug('ddb.deleteTable request:', deleteTableReq);
-
-  let ddb = this.base.ddb();
+  if(next){
```
fishcharlie: See comment above about this

lib/Table.js
```diff
@@ -515,56 +544,71 @@ const buildTableReq = Table.prototype.buildTableReq = function buildTableReq(nam
...
+    this.delete()
+      .then(result => {next(undefined,result)}, err => {next(err)})
+    return
+  }
 
-  const deferred = Q.defer();
+  return new Promise((resolve, reject) => {
+    debug('ddb.deleteTable request:', deleteTableReq);
```
fishcharlie: I think this should be at the top

lib/Table.js
```diff
@@ -515,56 +544,71 @@ const buildTableReq = Table.prototype.buildTableReq = function buildTableReq(nam
...
 };
 
 Table.prototype.update = function (next) {
   // let ddb = this.base.ddb();
   // ddb.updateTable();
-  const deferred = Q.defer();
-  deferred.reject(new Error('TODO'));
-  return deferred.promise.nodeify(next);
+
+  if(next){
```
fishcharlie: See comment above about this

lib/Table.js
```diff
@@ -515,56 +544,71 @@ const buildTableReq = Table.prototype.buildTableReq = function buildTableReq(nam
...
+      resolve(data);
+    });
+  })
 };
 
 Table.prototype.update = function (next) {
   // let ddb = this.base.ddb();
   // ddb.updateTable();
-  const deferred = Q.defer();
-  deferred.reject(new Error('TODO'));
```
fishcharlie: I wonder who wrote this code, and what the purpose of this is. Before this gets merged in, I should probably figure out how to look at the blame, to figure out when this was written to track it down and figure out how to implement it.

If you are able to do that quickly that'd be awesome. But I think it needs to be done before this gets merged in, or else I think it'd be a lot harder afterwards.

package.json
```diff
@@ -50,8 +50,7 @@
     "debug": "^2.6.8",
     "deep-equal": "^1.0.1",
     "hooks": "0.3.2",
-    "object-path": "^0.11.4",
-    "q": "^1.5.0"
```
fishcharlie: YAY 🎉

fishcharlie:
@Aedalus One other question. Do you think we should **add** tests for this? I mentioned that we shouldn't change the tests in this PR. But I'd totally be in for adding tests.

fishcharlie:
@Aedalus And I just saw your comment about code coverage. It's been doing that for a while now. I feel like it's some type of glitch. With all the changes you made, and no changes to the tests, losing `0.2%` coverage makes 0 sense to me. So if you want to take a look at it, feel free. But it's probably not related to this PR, and I'll merge regardless of that. If you want to look at it feel free to open a separate PR tho!

test/Model.js
```diff
@@ -1734,7 +1734,7 @@ describe('Model', function (){
             })
             .then(function (leo) {
               leo.expires = null;
-              console.log(leo);
+              // console.log(leo);
```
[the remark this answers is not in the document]
Aedalus: Yeah, np! It was a little weird to me there was just a single test that was logging. I'll just comment it back in for now.

lib/Model.js
```diff
@@ -1,7 +1,10 @@
 'use strict';
 
+process.on('unhandledRejection', (reason, promise) => {
```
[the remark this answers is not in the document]
Aedalus: Woops! Good catch, accidentally left that in after debugging.

lib/Model.js
Aedalus: So this is something I definitely struggled with, and would be up for going a different direction. 

If we return a promise and the User doesn't catch it, Node will throw an unhandled promise rejection. Q promises don't dispatch the same node event, so they just fail silently. So we can return the promise, but if they are only using a callback Node will write to std err, which hopefully would never break anything, but would at least likely raise a few eyebrows to whether something changed. I feel like failing silently definitely isn't the way to go though. If you have any ideas for how you want to handle it, let me know!

Definitely up for abstracting it with the above though. I'll try to implement the above. Unfortunately the way some functions dynamically determine if a param is a config or a callback makes it pretty difficult to do the abstraction outside the function. Would definitely make the code cleaner if support was at least dropped for passing in a cb if omitting options, but that would definitely be a breaking change.

Aedalus: Thinking on the above, is the plan for the above to continue to support callbacks, or eventually move to entirely promises? Would it make sense to start deprecating certain things like callbacks, or at least passing callbacks in place of options? Might make any eventual breaking changes to the API a lot easier, but I don't know if that's the eventual plan.

test/Model.js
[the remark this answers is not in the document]
fishcharlie: @Aedalus Totally get it. Probably a separate PR would be best tho just so they stay focused on one task.

lib/Model.js
fishcharlie: Using my first example above. Wouldn't we be catching the error there tho? Isn't the only difference the fact that we are returning the promise and abstracting it a bit more?

As of right now there is no plan to deprecate callbacks. It's something I'd be open to discussing. But as of right now there has been no discussion regarding it.

fishcharlie:
Looks like test failed due to timeout. Restarting.

fishcharlie:
@Aedalus If you could do me a favor and mark the comments I made as resolved once you finish them I'd really appreciate it. Just so we can keep track of what changes still need to be addressed before merging. I marked a few as complete based on the last commit you made.

fishcharlie:
@Aedalus Looks like this build is failing for some reason. Wanna take a look at that at some point?

fishcharlie:
@Aedalus Were you able to make any progress on this?

Aedalus:
@fishcharlie I've made some, work's just been a little crazy. Hoping to finish it up by this weekend so the PR isn't just rotting here.

lib/Model.js
Aedalus: I was trying to keep it consistent with the other function apis. If you want though, I can remove it from this change.

lib/Model.js
Aedalus: This should be abstracted into it's own file/method now. I've called it `linkPromiseToCallback`, and put it in a new `helpers.js` file. Not in love with either though, so if you want me to change the name/location let me know.

lib/Table.js
Aedalus: This should be taken out in the latest changes.

Aedalus:
@fishcharlie Newest changes are in. I tried to follow your suggestion with removing the `if(next)` statements, and abstracting it into another function. Let me know if you see anything off.

fishcharlie:
Another issue I just realized.

https://github.com/fishcharlie/dynamoose/blob/pluginsupport-async/lib/Scan.js#L207

That line. We are passing `deferredMain` in, so that will cause problems. So we will need to make sure that works fine.

fishcharlie:
@Aedalus Any progress on this? 😃

fishcharlie:

PR #464, opened 2018-11-28
Version 1.2.0

fishcharlie:
## Version 1.2.0

AWS reInvent 2018 has wrapped up, and this release includes support for all of the DynamoDB goodies announced. This version also includes beta support for plugins! Plugin support has the potential to have breaking changes in the future with no warning, so please be aware of that when using the system. Please give feedback on plugins by creating issues on the Dynamoose repository.

Please comment or [contact me](https://charlie.fish/contact) if you have any questions about this release.

### Major New Features

- **[DynamoDB Transaction Support](https://github.com/dynamoosejs/dynamoose/pull/472)** #472 
- **[DynamoDB On-Demand Support](https://github.com/dynamoosejs/dynamoose/pull/471)** #471 
- **[Plugins - BETA](https://github.com/dynamoosejs/dynamoose/pull/325)** #325 

### General

- **[ES6 Template Literals](https://github.com/dynamoosejs/dynamoose/pull/465)** #465 

### Bug Fixes

- **[Return LastKey when using RAW Scan](https://github.com/dynamoosejs/dynamoose/pull/475)** #475 
- **[Sorting index project array before comparing](https://github.com/dynamoosejs/dynamoose/pull/455)** #455

### Other

- **[Adding Node.js version 11 to Travis CI test suite](https://github.com/dynamoosejs/dynamoose/pull/468)** #468 
- **[Adding more NPM Keywords](https://github.com/dynamoosejs/dynamoose/pull/469)** #469 
- **[Updating NPM Dependencies](https://github.com/dynamoosejs/dynamoose/pull/466)** #466  #473

PR #472, opened 2018-11-29
DynamoDB Transaction Support

### LINKED ISSUE #462: DynamoDB Transaction Support
### Summary:

Today AWS announced [DynamoDB Transaction Support](https://aws.amazon.com/blogs/aws/new-amazon-dynamodb-transactions/). This ticket is for tracking adding support for that into Dynamoose.

Please feel free to comment with questions, suggestions, or comments about this functionality.

- https://aws.amazon.com/blogs/aws/new-amazon-dynamodb-transactions/
- https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html

### Type (select 1):
- [ ] Bug report
- [x] Feature suggestion
- [ ] Question
- [ ] Other suggestion

### Other:

fishcharlie:
### Summary:

This PR adds support for DynamoDB Transactions.

#### General
```
const transactionPromise = await dynamoose.transaction([
  User.transaction.update({id: "user1"}, {$ADD: {balance: -100}}),
  Charge.transaction.create({userid: "user1", product: "product1", amount: 100, status: "successful"}),
  Product.transaction.update({id: "product1"}, {$ADD: {inventory: -1}}),
  Credit.transaction.delete({id: "credit1"})
]);
```

### GitHub linked issue:
Closes #462 

### Type (select 1):
- [ ] Bug fix
- [x] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/docs/_docs/config.md
+++ b/docs/_docs/config.md
@@ -153,3 +153,24 @@ dynamoose.AWS = AWSXRay.captureAWS(require('aws-sdk'));
 ### dynamoose.setDocumentClient(documentClient)
 
 Sets the document client for DynamoDB. This can be used to integrate with Amazon Web Services DAX.
+
+### dynamoose.transaction(items[, options][, cb])
+
+Allows you to run DynamoDB transactions. Accepts an array as the `items` parameter. Every item in the `items` array must use the `Model.transaction` methods.
+
+```js
+dynamoose.transaction([
+  User.transaction.update({id: "user1"}, {$ADD: {balance: -100}}),
+  Charge.transaction.create({userid: "user1", product: "product1", amount: 100, status: "successful"}),
+  Product.transaction.update({id: "product1"}, {$ADD: {inventory: -1}}),
+  Credit.transaction.delete({id: "credit1"})
+]).then(function (result) {
+  console.log(result);
+}).catch(function (err) {
+  console.error(err);
+});
+```
+
+`options` properties:
+
+- `type` (string): The type of transaction Dynamoose will run. Can either be "get", or "write". By default if all of your items are of type "get", it will default to "get", otherwise, will default to "write".
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -206,7 +206,7 @@ Model.compile = function compile (name, schema, options, base) {
     return table.buildTableReq(table.name, table.schema);
   };
 
-  function createTransactionFunction(func, val, optionsIndex) {
+  function createTransactionFunction(func, val, optionsIndex, model) {
     return async function () {
       let args = [...arguments];
       if (typeof args[args.length - 1] === "function") {
@@ -224,7 +224,7 @@ Model.compile = function compile (name, schema, options, base) {
       }
 
       const requestObj = await func(...args);
-      return {[val]: requestObj};
+      return {[val]: requestObj, "$__": {newModel$: model}};
     };
   }
   NewModel.transaction = [
@@ -237,7 +237,7 @@ Model.compile = function compile (name, schema, options, base) {
     {method: NewModel.create, key: "Put", name: "create", optionsIndex: 1},
     {method: NewModel.update, key: "Update", name: "update", optionsIndex: 2},
   ].reduce((original, newItem) => {
-    original[newItem.name] = createTransactionFunction(newItem.method, newItem.key, newItem.optionsIndex);
+    original[newItem.name] = createTransactionFunction(newItem.method, newItem.key, newItem.optionsIndex, NewModel);
     return original;
   }, {});
 
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -207,8 +207,7 @@ Model.compile = function compile (name, schema, options, base) {
   };
 
   function createTransactionFunction(func, val, optionsIndex, model) {
-    return async function () {
-      let args = [...arguments];
+    return async function (...args) {
       if (typeof args[args.length - 1] === "function") {
         console.warning("Dynamoose Warning: Passing callback function into transaction method not allowed. Removing callback function from list of arguments.");
         // Callback function passed in which is not allowed, removing that from arguments list
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -998,7 +998,9 @@ Model.prototype.delete = function(options, next) {
     });
   }
 
-  if(model$.options.waitForActive) {
+  if (options.returnRequest) {
+    deferred.resolve(getDelete);
+  } else if (model$.options.waitForActive) {
     model$.table.waitForActive().then(deleteItem).catch(deferred.reject);
   } else {
     deleteItem();
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -210,6 +210,7 @@ Model.compile = function compile (name, schema, options, base) {
     return async function () {
       let args = [...arguments];
       if (typeof args[args.length - 1] === "function") {
+        console.warning("Dynamoose Warning: Passing callback function into transaction method not allowed. Removing callback function from list of arguments.");
         // Callback function passed in which is not allowed, removing that from arguments list
         args.pop();
       }
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -224,6 +224,11 @@ Model.compile = function compile (name, schema, options, base) {
       }
 
       const requestObj = await func(...args);
+
+      if (val === "Update") {
+        delete requestObj.ReturnValues;
+      }
+
       return {[val]: requestObj, "$__": {newModel$: model}};
     };
   }
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -454,7 +454,7 @@ Model.get = function(NewModel, key, options, next) {
 
   if (options.returnRequest) {
 	  deferred.resolve(getReq);
-  } else if(newModel$.options.waitForActive) {
+  } else if (newModel$.options.waitForActive) {
     newModel$.table.waitForActive().then(get).catch(deferred.reject);
   } else {
     get();
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -336,7 +336,9 @@ Model.prototype.put = function(options, next) {
     model = this;
     model$ = this.$__;
 
-    if(model$.options.waitForActive) {
+    if (options.returnRequest) {
+      deferred.resolve(item);
+    } else if (model$.options.waitForActive) {
       model$.table.waitForActive().then(putItem).catch(deferred.reject);
     } else {
       putItem();
@@ -890,7 +892,9 @@ Model.update = function(NewModel, key, update, options, next) {
     });
   }
 
-  if(newModel$.options.waitForActive) {
+  if (options.returnRequest) {
+    deferred.resolve(updateReq);
+  } else if (newModel$.options.waitForActive) {
     newModel$.table.waitForActive().then(updateItem).catch(deferred.reject);
   } else {
     updateItem();
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -206,6 +206,40 @@ Model.compile = function compile (name, schema, options, base) {
     return table.buildTableReq(table.name, table.schema);
   };
 
+  function createTransactionFunction(func, val, optionsIndex) {
+    return async function () {
+      let args = [...arguments];
+      if (typeof args[args.length - 1] === "function") {
+        // Callback function passed in which is not allowed, removing that from arguments list
+        args.pop();
+      }
+
+      // Adding returnRequest to options
+      if (args.length >= (optionsIndex + 1)) {
+        args[optionsIndex] = {...args[optionsIndex], returnRequest: true};
+      } else if (args.length < optionsIndex) {
+        args[optionsIndex] = {returnRequest: true};
+      } else {
+        args.push({returnRequest: true});
+      }
+
+      const requestObj = await func(...args);
+      return {[val]: requestObj};
+    };
+  }
+  NewModel.transaction = [
+    // method - the method to be called when the user calls this method
+    // key - the key to be attached to the returned request object to pass into the DynamoDB transaciton request
+    // name - the name to attach to the resulting object for NewModel.transaction
+    // optionsIndex - the zero-based index for where the options parameter should be in the method arguments list
+    {method: NewModel.get, key: "Get", name: "get", optionsIndex: 1},
+    {method: NewModel.delete, key: "Delete", name: "delete", optionsIndex: 1},
+    {method: NewModel.create, key: "Put", name: "create", optionsIndex: 1},
+    {method: NewModel.update, key: "Update", name: "update", optionsIndex: 2},
+  ].reduce((original, newItem) => {
+    original[newItem.name] = createTransactionFunction(newItem.method, newItem.key, newItem.optionsIndex);
+    return original;
+  }, {});
 
   // apply methods and statics
   applyMethods(NewModel, schema);
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -452,7 +452,9 @@ Model.get = function(NewModel, key, options, next) {
     });
   }
 
-  if(newModel$.options.waitForActive) {
+  if (options.returnRequest) {
+	  deferred.resolve(getReq);
+  } else if(newModel$.options.waitForActive) {
     newModel$.table.waitForActive().then(get).catch(deferred.reject);
   } else {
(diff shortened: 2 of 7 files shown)


fishcharlie:
I still need to write tests for this PR. Not quite sure how I'm going to do that yet since DynamoDB Local hasn't added support for transactions. Might have to do some type of AWS mock or something like that. Still undecided about that.

Also need to run some manual tests on this as well to ensure it's working correctly.

fishcharlie:
Hi @jutaz! Thanks for offering to help on this! I could for sure use some help. Currently we have 1 test failing due to the `UnknownOperationException: An unknown operation was requested.`. If you want to take a look at it and see what you can do just branch off of my branch `fishcharlie:ddbtransactionsupport`, then you can submit a PR to merge your work into my branch or something like that.

Overall, a code review would be helpful as well. Just having a second pair of eyes to look over would be great.

In terms of patching the AWS client itself. We have a method `setDDB` that we could take advantage of to inject a custom DynamoDB instance. Currently we don't do that for any tests yet tho, so that is why I'm a bit hesitant about it.

I also tweeted about the fact that transactions aren't support on DynamoDB Local yet. Looks like someone from AWS replied. So hopefully that put it on their radar that it's a priority.

https://twitter.com/char_fish/status/1068347718928523267

fishcharlie:
@jutaz A code review, and finishing the tests are the two things we need for this to be done. Once it gets done, I'm planning on getting a release out to NPM shortly after that is done (#464). I'd **LOVE** to get that done by tomorrow or this weekend. But it all depends on when those things get done.

fishcharlie:
Oh sorry. One more thing. We need to add support for `ConditionCheck` (https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB.html#transactWriteItems-property).

I know this is a huge feature for DynamoDB and one that has a lot of excitement around it. So I do want to get it out ASAP. So I MIGHT consider moving that to a future release if this other stuff drags on for too long.

At some point I'd also love to add support for `.save` on a model instance and having transaction support for that. But that is for sure a future release.

jutaz:
I'll take a look at the failing test, and if I'll be able to fix it - will submit a PR to your branch.

> Overall, a code review would be helpful as well. Just having a second pair of eyes to look over would be great.

Yep, will do that as well 👍 


I think that it might be worth just releasing the initial transactions support, even without `ConditionCheck`, even if it is a really good addition. That can be added, if I understand correctly, without any breaking changes anyway, so can be a follow up. And if the AWS local instance supports transactions by default - that will make it so much easier to test.

docs/_docs/config.md
```diff
@@ -153,3 +153,24 @@ dynamoose.AWS = AWSXRay.captureAWS(require('aws-sdk'));
 ### dynamoose.setDocumentClient(documentClient)
 
 Sets the document client for DynamoDB. This can be used to integrate with Amazon Web Services DAX.
+
+### dynamoose.transaction(items[, options][, cb])
+
+Allows you to run DynamoDB transactions. Accepts an array as the `items` parameter. Every item in the `items` array must use the `Model.transaction` methods.
```
jutaz: Might be worth outlining that a regular, non-dynamoose query can also be passed into this method, as it simply accepts the request to send to Dynamo.

lib/Model.js
```diff
@@ -206,6 +206,45 @@ Model.compile = function compile (name, schema, options, base) {
     return table.buildTableReq(table.name, table.schema);
   };
 
+  function createTransactionFunction(func, val, optionsIndex, model) {
+    return async function () {
+      let args = [...arguments];
+      if (typeof args[args.length - 1] === "function") {
+        // Callback function passed in which is not allowed, removing that from arguments list
```
jutaz: Would probably be a good idea to warn about that - might lead to unexpected results, if callback is being passed in, and never called.

lib/Model.js
```diff
@@ -206,6 +206,45 @@ Model.compile = function compile (name, schema, options, base) {
     return table.buildTableReq(table.name, table.schema);
   };
 
+  function createTransactionFunction(func, val, optionsIndex, model) {
+    return async function () {
+      let args = [...arguments];
```
jutaz: Can't we just do:
```js
return async function (...args) {
  // ...
```

I see we use spread in other places too, and even have object spread, so I'd suspect this is also available to us.

lib/index.js
```diff
@@ -130,4 +133,87 @@ Dynamoose.prototype.revertDDB = function () {
...
+  debugTransaction('Run Transaction');
+  const deferred = Q.defer();
+
+  options = options || {};
+  if(typeof options === 'function') {
+    next = options;
+    options = {};
+  }
+
+  if(items === null || items === undefined || items.length === 0) {
```
jutaz: I think we can just do `!Array.isArray(items) || items.length === 0`. This will also cover cases where something that's truthy is passed in. I know this is a bit nitpick-y, but will future-proof.

lib/index.js
```diff
@@ -130,4 +133,87 @@ Dynamoose.prototype.revertDDB = function () {
...
+
+  items = await Promise.all(items);
+  let transactionReq = {
+    TransactItems: items.map(item => {
+      const returnItem = {...item};
+      delete returnItem.$__;
+      return returnItem;
+    })
+  };
+  let transactionMethodName = items.map(obj => Object.keys(obj)[0]).every(key => key === "Get") ? "transactGetItems" : "transactWriteItems";
```
jutaz: We should probably throw an error if the types differ between the requests - IIRC Dynamo will not support mixed read/write requests in the same transaction initially, thus if mixed `Get` and `Write` queries are passed in, it will break.

What about something like this:
```js
const operations = new Set(...items.map(obj => Object.keys(obj)[0]));

if (operations.length !== 1) {
  // Ensure same type of operations are provided.
  deferred.reject(new errors.TransactionError('Please use same query types - either "read" or "write" in the same transaction.'));
}

// Operations only have a single item at this point.
let transactionMethodName = operations.has('Get') ? "transactGetItems" : "transactWriteItems"
```

lib/index.js
```diff
@@ -130,4 +133,87 @@ Dynamoose.prototype.revertDDB = function () {
...
+      deferred.reject(new errors.TransactionError('Invalid type option, please pass in "get" or "write"'));
+      return deferred.promise.nodeify(next);
+    }
+  } else {
+    debugTransaction("Using predetermined transaction method");
+  }
+  debugTransaction(`Using transaction method: ${transactionMethodName}`);
+
+  const transact = () => {
+    console.log(JSON.stringify(transactionReq));
```
jutaz: lets not forget to 🔥 this in final implementation.

lib/index.js
jutaz: Also - we should probably only do this computation _if_ `options.type` is not specified. i.e:
```js
let transactionMethodName;

if (options.type) {
  debugTransaction("Using custom transaction method");
  // ...
} else {
  debugTransaction("Using predetermined transaction method");
  transactionMethodName = // ...
}
```

lib/index.js
```diff
@@ -130,4 +133,87 @@ Dynamoose.prototype.revertDDB = function () {
...
+          if (item[prop] instanceof DynamoDBSet) {
+            item[prop] = item[prop].values;
+          }
+        });
+
+        model = new TheModel(item);
+        model.$__.isNew = false;
+        debugTransaction(`${transactionMethodName} parsed model`, model);
+        return model;
+      }).filter(item => !(schema.expires && schema.expires.returnExpiredItems === false && item[schema.expires.attribute] && item[schema.expires.attribute] < new Date())));
```
jutaz: This is minor, but for _consistency_ i'd instantiate the `new Date()` once, and then filter based on it here. This will mean that any sort of nanosecond/millisecond roll-overs will have no weird effect, and returned records will just have a single date reference.

fishcharlie:
@jutaz Awesome! Will take a look soon and get back to you.

docs/_docs/config.md
fishcharlie: @jutaz Sadly this is not actually the case. The custom `Model.transaction` methods attach the model itself to the object being passed in. This is so that `dynamoose.transaction` can correctly parse responses from DynamoDB. I THINK Dynamoose stores all the models somewhere tho. So maybe at some point we could change it to rely on that system, instead of needing to pass the model in. If we change that up at some point, then yes, I agree we should add that to the documentation.

lib/Model.js
fishcharlie: @jutaz Think we should do this as a `console.warning` or `debug`?

docs/_docs/config.md
jutaz: IIRC there's a place somewhere in Dynamoose in which we can get the Model reference by a table name, and that might be the key to make this work

lib/Model.js
fishcharlie: @jutaz Yes. Great idea! Will make that change soon.

lib/index.js
fishcharlie: @jutaz Great idea! Will do that soon.

lib/Model.js
jutaz: I'd do `console.warning` most of people (myself included) don't run local applications with debug turned on, unless there's a specific issue I'm debugging

lib/index.js
fishcharlie: @jutaz There are a lot of places I feel like throughout the code where we just rely on DynamoDB throwing the error. If DynamoDB ever decides to do something crazy and allow mixed queries, then no changes would be required to support that.

lib/index.js
fishcharlie: @jutaz Of course. Will do that. Thanks for the reminder.

lib/index.js
fishcharlie: @jutaz This code is written throughout the entire codebase. So I think a separate PR would need to address this. Not quite sure I understand what you are talking about tho. If you want to create a PR for that to change it throughout the codebase that'd be awesome! Or create an issue for this with a bit more detail on the proposed new solution.

lib/index.js
fishcharlie: @jutaz I don't think so actually. We want the default to be `items.map(obj => Object.keys(obj)[0]).every(key => key === "Get") ? "transactGetItems" : "transactWriteItems"` I believe. Then if `options.type` is specified, we override that. Otherwise we could end up where `transactionMethodName` is null or undefined.

fishcharlie:
@jutaz Just reviewed that PR. What else needs to be done do you think to get this finalized?

jutaz:
> What else needs to be done do you think to get this finalized?

I wasn't able to fully get to the read transaction - it seems to be failing a test that I added in 
https://github.com/fishcharlie/dynamoose/pull/2 due to missing `schema` variable, which I'm not 100% sure on where to access it

jutaz:
> @jutaz Just submitted a PR to your fork that should fix that. Once that gets merged in I will merge fishcharlie#2 in.

Merged that in 👍 

> Did you get around to doing any manual tests?

Did not get around to that just _yet_, but will do in the next hour.

> And after that gets merged in? Any thoughts on next steps? I think we are getting close to merging.

So far it seems to be in a pretty good shape!

fishcharlie:
@jutaz Awesome! So aiming to get this out tonight at some point. Will need to wrap up a few other things for the release (#464). But should hopefully get all that done tonight.

Keep me posted on the status of manual tests.

jutaz:
Alrighty, getting on to test this on real stuff. Hold tight on results.

jutaz:
One thing that I think we should improve upon:
screen shot 2018-11-30 at 9 31 00 pm

Currently `Get` transactions will return the model nested as `Item`, which probably isn't what we want - shouldn't we return just a model for each item?

<details>

<summary>Code to Reproduce</summary>

```js
const dynamoose = require('./');

dynamoose.AWS.config.update({
  region: 'us-east-1'
});


var dogSchema = new dynamoose.Schema({
  ownerId: {
    type: Number,
    validate: function(v) { return v > 0; },
  },
  id: {
    type: String,
    hashKey: true,
    index: true
  },
  race: {
     type: String,
     enum: ['Golden retriever', 'Beagle']
  },
  breed: {
    type: String,
    trim: true,
    required: true
  },
  color: {
    lowercase: true,
    type: [String],
    default: ['Brown']
  },
  age: Number
});

const DogModel = dynamoose.model('Justas-transactions-test', dogSchema);

dynamoose.transaction([
  DogModel.transaction.create({id: 10000, ownerId: 1, breed: 'whatevs'}),
  DogModel.transaction.create({id: 1, race: "Beagle", ownerId: 1, breed: 'whatevs'})
])
// Already exists
.catch((err) => {
  console.error(err);
  return null;
})
.then((data) => {
  console.log('create', data);
  return dynamoose.transaction([
    DogModel.transaction.get({id: '10000'}),
    DogModel.transaction.get({id: '1'})
  ]);
}).then((data) => {
  console.log(data);
  console.log(data[0].race);
}).catch(console.error);
```

</details>

fishcharlie:
@jutaz Also, just reminding about https://github.com/fishcharlie/dynamoose/pull/2/files#r237941170

jutaz:
> Any chance you could write a test for this that fails under the current situation? Then either you or I can fix it

Yep, pushing changes in a moment to my branch.

lib/Model.js
fishcharlie: Just addressed this in lasted commit.

lib/Model.js
fishcharlie: Just addressed this in latest commit

lib/index.js
fishcharlie: Just addressed this in latest commit.

lib/index.js
fishcharlie: Complete in latest commit

docs/_docs/config.md
jutaz: Nope, I think this is something that can be handled as a follow-up. Issue would be nice for this, I'd say 👍

lib/index.js
fishcharlie: @jutaz Just to make sure I'm clear. The major benefit to changing this is performance, correct?

Did address in latest commit.

jutaz: > Just to make sure I'm clear. The major benefit to changing this is performance, correct?

Performance, and also - this makes reading the code much easier: you have 2 logical branches that the code can possibly go, thus making it clearer to understand, based on external factors (in this case `options.type`'s value) which branch is executing.

docs/_docs/config.md
fishcharlie: See #476

lib/index.js
fishcharlie: See #477

lib/index.js
fishcharlie: @jutaz Got it!

fishcharlie:
@jutaz I think this is good to go. Wanna give it one last review before I merge it?

fishcharlie:
Add ConditionCheck Support to Dynamoose Transactions: #478

fishcharlie:
@jutaz **THANK YOU** SOOO much for your help on this!!! I can't express how appreciative I am of your help in regards to this. ❤️ the open source community.

fishcharlie:
Created issue #479 to track refactoring and removing code after DynamoDB Local adds Transaction support.

jutaz:
@fishcharlie Thanks to you too for maintaining this & opening this PR so quickly after the feature was announced! Looking forward to taking a bit deeper dive into codebase to try to simplify few things too. Will express my ideas via issues 😄

fishcharlie:
@jutaz Of course! I'm **SUPER** excited for this feature. I remember when it was announced I was super excited about it. Then they announced the on-demand, pay per request, the next day, and I still can't decide which I'm more excited for haha.

For sure submit ideas/problems/feedback via issues once you get a chance to dive in deeper!!

Will merge in soon, and get version 1.2.0 out within the next 6 hours or so (hopefully).

fishcharlie:
We should probably add a few more tests to increase the code coverage a bit more. I think a few more simple tests could help cover more cases.

https://coveralls.io/builds/20392553/source?filename=lib%2Findex.js#L211
https://coveralls.io/builds/20392553/source?filename=lib%2FModel.js#L264

PR #499, opened 2018-12-06
Throwing error when setting required attribute to empty string

### LINKED ISSUE #155: Attributes are deleted when empty string passed as value even when required: true
If I set an attribute in a schema to be required, set it to '' in the model, and then call .save(), the save fails due to the field being required (this makes sense to me).

If I pass that key with a value of '' to .update() it will success and delete the attribute from the document. The logic is that if the value is null, undefined, or '' then a delete operation is added to the update expression. I think that this update should be rejected since the field now no longer exists despite being required in the schema.

I've tried to work around this with a validation function that checks for empty string, however it seems that the validation doesn't run on empty strings (presumably because they will be deleted).

fishcharlie:
### Summary:

This PR does not allow deleting an attribute that is required in the schema when preforming an update action.

### GitHub linked issue:
Closes #155 

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [ ] No
- [x] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -747,6 +747,10 @@ Model.update = function(NewModel, key, update, options, next) {
 
     this.addRemove = function(name, item) {
       if(schema.hashKey.name !== name && (schema.rangeKey || {}).name !== name) {
+        let deleteAttr = schema.attributes[name];
+        if (deleteAttr.required) {
+          throw new errors.ValidationError('Required value missing: ' + this.name);
+        }
         this.REMOVE[name] = item;
       }
     };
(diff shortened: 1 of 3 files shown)


fishcharlie:
I'm not sure if this should be considered a breaking change or not. Users SHOULD not be trying to update a required attribute to something like an empty string. But we are relying on users code being correct. There is a chance this won't break anyone's code. But I think there is probably a good chance it would.

@brandongoode thoughts and review?

Issue #516, opened 2018-12-26
Sets not working


dobrynin:
### Summary:
I cannot get DynamoDB String Sets to work. I have not tried Number Sets but I assume they are also affected.

#### Schema
{
    mySet: [String],
}

I do not have a great understanding of the whole codebase, but I managed to solve my issue by removing `util.isArray(typeVal)` from the conditional at the beginning of this code block in `Attribute.js`. I'm not sure if this creates a separate issue somewhere else.

```
if (util.isArray(typeVal) || typeVal === 'list'){
    type = 'List';
  } else if ( (util.isArray(typeVal) && typeVal.length === 1) || typeof typeVal === 'function') {
    this.isSet = util.isArray(typeVal);
    let regexFuncName = /^Function ([^(]+)\(/i;
    let found = typeVal.toString().match(regexFuncName);
    type = found[1];
    if (type === 'Object') {
      type = 'Map';
    }
  } else if (typeof typeVal === 'object' || typeVal === 'map'){
    type = 'Map';
  } else {
    type = typeof typeVal;
  }
```

### Current output and behavior:

{
    mySet: [],
}

### Expected output and behavior:

{
    mySet: ['Elements', 'in', 'set'],
}

### Environment:

Operating System: Arch Linux
Operating System Version: 4.19.2
Node.js version (`node -v`): 11.2.0
NPM version: (`npm -v`): 6.4.1
Dynamoose version: 1.3.0
Dynamoose Plugins: N/A

-

### Dynamoose Plugins:

- [ ] Yes, I believe that one or multiple 3rd party Dynamoose plugins are effecting this issue
- [ ] No, I believe this issue is independent of any 3rd party Dynamoose plugins I'm using
- [ ] Unknown, I'm unsure if Dynamoose plugins are effecting this issue
- [x ] I am not using any Dynamoose plugins

### Type (select 1):
- [x] Bug report
- [ ] Feature suggestion
- [x] Question
- [ ] Other suggestion
- [ ] Something not listed here

### Other:
- [x] I have read through the Dynamoose documentation before posting this issue
- [ x I have searched through the GitHub issues (including closed issues) and pull requests to ensure this issue has not already been raised before
- [x] I have searched the internet and Stack Overflow to ensure this issue hasn't been raised or answered before
- [x] I have tested the code provided and am confident it doesn't work as intended
- [N/A] I have ensured that all of my plugins that I'm using with Dynamoose are listed above
- [x] I have filled out all fields above
[...]

PR #521, opened 2018-12-31
Always use Array.isArray to check if array

dobrynin:
### Summary:
Change usage of `array instanceOf Array` and `util.isArray(array)` to `Array.isArray(array)`.
My team found this to be an issue when trying to use the `Model.batchPut`. 

### Other information:

- `array instanceof Array` breaks when multiple `Array` globals exist: http://web.mit.edu/jwalden/www/isArray.html

- `util.isArray` is deprecated: https://nodejs.org/api/util.html#util_util_isarray_object

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Documentation improvement
- [ ] Testing improvement

- [ ] Test added to report bug (GitHub issue #--- @---)
- [ ] Something not listed here

### Is this a breaking change? (select 1):
- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

### Is this ready to be merged into Dynamoose? (select 1):
- [x] Yes
- [ ] No

### Are all the tests currently passing on this PR? (select 1):
- [x] Yes
- [ ] No

### Other:
- [x] I have searched through the GitHub pull requests to ensure this PR has not already been submitted
- [x] I have updated the Dynamoose documentation (if required) given the changes I made
- [x] I have added/updated the Dynamoose test cases (if required) given the changes I made
- [x] I have run `npm test` from the root of the project directory to ensure all tests continue to pass
- [x] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoosejs/dynamoose/blob/master/LICENSE.txt)
- [x] All of my commits and commit messages are detailed, explain what changes were made, and are easy to follow and understand
- [x] I have confirmed that all my code changes are indented properly using 2 spaces
- [x] I have filled out all fields above

PR #525, opened 2019-01-01
Fix parsing of sets when saveUnknown=true

dobrynin:
### Summary:
When `saveUnknown === true`, Dynamoose incorrectly parses sets as defined in a `Schema`. This PR resolves this issue.

### GitHub linked issue:
#516 

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -785,5 +785,8 @@ module.exports.lookupType = function (dynamoObj) {
   if(dynamoObj.NS !== null && dynamoObj.NS !== undefined) {
     return [Number];
   }
+  if(dynamoObj.SS !== null && dynamoObj.SS !== undefined) {
+    return [String];
+  }
 
 };
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -205,16 +205,16 @@ Attribute.prototype.setTypeFromRawValue = function(value) {
     typeVal = value.type;
   }
 
-  if ( (util.isArray(typeVal) && typeVal.length === 1) || typeof typeVal === 'function') {
+  if ((util.isArray(typeVal) && typeof typeVal[0] !== 'function') || typeVal === 'list'){
+    type = 'List';
+  } else if ( (util.isArray(typeVal) && typeVal.length === 1) || typeof typeVal === 'function') {
     this.isSet = util.isArray(typeVal);
-    const regexFuncName = /^Function ([^(]+)\(/i;
-    const found = typeVal.toString().match(regexFuncName);
+    let regexFuncName = /^Function ([^(]+)\(/i;
+    let found = typeVal.toString().match(regexFuncName);
     type = found[1];
     if (type === 'Object') {
       type = 'Map';
     }
-  } else if (util.isArray(typeVal) || typeVal === 'list') {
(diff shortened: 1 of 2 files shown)


lib/Attribute.js
```diff
@@ -205,16 +205,16 @@ Attribute.prototype.setTypeFromRawValue = function(value) {
     typeVal = value.type;
   }
 
-  if (util.isArray(typeVal) || typeVal === 'list'){
-    type = 'List';
-  } else if ( (util.isArray(typeVal) && typeVal.length === 1) || typeof typeVal === 'function') {
+  if ( (util.isArray(typeVal) && typeVal.length === 1) || typeof typeVal === 'function') {
```
fishcharlie: Will this cause merge conflicts with #521? Maybe we should ensure that we merge that PR in first, or?

fishcharlie:
@dobrynin Are we also sure this is not a breaking change? Seems very likely this could break people's code.

lib/Attribute.js
dobrynin: Yes it probably will cause a conflict. I think when you merge this into master you should be able to reconcile the differences?

dobrynin:
@fishcharlie, I've simplified/improved the code. I think the logic I used to have could have false positives. I suppose this a breaking change because it will change the behavior of the package, although I believe it will bring it in line with what the documentation says.

I've also added tests.

fishcharlie:
@dobrynin 

> I suppose this a breaking change because it will change the behavior of the package, although I believe it will bring it in line with what the documentation says.

How will people need to change their code in order to fix any breaking changes this PR might cause?

fishcharlie:
@dobrynin Looks like we have a merge conflict

dobrynin:
Unfortunately I think this will be a breaking a change. Consider this example:

You have data in a DynamoDB table with a nested list field: `myList`.
You have `saveUnknown` activated.
You use this syntax for your schema:

```javascript
myMap: {
  type: 'map',
  map: {
    myList: [String]
  }
}
```

Previously this would have been handled fine. But with this change, an error will be thrown because Dynamoose will interpret `myList` as a String Set but will encounter a list when traversing the table.

fishcharlie:
@dobrynin Ok. Might be a while before this gets released then. Is this something you need for your own projects or something like that? I really don't want to negatively effect your software by it taking a while for this to get released.

One thing that could REALLY help get this released quicker would be writing up a detailed summary that we can include in the release notes about how people can update their code to handle this change. So like migration changes required to ensure compatibility with this change.

Also tagging @brandongoode to see if he has any input in this.

dobrynin:
@fishcharlie It turns out that this change actually isn't crucial for my team any longer. We've recently invested some time into updating our schemas and data (as well as Dynamoose itself) and have now been able to turn off `saveUnknown`. This PR still corrects a bug, but there is no pressure from our end to get it merged in.

As for writing up a migration document, I'd be happy to do that when I find the time.

PR #554, opened 2019-01-31
Parse raw filter queries

manuelxaguilar:
### Summary:

When using a raw AWS FilterExpression on a scan, the results that are saved on the database as strings but that are declared on the schema as a different type (such as a prop of type Array with a default value of [] being stored as '[]') are not getting parsed before returning the response (as opposed to when you do a simple scan).
Added the already integrated parsing mechanism as a step for these cases.

### GitHub linked issue:

https://github.com/dynamoosejs/dynamoose/issues/550

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [ ] No
- [x] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -556,7 +556,7 @@ Attribute.prototype.parseDynamo = async function(json) {
       return;
     }
     if(isSet) {
-      const set = json[`${type  }S`] || json;
+      const set = json[type + 'S'] || set;
       return (await Promise.all(set.map(async function (v) {
         if(errorHandlingTransform) {
           return await errorHandlingTransform(v);
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -556,7 +556,7 @@ Attribute.prototype.parseDynamo = async function(json) {
       return;
     }
     if(isSet) {
-      const set = json[type + 'S'];
+      const set = json[`${type  }S`] || json;
       return (await Promise.all(set.map(async function (v) {
         if(errorHandlingTransform) {
           return await errorHandlingTransform(v);
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -556,7 +556,7 @@ Attribute.prototype.parseDynamo = async function(json) {
       return;
     }
     if(isSet) {
-      const set = json[type + 'S'] || set;
+      const set = json[type + 'S'] || json;
       return (await Promise.all(set.map(async function (v) {
         if(errorHandlingTransform) {
           return await errorHandlingTransform(v);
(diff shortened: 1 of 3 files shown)


lib/Attribute.js
```diff
@@ -556,7 +556,7 @@ Attribute.prototype.parseDynamo = async function(json) {
       return;
     }
     if(isSet) {
-      const set = json[type + 'S'];
+      const set = json[type + 'S'] || json;
```
fishcharlie: I'm concerned about this change. Isn't the Dynamoose JSON stored as something like the following?

```json
{
    "property": {
        "S": "Hello World"
    }
}
```

This change looks like it's assuming like it's the following.

```json
{
    "property": "Hello World"
}
```

lib/Scan.js
```diff
@@ -82,20 +82,18 @@ Scan.prototype.exec = async function (next) {
...
 
           Object.keys(item).forEach(function (prop) {
             if (item[prop] instanceof DynamoDBSet) {
               item[prop] = item[prop].values;
             }
           });
 
-          model = new Model(item);
-          model.$__.isNew = false;
+          const model = await toModel(item)
```
fishcharlie:
```suggestion
          const model = await toModel(item);
```

lib/Scan.js
```diff
@@ -82,20 +82,18 @@ Scan.prototype.exec = async function (next) {
...
+        let returnItems = await Promise.all(data.Items.map(async function (item) {
 
           Object.keys(item).forEach(function (prop) {
             if (item[prop] instanceof DynamoDBSet) {
               item[prop] = item[prop].values;
             }
           });
 
-          model = new Model(item);
-          model.$__.isNew = false;
```
fishcharlie: Why did this line get removed?

lib/Scan.js
fishcharlie: Mind also describing why this change was made? Or the reasoning behind it?

comsky:
Hi there!
Is it still on progress? I'm stuck at the same issue.

My GraphQL query return type includes JSON format.
When I use scan with filter expression, it doesn't parse the result.

**1. scan.({Id: {eq: '12345'}}).exec()**
=> parsed result

**2. scan.({
FilterExpression: '#id = :val0',
ExpressionAttributeNames: { '#id': 'Id' }
ExpressionAttributeValues: { ':val0': '12345' }
}).exec()**
=> same result, but raw string format

PR #564, opened 2019-02-06
fix(plugin): fix resolution shouldContinue for a plugin

hweeks:
### Summary:

This is an update to the tests to catch error states, fixes the
wording in the testing, and covers the update to the NewModel
emit function. This is a blocking issue for me in use, and allows
some features to be implemented to align more fully with 
current mongoose. 

> there are a few other features I would like to add, if this is good work I can discuss further 

### Code sample:
```js
const VerifyUniquePlugin = (plugin, declaredSchema) => {
  plugin.setName('VerifyUniquePlugin')
  plugin.setDescription('A simple plugin to check if a key is supposed to be unique and enforce it.')
  plugin.on('model:put', 'request:pre', async (request) => {
    const {event, model} = request
    const putQuery = event.item.Item
    let foundValues
    const uniqueItems = Object.keys(putQuery).filter(key => declaredSchema.attributes[key].options.unique)
    if (uniqueItems.length > 0) {
      foundValues = await Promise.all(uniqueItems.map(async itemName => {
        const findQuery = {[itemName]: putQuery[itemName].S}
        const results = await model.scan(findQuery).exec()
        return results
      }))
    }    
    return new Promise((resolve)=> {
      const response = {}
      if (!lodash.isEmpty(lodash.flatten(foundValues))) {
        response['reject'] = 'This is a duplicate, can not create another.'
      }
      response['resolve'] = 'Value is unique.'      
      resolve(response)
    })
    
  })
  return plugin
}
```

#### Schema
```js
const schema = {
  name: {
    type: String,
    unique: true
  }
}
```

#### Model
```js
const builtModel = dynamoose.model('NoDupeModel', schema)
builtModel.plugin(VerifyUniquePlugin, schema)
```

#### General
```js
builtModel.create({name: 'John'})
// this one should not create another entry
builtModel.create({name: 'John'})
```

it also changes how this works

```js
builtModel.batchPut([{name: 'John'},{name: 'John'}])
```

this case should only create a single model.

### Other information:

I understand the Plugins are in very early beta, but I'd very much like to use them. This work increases coverage and fixes what I believe is a miss. If I've made a bad assumption in my plugin, please let me know. As another side note, the plugin docs have the wrong event names included. Should I fix that here, or as a separate docs only commit?

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] Test added to report bug 

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:
  - [x] `npm test`
  - [x] `eslint .`

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Dynamoose has a few helpers for local development. These are included in the scr
 
 ## npm test
 
-This will run all of our tests via grunt. This spins up a local Dynamo instance, runs mocha, and closes the Dynamo instance at completion.
+This will run all of our tests via grunt. This spins up a local DynamoDB Local instance, runs Mocha, and closes the DynamoDB Local instance at completion.
 
 ## npm run lint
 
--- a/README.md
+++ b/README.md
@@ -58,3 +58,15 @@ The Dynamoose Changelog can be found in the [CHANGELOG.md](//github.com/dynamoos
 ## Roadmap
 
 The Dynamoose Roadmap can be found in the [ROADMAP.md](//github.com/dynamoosejs/dynamoose/blob/master/ROADMAP.md) file. Help is always appreciated on these items. If you are able to help submit a PR so we can review and improve Dynamoose!
+
+## Development
+
+Dynamoose has a few helpers for local development. These are included in the scripts section of our `package.json`.
+
+## npm test
+
+This will run all of our tests via grunt. This spins up a local Dynamo instance, runs mocha, and closes the Dynamo instance at completion.
+
+## npm run lint
+
+This will run our lint configuration against all of our source files and alert you to any issues.
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -133,7 +133,7 @@ Model.compile = function compile (name, schema, options, base) {
       if (result) {
         if (result.resolve) {
           deferred.resolve(result.resolve);
-          return false;
+          return true;
         } else if (result.reject) {
           deferred.reject(result.reject);
           return false;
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -58,3 +58,15 @@ The Dynamoose Changelog can be found in the [CHANGELOG.md](//github.com/dynamoos
...
 
 The Dynamoose Roadmap can be found in the [ROADMAP.md](//github.com/dynamoosejs/dynamoose/blob/master/ROADMAP.md) file. Help is always appreciated on these items. If you are able to help submit a PR so we can review and improve Dynamoose!
+
+## Development
+
+Dynamoose has a few helpers for local development. These are included in the scripts section of our `package.json`.
+
+## npm test
+
+This will run all of our tests via grunt. This spins up a local Dynamo instance, runs mocha, and closes the Dynamo instance at completion.
```
fishcharlie:
```suggestion
This will run all of our tests via grunt. This spins up a local DynamoDB Local instance, runs Mocha, and closes the DynamoDB Local instance at completion.
```

fishcharlie:
> As another side note, the plugin docs have the wrong event names included. Should I fix that here, or as a separate docs only commit?

Separate docs only PR would be best!

fishcharlie:
@hweeks Also! If you are creating public plugins I would love to see them!! This plugins project is in very early stages. I haven't heard about a lot of interest in it so far. I'm glad to see you are interested in it. Please let me know how I can support your work in this area!!

Also in this GitHub organization there is a repo for plugin ideas. Not sure exactly how you are using it, but thought I'd bring it up in case you are interested.

PR #570, opened 2019-02-10
Fixing Deployment

fishcharlie:
### Summary:

This PR should fix the problem where multiple jobs were trying to deploy.

- [x] Yes
- [ ] No

### Other:


travis-deploy
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -ev
+npm install
```
hweeks: we should run the tests here too, always wanna be triple sure :)

fishcharlie: @hweeks The default example from semantic-release looks like, with the important part being: `Advanced: optionally overwrite your default `script` step to skip the tests script: skip`. Are we sure we want to rerun the tests here? I think they should already run as part of the job as well.

```
language: node_js

node_js:
  - 8
  - 6

jobs:
  include:
    # Define the release stage that runs semantic-release
    - stage: release
      node_js: lts/*
      # Advanced: optionally overwrite your default `script` step to skip the tests
      # script: skip
      deploy:
        provider: script
        skip_cleanup: true
        script:
          - npx semantic-release
```

hweeks: if that's how travis will execute that's fine. good point

PR #573, opened 2019-02-10
Attribute Improvements

fishcharlie:
### Summary:

This PR reworks out some of the attributes logic works.

- [ ] Yes
- [x] No

### Other:


lib/Attribute.js
```diff
@@ -196,42 +197,54 @@ Attribute.prototype.types = {
...
+// No type defined - assume this is not a type definition and we must grab type directly from value.
+// Examples of `value` parameters that might be passed in:
+// { type: [Function: String], required: true, default: 'Mittens' }
+// function String() { [native code] }
+// { type: [Function: Number], validate: [Function: validate], default: 888 }
+// 1
+// false
+// "puddles"
+// [ 'cute', 'fuzzy' ]
+// { name: 'mittens', age: 1 }
```
hweeks: Consider switching this doc over to the jsDoc standard.

[the standard](http://usejsdoc.org/about-getting-started.html)

lib/Attribute.js
```diff
@@ -196,42 +197,54 @@ Attribute.prototype.types = {
...
-    let found = typeVal.toString().match(regexFuncName);
-    type = found[1];
-    if (type === 'Object') {
-      type = 'Map';
-    }
-  } else if (typeof typeVal === 'object' || typeVal === 'map'){
-    type = 'Map';
+  const typeVal = value.type || value;
+
+  if (typeof typeVal === 'function') {
```
hweeks: Could we use any of the lodash helpers here?

[lang helpers](https://lodash.com/docs/4.17.11#isArray)

lib/utils/isDynamoObject.js
```diff
@@ -0,0 +1,9 @@
+const AWS = require('aws-sdk');
+
+module.exports = (item) => {
+  if (AWS.DynamoDB.Converter.output(item)) {
```
hweeks: why not just return this check? An `if else` that returns a `Boolean` it is generally cleaner to return the condition.

test.js
```diff
@@ -0,0 +1,5 @@
+const AWS = require('aws-sdk');
+
+console.log(AWS.DynamoDB.Converter.output({'NS': [1,2,3]}));
+console.log(new Set([1,2,3]));
+console.log(AWS.DynamoDB.Converter.input([1,2,3]));
```
hweeks: Why is this file in the root directory?

lib/Attribute.js
fishcharlie: @hweeks `lodash` is not a dependency of this project. It was removed in a PR a decent amount of time ago I think.

lib/utils/isDynamoObject.js
fishcharlie: @hweeks Good point!

test.js
fishcharlie: Wasn't sure if I was gonna work on this PR further before merging. So I kept it for now as a test file.

lib/Attribute.js
fishcharlie: @hweeks This is an internal method. I'm not sure how that works with jsdoc. Users of Dynamoose shouldn't need to use this at all.

lib/Attribute.js
hweeks: i was thinking it may be worth using for these types of checks. it's more robust than a regex, and generally I'd have more faith in it. On top of that, this lib is almost always running on node, so the bundle size isn't as vital.

lib/Attribute.js
hweeks: In the new doc standard, we auto build lib docs. Having the JSDoc comments makes it easier for other devs to onboard, and reduces our double documentation writing efforts.

fishcharlie: @hweeks That makes sense. But is there a way to designate that this is an internal method? I don't think it'd be valuable to have our documentation cluttered with a bunch of internal methods that users won't ever have to use. If you are a new user to Dynamoose, on-boarding should be as easy as possible. Having to read through documentation that is filled with a bunch of internal methods that you don't need is kinda a waste.

lib/Attribute.js
fishcharlie: @hweeks I don't think this will work:

```
> _.isNumber(Number)
false
> _.isNumber(5)
true
```

lib/Attribute.js
hweeks: yup, [inner](http://usejsdoc.org/tags-inner.html)

lib/Attribute.js
hweeks: ahh, i misread what this was doing. I got it now. I was using lodash elsewhere in a plugin i was writing. Lemme read through this more thoroughly

lib/Attribute.js
fishcharlie: @hweeks Check out my latest commit and let me know your thoughts.

PR #581, opened 2019-02-15
Fix: Stop timestamp conditions being overwritten

tgandrews:
Condition values referencing timestamps could be overwritten by both
when either updateTimestamps flag was on or by the set method. This
would cause conditions to fail.

The change disables these modification paths in `Attribute.toDynamo` for
condition values.

### Other:

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -24,7 +24,7 @@ function Model (obj) {
   }
 }
 
-async function processCondition (req, options, schema) {
+async function processCondition (req, options, model) {
   if (options.condition) {
     if (req.ConditionExpression) {
       req.ConditionExpression = `(${req.ConditionExpression}) and (${options.condition})`;
@@ -45,9 +45,9 @@ async function processCondition (req, options, schema) {
         const k = keys[i];
 
         const val = options.conditionValues[k];
-        const attr = schema.attributes[k];
+        const attr = model.$__.schema.attributes[k];
         if (attr) {
-          req.ExpressionAttributeValues[`:${k}`] = await attr.toDynamo(val);
+          req.ExpressionAttributeValues[`:${k}`] = await attr.toDynamo(val, undefined, model, {'updateTimestamps': false});
         } else {
           throw new errors.ModelError(`Invalid condition value: ${k}. The name must either be in the schema or a full DynamoDB object must be specified.`);
         }
@@ -419,7 +419,7 @@ Model.conditionCheck = async function (NewModel, key, options, next) {
       const rangeKeyName = schema.rangeKey.name;
       conditionReq.Key[rangeKeyName] = schema.rangeKey.toDynamo(key[rangeKeyName], undefined, key);
     }
-    await processCondition(conditionReq, options, schema);
+    await processCondition(conditionReq, options, NewModel);
 
     debug('Condition Check', conditionReq);
     deferred.resolve(conditionReq);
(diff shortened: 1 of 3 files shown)


lib/Model.js
```diff
@@ -420,7 +420,7 @@ Model.conditionCheck = async function(NewModel, key, options, next) {
       let rangeKeyName = schema.rangeKey.name;
       conditionReq.Key[rangeKeyName] = schema.rangeKey.toDynamo(key[rangeKeyName], undefined, key);
     }
-    processCondition(conditionReq, options, schema);
+    processCondition(conditionReq, options, NewModel);
```
fishcharlie: I just realized, should this be `await`'ed?

lib/Model.js
```diff
@@ -45,9 +45,9 @@ async function processCondition(req, options, schema) {
         const k = keys[i];
 
         const val = options.conditionValues[k];
-        const attr = schema.attributes[k];
+        const attr = model.$__.schema.attributes[k];
         if(attr) {
-          req.ExpressionAttributeValues[`:${k}`] = await attr.toDynamo(val);
+          req.ExpressionAttributeValues[`:${k}`] = await attr.toDynamo(val, undefined, model, { ...options, updateTimestamps: false });
```
fishcharlie: Do we really want to be passing in options here?

lib/Model.js
hweeks: this is a good catch.

lib/Model.js
```diff
@@ -45,9 +45,9 @@ async function processCondition(req, options, schema) {
         const k = keys[i];
 
         const val = options.conditionValues[k];
-        const attr = schema.attributes[k];
+        const attr = model.$__.schema.attributes[k];
```
hweeks: have you seen a miss in this pattern else where?

tgandrews: Not that I've noticed but I did see two instances of what looks like an incorrect call to `toDynamo`.

https://github.com/dynamoosejs/dynamoose/blob/9fd901a4a1755cdd19cdc808276a6895c8cc9434/lib/Model.js#L785
and
https://github.com/dynamoosejs/dynamoose/blob/9fd901a4a1755cdd19cdc808276a6895c8cc9434/lib/Model.js#L1527

tgandrews:
@fishcharlie and @hweeks I have pushed up the changes requested

tgandrews:
I had to increase the lint limit for functions so that the tests would allowed and I don't feel if I'm the right person to be making refactors to break Model or at leasts the tests into smaller pieces.

tgandrews:
I'm also getting CI failures for coveralls saying:
```
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```

Is this a known thing?

fishcharlie:
@tgandrews That is fine about the linter. The CI errors is a known error. I don't know how to fix it. See https://github.com/lemurheavy/coveralls-public/issues/632#issuecomment-464542765.

I'll try to review this today.

tgandrews:
@fishcharlie what needs to happen for this to be merged?

fishcharlie:
@tgandrews Was gonna wait until @hweeks re-reviewed. But will merge now.

tgandrews:
Thanks @fishcharlie for the quick response on this.

fishcharlie:
@tgandrews Of course!

Issue #583, opened 2019-02-18
Validate is called on Model.get


Tirke:
### Summary:

`Model.get()` is calling validate on attributes but I don't think it's necessary (and it's breaking our way to validate complex models). I've written a test showing this behavior.

### Code sample:
```javascript
const model = dynamoose.model('no-validate', new dynamoose.Schema(
      {
        id: {
          type: String,
          required: true,
          hashKey: true,
          validate: function (val, model) {
            if (!model.name){
              // Is called on get :(
              throw new Error('should not throw');
            }
            return true;
          }
        },
        projectId: {
          type: String,
          required: true,
          rangeKey: true,
          index: {
            global: true,
            name: 'projectId-index',
            project: true,
          },
        },
      }));


    let error, res;
    try {
      await model.create({id: 'idid', projectId:'project', name: 'Rob'});
      res = await model.get({id: 'idid', projectId: 'project'});
    } catch (e) {
      error = e;
    }
    should.not.exist(error);
    res.id.should.eql(1);
    res.name.should.eql('Rob');
```

### Current output and behavior:

Validate is called when calling `Model.get()`

### Expected output and behavior:

Validate shouldn't be called when calling `Model.get()`

### Environment:

Operating System: MacOS Mojave
Operating System Version: latest
Node.js version (`node -v`): 8.15.0
NPM version: (`npm -v`): 6.4.1
Dynamoose version: latest
Dynamoose Plugins: none

-

### Dynamoose Plugins:

- [ ] Yes, I believe that one or multiple 3rd party Dynamoose plugins are effecting this issue
- [ ] No, I believe this issue is independent of any 3rd party Dynamoose plugins I'm using
- [ ] Unknown, I'm unsure if Dynamoose plugins are effecting this issue
- [x] I am not using any Dynamoose plugins

### Type (select 1):
- [x] Bug report
- [ ] Feature suggestion
- [ ] Question
- [ ] Other suggestion
- [ ] Something not listed here

### Other:
- [x] I have read through the Dynamoose documentation before posting this issue
- [x] I have searched through the GitHub issues (including closed issues) and pull requests to ensure this issue has not already been raised before
- [x] I have searched the internet and Stack Overflow to ensure this issue hasn't been raised or answered before
- [x] I have tested the code provided and am confident it doesn't work as intended
- [x] I have ensured that all of my plugins that I'm using with Dynamoose are listed above
- [x] I have filled out all fields above
[...]

PR #584, opened 2019-02-18
fix: model.get is calling validate when it should not

Tirke:
re #583

- [ ] Yes
- [x] No

### Other:

### DIFF
--- a/test/Model.js
+++ b/test/Model.js
@@ -3574,4 +3574,44 @@ describe('Model', function (){
       should.not.exist(result);
     });
   });
+  
+  it('should not call validate on the model when calling get', async function(){
+    const model = dynamoose.model('no-validate', new dynamoose.Schema(
+      {
+        id: {
+          type: String,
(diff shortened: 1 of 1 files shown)


hweeks:
@fishcharlie any movement on this concept? is there a reason why we need to validate on both put and get?

fishcharlie:
@hweeks Parity with mongoose I think is important in this case. That being said, just from a pure logic standpoint, I'm of the opinion that we should validate in both places. Seems like an extra safe guard in case the data gets updated in DynamoDB directly, without Dynamoose.

If someone wants to actually implement this, feel free. I need to get around to confirming Mongoose behavior. After I confirm that I'll merge it in. Needs to be implemented tho as well.

hweeks:
I believe I sent you the code in Mongoose in slack. Double check that link

fishcharlie:
@hweeks Can’t find it.

PR #598, opened 2019-03-18
Ts model

hweeks:
### Summary:

This adds typescript to existing Model, and breaks down the major bits. I've pulled out the libs, model and Model from each other to reduce overall file size, but not affect readability.

### Other information:

This has no new code, it's a refactor in place.

- [ ] Yes
- [x] No

### Other:


lib/Dynamoose.ts
```diff
@@ -141,7 +142,7 @@ class Dynamoose {
       schema = new Schema(schema, options);
     }
 
-    const model = Model.compile(name, schema, options, this);
+    const model = compile(name, schema, options, this);
```
fishcharlie: Hmm, not a huge fan of this. Doesn't really say what it is compiling. Too broad in my opinion.

lib/Model.ts
```diff
@@ -1,20 +1,16 @@
-'use strict';
-
-const util = require('util');
-const Q = require('q');
-const hooks = require('hooks');
-const Attribute = require('./Attribute');
-const Table = require('./Table');
```
fishcharlie: Where did this get moved to?

---

Oh I see, got moved into NewModel. Totally fine.

lib/Model.ts
```diff
@@ -1,20 +1,16 @@
...
-const Query = require('./Query');
-const Scan = require('./Scan');
-const Plugin = require('./Plugin');
```
fishcharlie: Where did this get moved to?

---

Oh I see, got moved into NewModel. Totally fine.

lib/ModelLibs.ts
```diff
@@ -0,0 +1,155 @@
...
+      }
+
+      deferredBatch.resolve(data);
+    });
+
+    return deferredBatch.promise;
+  });
+
+  return Q.all(batchList).then((resultList) => reduceBatchResult(resultList));
+}
```
fishcharlie: New line at end of files. Is there a lint rule we can add to ensure this happens?

```suggestion
}

```

lib/NewModel.ts
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    table.init((err) => {
+      if (err) {
+        throw err;
+      }
+    });
+
+    return NewModel;
+  }
```
fishcharlie: New line at end of files.

```suggestion
  }
  
```

fishcharlie: Also, why is this indented? Are we missing a closing `}` somewhere?

lib/Model.ts
```diff
@@ -1215,7 +859,7 @@ query('id').eq(1).where('name').begienwith('foxy')
 Model.query = function (NewModel, query, options, next) {
   if (typeof options === 'function') {
     next = options;
-    options = null;
+    options = undefined;
```
fishcharlie: Why are we switching this to undefined? It's done in a few other places also.

tslint.json
```diff
@@ -1,61 +1,61 @@
...
+        "variable-declaration": "onespace"
+      }
+    ],
+    "no-internal-module": true,
+    "no-trailing-whitespace": true,
+    "no-null-keyword": true,
+    "prefer-const": true,
+    "jsdoc-format": true
+  }
 }
```
fishcharlie: Is there a way to have this be stricter? There seems to be linting stuff in ESLint that was being caught that isn't here. Some of which I commented on.

fishcharlie:
@hweeks This also needs to have `master` merged into it.

lib/Dynamoose.ts
hweeks: With TS it will describe the output in fine detail and allow end users to know exactly what is happening. Do you want a more descriptive name?

lib/ModelLibs.ts
hweeks: I think so, tslint is a subset of eslint.

lib/NewModel.ts
hweeks: it's not reformatted completely yet

lib/Model.ts
hweeks: undefined over null is a tsconfig setting i've always used, it's mostly my preference as undefined is the default and null is an assigned value.

PR #602, opened 2019-03-19
fix(schema): fixed circular reference with `JSON.stringify` in Schema

Tirke:
### Summary:

When trying to delete a instance of a model with the static `Model.delete` method, dynamoose will throw a `TypeError: Converting circular structure to JSON`.

The culprit being :
https://github.com/dynamoosejs/dynamoose/blob/cedea45b1bfc07f899de871c76da4ea0512a4a64/lib/Schema.js#L215

If I understand thing correctly this is used as the standard trick to clone an object. Trying to `stringify` an object with circular references will throw an error.

The fix I'm suggesting is taken straight from SO. We can also use something like [cycle.js](https://github.com/douglascrockford/JSON-js/blob/master/cycle.js).
Another possibility is to use a real `clone` library. 
I can make any change you want, just tell me.

- [x] Yes
- [ ] No

### Other:


hweeks:
Are there any other places this should be implemented? Very clean commit btw, thanks for the input!

Tirke:
We're not seeing the bug in other places in our application.
But I can imagine this could be potentially problematic
https://github.com/dynamoosejs/dynamoose/blob/cedea45b1bfc07f899de871c76da4ea0512a4a64/lib/Attribute.js#L164-L168

fishcharlie:
Someone wanna look at f8a9fd9 to make sure I fixed merge conflicts correctly before I merge?

PR #619, opened 2019-04-26
fix(model): allows string for array attribute in contains condition

AndreasEK:
### Summary:
When using the contains function for a conditional update, dynamoose requires the value to be an array, while it needs to be a string instead.

Dynamoose deduces from the attribute type being an array, that the value needs to be an array, too. This is not the case for the contains function. See https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html

### GitHub linked issue:
Closes https://github.com/dynamoosejs/dynamoose/issues/618

- [X] Yes
- [ ] No

### Other:

### DIFF
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -47,7 +47,8 @@ async function processCondition (req, options, model) {
         const val = options.conditionValues[k];
         const attr = model.$__.schema.attributes[k];
         if (attr) {
-          req.ExpressionAttributeValues[`:${k}`] = await attr.toDynamo(val, undefined, model, {'updateTimestamps': false});
+          const noSet = options.condition.startsWith('contains');
+          req.ExpressionAttributeValues[`:${k}`] = await attr.toDynamo(val, noSet, model, {'updateTimestamps': false});
         } else {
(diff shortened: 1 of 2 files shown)


lib/Model.js
```diff
@@ -47,7 +47,8 @@ async function processCondition (req, options, model) {
         const val = options.conditionValues[k];
         const attr = model.$__.schema.attributes[k];
         if (attr) {
-          req.ExpressionAttributeValues[`:${k}`] = await attr.toDynamo(val, undefined, model, {'updateTimestamps': false});
+          const noSet = options.condition.startsWith('contains');
```
fishcharlie: Any opinion on this? Is this really the safest way to do this? Seems like it could create edge cases. I could be wrong tho.

hweeks: This _should_ be fine. I can't really imagine too many words that start with contains.

AndreasEK: There might be more edge cases (like, more cases in which the attribute should NOT be a set), but at least it catches this one. Also, from reading the documentation, I did not find another case. One could change it to check for 'contains(' instead of 'contains', but I found it sufficient like this. As @hweeks said, there are no other valid conditions that start with 'contains'.

PR #625, opened 2019-05-09
Debug casts object to string

purefan:
### Summary:
Very tiny improvement, just a log that prints an object as a string, I doubt this is the intended behavior because it's not very useful in my opinion, literally a one character change

- [X] Yes
- [ ] No


fishcharlie:
@purefan Please follow our [Contributing Guide](https://github.com/dynamoosejs/dynamoose/blob/master/CONTRIBUTING.md) your comment message needs to be formatted correctly in order to handle releases.

PR #626, opened 2019-05-09
feat(plugin): added item prop to event object passed to model:put put:called listener.

dolsem:
### Summary:
I've added an additional prop to the event object that is passed to plugin callbacks during `put:called` stage of  `model:put` event. The prop is `item` and it is equal to the model class instance that is being put into the database. I need this for the [optimistic locking plugin](https://github.com/dynamoosejs/dynamoose-plugin-ideas/issues/3), and I imagine it can be useful to other plugin creators as well. One use case is if we want to get updatedAt timestamp of the item being saved before the timestamp gets updated by Dynamoose.

### Other information:
As I was modifying the code I split `NewModel._emit()` call statements into multiple lines for better readability, and updated style of every statement so that they match. Hopefully you will find this tiny change useful. I also threw in extra tests for emit object properties for every stage of the `model:put` event.

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/docs/api/plugins.md
+++ b/docs/api/plugins.md
@@ -244,7 +244,7 @@ Dynamoose will emit this event when a put is called on a model.
 	event: {
 		options: _____, // options passed to put (object) (warning: in some cases this can be the callback function in the `put:called` stage)
 		callback: _____, // callback passed to put (function) (warning: in some cases this can be the null if options is not passed in) (only valid on `put:called`)
-		item: _____ // item that will be/has been saved to DynamoDB (function) (only valid on `request:*`)
+		item: _____ // item that will be/has been saved to DynamoDB (object) (model class instance on `put:called`, PutItem request object on `request:*`)
 		error: _____ // the error object that was received from DynamoDB (object) (only valid on `request:post`)
 	}
 	action: {
--- a/lib/Model.js
+++ b/lib/Model.js
@@ -433,7 +433,7 @@ Model.prototype.put = async function (options, next) {
   debug('put', this);
   const deferred = Q.defer();
   const emitObjectPutCalled = {
-    'event': {'callback': next, options},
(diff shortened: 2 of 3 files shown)


dolsem:
@fishcharlie any update on this? Would be nice to get a review soon, since optimistic locking plugin depends on it, and optimistic locking is crucial in production.

fishcharlie:
@dolsem Please keep in mind plugin support for Dynamoose is **very much** still in beta. @hweeks and I were talking about reworking the entire system to be more in line with Mongoose. But it was very much just initial discussions. Nothing concrete at all.

Other than that with that one review comment I had, I'll merge.

Sorry for the delay in looking over this, been super busy with other stuff.

dolsem:
@fishcharlie okay, no problem.

dolsem:
> There are a lot of cases where function calls are seperated onto two lines. Can we make this all one line.
> 
> Example:
> 
> ```js
> emitObject.should.have
> .propertyByPath('event', 'type').which.is.eql('model:put');
> ```
> 
> Should all be on one line.

I personally find code that does not have long lines more readable, but sure, if that's what you require.

hweeks:
Looks like a good addition to functionality and well written. I’m cool with this post style cleanup.

fishcharlie:
@dolsem In my opinion, it makes it harder to know where to split the line. I mean should you have split it before `which`? It'll never be consistent, and everyone will handle it slightly differently. Text wrap for text editors does the job of making it easier to read.

At the end of the day it's 100% personal opinion. I don't think we do multiline function calls anywhere else in the project, so I think the consistency argument has to win out in this case.

fishcharlie:
@dolsem Do me a favor and when you fix those changes please be sure to tag me in a GitHub comment on this PR so I can re-review and merge.

dolsem:
@fishcharlie should be all set

PR #634, opened 2019-06-03
Feature/typings update

C45tr0:
### Summary:
Adding some options that were missing in the typings. And fixing a few that said they were required, but not in the docs.

PR #629 - Does have most of my changes plus some other renaming/reording. This also has the addition to streamOptions. I can rebase off of that or whatever is desired.

### Is this ready to be merged into Dynamoose? (select 1):
- [x] Yes
- [ ] No

### Other:
- [~] I have searched through the GitHub pull requests to ensure this PR has not already been submitted
- [ ] I have updated the Dynamoose documentation (if required) given the changes I made
- [ ] I have added/updated the Dynamoose test cases (if required) given the changes I made
- [x] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoosejs/dynamoose/blob/master/LICENSE.txt)
- [x] All of my commits and commit messages are detailed, explain what changes were made, and are easy to follow and understand
- [x] I have filled out all fields above

PR #638, opened 2019-06-17
fix(types): added streamOptions to ModelOption

dolsem:
### Summary:
Dynamoose typings file does not include `streamOptions` field inside `ModelOption` interface, which is used for model options supplied to `dynamoose.model()`. This PR adds that field.

#634 is already adding that field, but is using obscure `object` type instead, which can lead to runtime errors.

- [x] Yes
- [ ] No

### Other:


dolsem:
Seems like test suite is failing because of an unrelated issue with transactions.

PR #664, opened 2019-08-09
feat: use FilterExpression strings in Query

repl-sean-heintz:
### Summary:

Currently Queries (and Scans) are using the legacy QueryFilter API to perform post-query filtering. 
This PR changes the existing filter implementation for Query to use the FilterExpression attribute instead of QueryFilter. While I tried to make this change as minimal as possible, the following operations are now possible when filtering attributes:

1. Using NOT before BETWEEN
2. Using NOT before BEGINS_WITH
3. Using NOT before IN

Please note that this PR does **not** address the following:
1. Using FilterExpression in `scan` operations. I attempted this, but some very strange patterns in the Scan code made me abandon ship before getting anywhere. It seems like there's some skeletons buried in there.

2. Using `KeyConditionExpression` instead of `KeyConditions`. I'd imagine that the same `FilterExpression` logic could be used to construct this, however. Worth looking in to on a future PR.

### Code sample:

There are no changes to the exposed Dynamoose APIs.

### GitHub linked issue:

#---

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/lib/FilterExpression.js
+++ b/lib/FilterExpression.js
@@ -1,0 +1,148 @@
+'use strict';
+const AWS = require('aws-sdk');
+const errors = require('./errors');
+
+function FilterExpressionNode (operation, attribute, value) {
+  this.operation = operation;
+  this.attribute = attribute;
+  this.value = value;
+
+  this.prev = null;
+  this.next = null;
+}
+
+function FilterExpression () {
+  this.head = null;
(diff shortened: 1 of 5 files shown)


repl-sean-heintz:
Noticed some issues while leveraging this in my own project. Will re-open once I've resolved the issues.

PR #682, opened 2019-09-28
fix: fixing cannot read property 'type' of null

quan-qodespace:
Fixes a bug where a nested object property is null, this would cause this line to error because typeof null returns 'object' but obj.type would cause an exception.

### Summary:
Cannot read property 'type' of null error when working with saveUnknown is true and contains nested array of objects and when there's a null property.

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -191,7 +191,7 @@ Attribute.prototype.setTypeFromRawValue = function (value) {
   // no type defined - assume this is not a type definition and we must grab type directly from value
   let type;
   let typeVal = value;
-  if (value.type) {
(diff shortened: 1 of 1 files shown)


teamclicker:
Am seeing this also. Any workaround until this gets merged?

Gabiras12:
@fishcharlie can we merge this? 
- try to save a unknown Attribute 
- unknown attribute is an object with a property with value null 
- Cannot read property 'type' of null

fishcharlie:
@Gabiras12 Nope. This PR was submitted against v1. v1 is no longer maintained. Please upgrade to v2.

PR #689, opened 2019-11-09
Fix model delete with timestamp rangekey

### LINKED ISSUE #607: Model can't delete when rangeKey is createdAt
### Summary:

When call Model.delete or model.delete, it happens error and can't delete.

#### Schema
```
schema: {
    createdAt: {
      rangeKey: true,
    },
    schemaOptions: {
        timestamps: true,
    },
}
```

#### General

Error happens when call
```
Model.delete({id:'aaaa',createdAt: 1553137260667})
```
or
```
cosnt model = await Model.get({id:'aaaa',createdAt:1553137260667})
model.delete()
```

```
TypeError: Cannot read property 'updateTimestamps' of undefined
    at Attribute.toDynamo (/node_modules/dynamoose/dist/Attribute.js:390:30)
    at NewModel.Model.delete (/node_modules/dynamoose/dist/Model.js:1007:65)
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

Maybe, in Attribute.js 

```
Attribute.prototype.toDynamo = async function (val, noSet, model, options) {
...
 if (isTimestamp && options.updateTimestamps === false) {
        runSet = false;
    }
    if (isExpires && options.updateExpires === true) {
        val = await this.default();
        runSet = true;
    }

```
this will change as below?

```
Attribute.prototype.toDynamo = async function (val, noSet, model, options) {
...
 if (isTimestamp && options && options.updateTimestamps === false) {
        runSet = false;
    }
    if (isExpires && options && options.updateExpires === true) {
        val = await this.default();
        runSet = true;
    }

```

### Environment:

Node.js version (`node -v`): 8.10
Dynamoose version: 1.7.2

### Dynamoose Plugins:

- [ ] Yes, I believe that one or multiple 3rd party Dynamoose plugins are effecting this issue
- [ ] No, I believe this issue is independent of any 3rd party Dynamoose plugins I'm using
- [ ] Unknown, I'm unsure if Dynamoose plugins are effecting this issue
- [x] I am not using any Dynamoose plugins

### Type (select 1):
- [x] Bug report
- [ ] Feature suggestion
- [ ] Question
- [ ] Other suggestion

### Other:
- [ ] I have read through the Dynamoose documentation before posting this issue
- [ ] I have searched the internet and Stack Overflow to ensure this issue hasn't been raised or answered before
- [ ] I have tested the code provided and am confident it doesn't work as intended
- [ ] I have ensured that all of my plugins that I'm using with Dynamoose are listed above
- [ ] I am running the latest version of Dynamoose

dolsem:
### GitHub linked issue:

Closes #607 

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/lib/Attribute.js
+++ b/lib/Attribute.js
@@ -419,11 +419,11 @@ Attribute.prototype.toDynamo = async function (val, noSet, model, options) {
     isExpires = true;
   }
 
-  if (isTimestamp && options.updateTimestamps === false) {
+  if (isTimestamp && (!options || options.updateTimestamps === false)) {
     runSet = false;
   }
 
-  if (isExpires && options.updateExpires === true) {
+  if (isExpires && options && options.updateExpires === true) {
     val = await this.default();
     runSet = true;
   }
--- a/test/Model.js
+++ b/test/Model.js
@@ -1133,9 +1133,12 @@ describe('Model', function () {
     });
   });
 
+  it('Delete item with timestamp rangeKey', async () => {
+    const cat = await Cats.CatWithTimestampRangeKey.create({'ownerId': 1});
+    await cat.delete();
+  });
+
   it('Get missing item', (done) => {
-
-
     Cats.Cat.get(1, (err, model) => {
       should.not.exist(err);
       should.not.exist(model);
--- a/test/fixtures/Cats.js
+++ b/test/fixtures/Cats.js
@@ -394,6 +394,20 @@ module.exports = function (dynamoose) {
     'defaultReturnValues': 'NONE'
   });
 
+  const CatWithTimestampRangeKey = dynamoose.model('CatWithTimestampRangeKey', {
+    'ownerId': {
+      'type': Number,
+      'hashKey': true
+    },
+    'createdAt': {
+      'type': Date,
+      'rangeKey': true
+    },
+    'name': String
+  }, {
+    'timestamps': true
+  });
+
   return {
     Cat,
     Cat1,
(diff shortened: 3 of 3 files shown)


test/Model.js
```diff
@@ -1133,9 +1133,12 @@ describe('Model', function () {
     });
   });
 
-  it('Get missing item', (done) => {
-
+  it('Delete item with timestamp rangeKey', async () => {
```
fishcharlie: Are we sure this test fails without the changes you made?

lib/Attribute.js
```diff
@@ -419,11 +419,11 @@ Attribute.prototype.toDynamo = async function (val, noSet, model, options) {
     isExpires = true;
   }
 
-  if (isTimestamp && options.updateTimestamps === false) {
+  if (isTimestamp && (!options || options.updateTimestamps === false)) {
```
fishcharlie: I’m on my phone so can’t look at this super in depth. Is this changing existing behavior tho? Would this be considered a breaking change? Like if the user doesn’t pass in options, it looks like this might be changing the behavior of that.

test/Model.js
dolsem: Yes. That's why I always make first commit with tests only - it can be checked out to compare the effect of code changes on added tests.

lib/Attribute.js
dolsem: This function doesn't normally get called by user directly, and currently if it is called without passing options, this expression throws error when trying to access `updateTimestamps` field on `undefined`. I also made sure to update these expressions in a way that not passing options results in the same behavior as passing options without `updateTimestamps` and `updateExpires` specified, so that if a call to this function broke in some other place, it would now be fixed with correct behavior.

dolsem: I've just found one such "other place" and mentioned it in #384.

PR #726, opened 2020-03-09
dynamoose.undefined

fishcharlie:


dolsem:
I'm curious, why a special symbol instead of just using `null`?

fishcharlie:
@dolsem `null` is a specific DynamoDB type. I haven't quite thought through how Dynamoose should implement it, or use cases for that yet. But the idea behind using a custom symbol vs `null` is just to maintain flexibility in the future for adopting the `null` DynamoDB type. Not saying that we'll use the default `null` value for that, but wanted to provide that flexibility if needed.

It also does feel more like `undefined` as opposed to `null`. `null` to me more means that it's purposefully set to nothing, whereas `undefined` means it wasn't even set in the first place. In the document being saved, the result will be `undefined`. The only difference for `dynamoose.undefined` is that it will bypass the `default` attribute setting.

What are your thoughts? It still is in beta, so I'd love some discussion around this if you feel like there is a better way.

dolsem:
Idk, this API feels a bit weird to me, a bit extra. First of all, I can't think of a good use case that needs to treat absent fields and null values as two distinct scenarios, I would call it bad design, because it results in confusion that can be avoided. Second, if I design a schema that has a field with a default value, I would expect that field to be present on every record, and for that reason I would always prefer to set it to null explicitly, as opposed to not defining it.

If you are confident you want to have this for the sake of maximum API flexibility, I would at least suggest using upper case instead, to signify it's a special constant and to avoid any possible collision with the built-in `undefined`, e.g. to allow destructuring, like `import { UNDEFINED } from 'dynamoose';`.

fishcharlie:
@dolsem 

> First of all, I can't think of a good use case that needs to treat absent fields and null values as two distinct scenarios, I would call it bad design, because it results in confusion that can be avoided.

I totally agree. Although it is important to note, that [DynamoDB has a `null` type](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes.Null) that is different than just not having the attribute in the document being stored. At the end of the day, even if I disagree with it, it is something that I think we need to plan for supporting, and abide by DynamoDB design patterns.

> Second, if I design a schema that has a field with a default value, I would expect that field to be present on every record, and for that reason I would always prefer to set it to null explicitly, as opposed to not defining it.

This is a tough one. Currently how Dynamoose has been treating TTL is by using the `default` setting to set the TTL when the document is created. I have run into cases in my systems where I no longer want the document to expire. But setting it to `null` or `undefined` will cause the `default` value to just be reset. In Dynamoose v1.x this was handled by using a `expires.defaultExpires` property. Which felt incredibly messy to me. This has been replaced with `dynamoose.undefined` to allow those TTL values to be set to nothing so that the document doesn't expire.

I'd also say, I believe it's **very** important to have separation of concerns in this case. `required` is already an option for attributes. I think saying that `default` should behave and inherit the behavior of `required` doesn't make sense. If you have a property that should have a default and always exist, that property should adopt both `default` & `required`. But I don't think that should be something that Dynamoose should assume.

> If you are confident you want to have this for the sake of maximum API flexibility, I would at least suggest using upper case instead, to signify it's a special constant and to avoid any possible collision with the built-in `undefined`, e.g. to allow destructuring, like `import { UNDEFINED } from 'dynamoose';`.

I'm definitely open to this idea. Few questions. Do you see this as being an alias for `dynamoose.undefined`? Where `dynamoose.undefined === dynamoose.UNDEFINED`? Or just a straight replacement, and remove `dynamoose.undefined`?

I'd also love some more input on this. If anyone else has any other thoughts, please feel free to comment.

dolsem:
> I'd also say, I believe it's very important to have separation of concerns in this case. required is already an option for attributes. I think saying that default should behave and inherit the behavior of required doesn't make sense. If you have a property that should have a default and always exist, that property should adopt both default & required. But I don't think that should be something that Dynamoose should assume.

I completely agree, not saying it should expect the value to be required, I was just giving my two cents as a user with this.

> I'm definitely open to this idea. Few questions. Do you see this as being an alias for dynamoose.undefined? Where dynamoose.undefined === dynamoose.UNDEFINED? Or just a straight replacement, and remove dynamoose.undefined?

I would just refactor it into `dynamoose.UNDEFINED` tbh, so straight replacement.

dolsem:
> This is a tough one. Currently how Dynamoose has been treating TTL is by using the default setting to set the TTL when the document is created. I have run into cases in my systems where I no longer want the document to expire. But setting it to null or undefined will cause the default value to just be reset. In Dynamoose v1.x this was handled by using a expires.defaultExpires property. Which felt incredibly messy to me. This has been replaced with dynamoose.undefined to allow those TTL values to be set to nothing so that the document doesn't expire.

Ah, I see, that makes things more complicated for sure. Whatever you do here will probably feel hacky anyway. What happens if you set TTL to -1 in DynamoDB? Maybe you could treat this as a special value and remove the TTL attribute when you do the value mapping.

fishcharlie:
@dolsem

> Ah, I see, that makes things more complicated for sure. Whatever you do here will probably feel hacky anyway. What happens if you set TTL to -1 in DynamoDB? Maybe you could treat this as a special value and remove the TTL attribute when you do the value mapping.

Yeah, I thought this PR was a better route to go down. It has less custom internal logic, and can be applied to more use cases vs just one. It was something that crossed my mind tho. I just felt like in as many cases as possible, building broad functionality that can be used in many areas is better than building an ultra specific feature that has a very limited case. Not saying that other use cases will be adopted, but it does give more flexibility to allow for that.

I **totally** get why it feels messy tho. Like as a consumer of this package, I would feel the same way. The struggle is, from my perspective, it's the best way to handle it after weighing the pros and cons of everything. On some level, maybe something like `dynamoose.undefinedNoDefault` would be better? That way it's more specific and clear what is going on. But, it does get to be a longer name, and I think there are reasons why being that verbose might be a detriment.

> I completely agree, not saying it should expect the value to be required, I was just giving my two cents as a user with this.

I really appreciate your input on this. It's always good to get more feedback especially in this beta stage.

> I would just refactor it into `dynamoose.UNDEFINED` tbh, so straight replacement.

Got it.

fishcharlie:
#837 is the change to refactor it into `dynamoose.UNDEFINED`

PR #783, opened 2020-04-07
Updated documentation

MickL:
- Corrected error `and` actually was `or`
- Added info to better understand `query.all()` and `scan.all()`
- Added info to better not use scan
- Added info for .query() how it automatically uses hashKey / rangeKey
- Added info everywhere that hashKey is also named partition key
- Added info everywhere that rangeKey is also named sort key
- Renamed document to item, where ever possible without breaking the code or making too much confusion


docs/api/Model.md
```diff
@@ -173,7 +173,7 @@ User.get(1, {"return": "request"}, (error, request) => {
 });
 ```
 
-In the event you have a rangeKey for your model, you can pass in an object for the `hashKey` parameter.
+In the event you have a rangeKey (sort key) for your model, you can pass in an object for the `hashKey` (partition key) parameter.
```
fishcharlie, line 176: I think this should read something like the following.

```suggestion
In the event you have a rangeKey/sort key for your model, you can pass in an object for the `key` parameter which includes the range key/sort key & the hash key/partition key.
```

Then line 127 should be changed to the suggestion above.

docs/api/Model.md
```diff
@@ -126,7 +126,7 @@ async function printTableRequest() {
 
 ## Model.get(hashKey[, settings][, callback])
```
fishcharlie, line 127:
```suggestion
## Model.get(key[, settings][, callback])
```

See suggestion below.

docs/api/Model.md
```diff
@@ -278,7 +278,7 @@ User.batchGet([1, 2], {"return": "request"}, (error, request) => {
 });
 ```
 
-In the event you have a rangeKey for your model, you can pass in an object for the `hashKey` parameter.
+In the event you have a rageKey (sort key) for your model, you can pass in an object for the `hashKey` (partition key) parameter.
```
fishcharlie, line 281: This should read the same as above.

```suggestion
In the event you have a range key/sort key for your model, you can pass in an object for the `key` parameter which includes the range key/sort key & the hash key/partition key.
```

docs/api/Model.md
```diff
@@ -530,7 +530,7 @@ User.delete(1, {"return": "request"}, (error, request) => {
...
```
fishcharlie, line 533:
```suggestion
In the event you have a range key/sort key for your model, you can pass in an object for the `key` parameter which includes the range key/sort key & the hash key/partition key.
```

docs/api/Model.md
```diff
@@ -483,7 +483,7 @@ The `validate` Schema attribute property will only be run on `$SET` values. This
 
 ## Model.delete(hashKey[, settings][, callback])
```
fishcharlie, line 484:
```suggestion
## Model.delete(key[, settings][, callback])
```

docs/api/Model.md
```diff
@@ -644,7 +644,7 @@ User.batchDelete([1, 2], {"return": "request"}, (error, request) => {
 });
 ```
 
-In the event you have a rangeKey for your model, you can pass in an object for the `hashKey` parameter.
+In the event you have a rageKey (sort key) for your model, you can pass in an object for the `hashKey` (partition key) parameter.
```
fishcharlie, line 647:
```suggestion
In the event you have a range key/sort key for your model, you can pass in an object for the `key` parameter which includes the range key/sort key & the hash key/partition key.
```

docs/api/Query.md
```diff
@@ -4,15 +4,17 @@ Dynamoose provides the ability to query a model by using the `Model.query` funct
...
+Cat.query({"breed": {"contains": "Terrier"}}).exec() // will query all items where the hashKey (partition key) `breed` contains `Terrier`
 ```
 
 If you pass an object into `Model.query` the object for each key should contain the comparison type. For example, in the last example above, `contains` was our comparison type. This comparison type must match one of the comparison type functions listed on this page.
 
+**Please note:** `Model.query()` combines both the `KeyConditionExpression` and the `FilterExpression` from DynamoDB. If you query for an attribute that you defined as your hashKey (partition key) or rangeKey (sort key) DynamoDB will use `KeyConditionExpression`. This is the most performant and cost efficient way and we highly recommend setting hashKey (partition key) and rangeKey (sortKey) that you can query for. If you query for attributes that are not defined as your hashKey (partition key) or rangeKey (sortKey) DynamoDB might select more items at first which could have a bad impact on performance and costs. Also consider using secondary indexes instead.
```
fishcharlie, line 16: > **Please note:**

Not sure I'm a fan of the bold here. We can add it later if needed. But it adds unnecessary styling, and draws peoples attention to that. If you are new to DynamoDB and don't understand this concept, you should likely be reading most/all of the documentation. I don't think it's necessary to draw additional attention to this point and risk people skipping or skimming over other valuable information.

> This is the most performant and cost efficient way and we highly recommend setting hashKey (partition key) and rangeKey (sortKey) that you can query for.

I don't think this is always true. I can for sure think of legit instances where you don't need to query for a range/sort key. Of course you always need the hash/partition key. But I can totally think of legit instances where just querying the hash/partition key will meet a use case and is the most performant & cost effective solution.

> Also consider using secondary indexes instead.

Something doesn't feel right about this sentence. It feels overly broad. I'm not quite sure how to fix it tho. I don't think it really adds anything to this documentation, and I feel like raises more questions than it answers.

Not quite sure what to suggest changing it to tho. If you have any thoughts I'd love to hear them.

docs/api/Query.md
```diff
@@ -112,7 +114,9 @@ Cat.query("name").eq("Will").using("name-index"); // Run the query on the `name-
 
 ## query.all([delay[, max]])
 
-Unlike most other query functions that directly change the DynamoDB query request, this function is purely internal and unique to Dynamoose. This function sends continuous query requests upon receiving the response so long as the `lastKey` property exists on the response. This can be useful if you wish to get all the items from the table and don't want to worry about checking the `lastKey` property and sending a new query request yourself.
+If a query result is more than 1 MB (before filtering!), DynamoDB paginates the results so you would have to send multiple requests. Please see the [AWS DynamoDB documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html) for further informations.
```
fishcharlie, line 117:
```suggestion
Due to the fact that [DynamoDB paginates query results that are more than 1 MB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html) (before filtering!), Dynamoose provides the `query.all` method to automatically make subsequent requests to DynamoDB.
```

fishcharlie: Also not sure I'm a fan of having AWS links in the documentation. It's one more thing to maintain in case of 404 errors or something.

It should be assumed that if you are gonna use Dynamoose that you need to research DynamoDB. I don't think Dynamoose needs to be responsible for doing that.

I'm totally willing to reconsider that in the future maybe. But this is still a small project in terms of number of maintainers. Adding a link to AWS docs and trying to education people about DynamoDB I don't think should be part of Dynamoose documentation at this point when I don't have enough support in maintaining the project.

Also, the 1 MB limit, that could change at any point. Which again, is another point of having to maintain and update things. It just increases the chances of things getting out of date extremely quickly.

So far I've tried to give good advice to anyone asking for it. But going in depth to DynamoDB in the docs is a whole different game.

docs/api/Scan.md
```diff
@@ -123,7 +125,9 @@ Cat.scan().using("name-index"); // Run the scan on the `name-index` index
 
 ## scan.all([delay[, max]])
 
-Unlike most other scan functions that directly change the DynamoDB scan request, this function is purely internal and unique to Dynamoose. This function sends continuous scan requests upon receiving the response so long as the `lastKey` property exists on the response. This can be useful if you wish to get all the items from the table and don't want to worry about checking the `lastKey` property and sending a new scan request yourself.
+If a scan result is more than 1 MB (before filtering!), DynamoDB paginates the results so you would have to send multiple requests. Please see the [AWS DynamoDB documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html) for further informations.
```
fishcharlie, line 128: Pretty similar comments as above.

docs/api/Scan.md
```diff
@@ -1,6 +1,8 @@
 # Scan
 
-Dynamoose provides the ability to scan a model by using the `Model.scan` function. This function acts as a builder to construct your scan with the appropriate settings before executing it (`scan.exec`).
+Dynamoose provides the ability to scan a table (model) by using the `Model.scan` function. This function acts as a builder to construct your scan with the appropriate settings before executing it (`scan.exec`).
+
+**Please note:** The Scan operation operates on your entire table. For tables of real size, this can quickly use up all of your Read Capacity. If you're using it in your application's critical path, it will be very slow in returning a response to your users. The best option is not never use `scan()` unless you know what you are doing!
```
fishcharlie, line 5: > **Please note:**

Not a fan of the bold for same reasons as I discussed above.

> The Scan operation operates on your entire table.

This is not always true. It is possible to run a Scan on a secondary index, which has the potential to have a different number of items than the table itself.

I understand what you mean here, but I think there is a layer of confusion this could create.

> If you're using it in your application's critical path, it will be very slow in returning a response to your users.

I don't think this is always true either. Unless you are referring to large tables from above. But a table with let's say 3 items, that you always run a scan on, that very likely won't be very slow in returning a response to your users.

I do not believe Dynamoose should be giving database design or modeling advice. Every single use case is different. I believe we should present the information, but not make recommendations here. That opens a whole other layer of support if there is a use case for it, and I can totally imagine the emails and questions about `You said this was a bad idea, but I have this legit use case. Why is it a bad idea?`.

So I'm really against making assumptions about use cases or applications for Dynamoose.

> The best option is not never use `scan()` unless you know what you are doing!

Same reason here. Saying something along the lines of `The Scan operation operates on your entire table.` I believe makes this clear. I'm not a fan of being opinionated here. There are many very legit use cases for Scan. It's important to know what you are doing for sure. But isn't that true for many applications? I think the important part is that it has no partitions or filtering before looking through the items. It scans everything. I think that gets the point across.

I think we should leave the education about what Scan does behind the scenes and discouraging it to AWS. The goal of this documentation is to explain Dynamoose, not DynamoDB.

I think the best example of this, is you don't see [Express.js](https://expressjs.com) explaining how the web works. I mean even in their [`res.status`](https://expressjs.com/en/4x/api.html#res.status) documentation they don't even explain that 2xx means success, 4xx means client error, or 5xx means server error. They give the information about how Express works and leaves the rest to you to research on your own.

docs/api/Condition.md
```diff
@@ -24,7 +24,7 @@ new dynamoose.Condition().where("id").eq(1).and().where("name").eq("Bob");
 new dynamoose.Condition().where("id").eq(1).where("name").eq("Bob");
 ```
 
-## condition.and()
+## condition.or()
```
fishcharlie, line 27: 😆😬 My bad!!!

MickL:
I dont think you have to add the "do not merge" label. When you request changes the PR cant be merged anyway.

docs/api/Model.md
MickL: > Few suggestions. Still not sure I'm on board about the `Item` vs `Document` thing yet. I think that should maybe be a separate PR and we rename every instance of it.

Yeah lets do that in a separate step!

MickL:
I reverted item to document. Also added info for Model.functions with info that the primaryKey can be a string with the hashKey OR an object containing hashKey and rangeKey

docs/api/Query.md
MickL: I changed to:

> Please note: `Model.query()` combines both the `KeyConditionExpression` and the `FilterExpression` from DynamoDB. If you query for an attribute that you defined as your hashKey (partition key) or rangeKey (sort key) DynamoDB will use `KeyConditionExpression`. This could be the most performant and cost efficient way to query for. If querying for attributes that are not defined as your hashKey (partition key) or rangeKey (sortKey) DynamoDB might select more items at first (that can be found by hashKey and rangeKey) and then filter the result. This could have a bad impact on performance and costs.

What do you think?

docs/api/Query.md
MickL: I changed to:

> If a query result is more than the limit of your DynamoDB table (before filtering!), DynamoDB paginates the results so you would have to send multiple requests.

What do you think?

docs/api/Scan.md
MickL: Changed to the same as above, but with "scan" instead of "query". Will keep them in sync and close this conversation.

docs/api/Scan.md
MickL: I copied this 1 to 1 from https://www.dynamodbguide.com/scans , written by a DynamoDB expert. IMO in a case where the user can do very bad things we should give an advice. I mean the scan function is there, everyone can use it, but he has been warned at least...

docs/api/Model.md
```diff
@@ -1,26 +1,21 @@
 # Model
 
-The Model object represents one DynamoDB table. It takes in both a name and a schema and has methods to retrieve, and save items in the database.
+The Model object represents a table in DynamoDB. It takes in both a name and a schema and has methods to retrieve, and save items in the database.
 
 ## new dynamoose.Model(name, schema[, config])
 
-This method is the basic entry point for creating a model in Dynamoose. When you call this method a new model is created, and it returns a Document initializer that you can use to create instances of the given model.
```
fishcharlie, line 7: Would like to hear more about the reasoning for removing `Document` from the language here. I think it's important to have a reference that it's a Document initializer. That way it's clear that functions like `document.save` work correctly. It doesn't just create an initializer, it creates a Document initializer.

docs/api/Model.md
```diff
@@ -1,26 +1,21 @@
...
+This method is the basic entry point for creating a model in Dynamoose. When you call this method a new model is created, and it returns an initializer that you can use to create instances of the given model.
 
-The `schema` parameter can either be an object OR a Schema instance. If you pass in an object for the `schema` parameter it will create a Schema instance for you automatically.
+The `schema` parameter can either be an object OR a Schema instance, see [Schema](Schema.md) for details. If you pass in an object for the `schema` parameter it will create a Schema instance for you automatically.
```
fishcharlie, line 9: No need for extra words here.

```suggestion
The `schema` parameter can either be an object OR a [Schema](Schema.md) instance. If you pass in an object for the `schema` parameter it will create a Schema instance for you automatically.
```

docs/api/Model.md
```diff
@@ -1,26 +1,21 @@
...
 
 ```js
-const dynamoose = require("dynamoose");
```
fishcharlie, line 12: Why was this removed? Model is a common starting point for Dynamoose. I think it's important to understand how this is happening.

docs/api/Model.md
```diff
@@ -1,26 +1,21 @@
...
-
 const Cat = new dynamoose.Model("Cat", {"name": String});
-```
-
-```js
-const dynamoose = require("dynamoose");
-
+const Cat = new dynamoose.Model("Cat", {"name": String}, { create: false });
```
fishcharlie, line 13: Any reason for adding this? Right below we go over the `config` parameter. Why introduce it here? And why just `create`?

```suggestion
const Cat = new dynamoose.Model("Cat", {"name": String}, {"create": false});
```

docs/api/Model.md
```diff
@@ -1,26 +1,21 @@
...
 const Cat = new dynamoose.Model("Cat", new dynamoose.Schema({"name": String}));
+const Cat = new dynamoose.Model("Cat", new dynamoose.Schema({"name": String}), { create: false });
```
fishcharlie, line 15: Same question as above.

fishcharlie:
```suggestion
const Cat = new dynamoose.Model("Cat", new dynamoose.Schema({"name": String}), {"create": false});
```

docs/api/Model.md
```diff
@@ -36,9 +31,9 @@ The config parameter is an object used to customize settings for the model.
...
 | update | If Dynamoose should update the capacity of the existing table to match the model throughput. | Boolean | false |
-| expires | The setting to describe the time to live for documents created. If you pass in a number it will be used for the `expires.ttl` setting, with default values for everything else. If this is `null`, no time to live will be active on the model. | Number \| Object | null |
-| expires.ttl | The default amount of time the document should stay alive from creation time in milliseconds. | Number | null |
-| expires.attribute | The attribute name for where the document time to live attribute. | String | `ttl` |
+| expires | The setting to describe the time to live for items created. If you pass in a number it will be used for the `expires.ttl` setting, with default values for everything else. If this is `null`, no time to live will be active on the model. | Number \| Object | null |
+| expires.ttl | The default amount of time the item should stay alive from creation time in milliseconds. | Number | null |
+| expires.attribute | The attribute name for where the item time to live attribute. | String | `ttl` |
```
fishcharlie, line 36:
```suggestion
| expires | The setting to describe the time to live for documents created. If you pass in a number it will be used for the `expires.ttl` setting, with default values for everything else. If this is `null`, no time to live will be active on the model. | Number \| Object | null |
| expires.ttl | The default amount of time the document should stay alive from creation time in milliseconds. | Number | null |
| expires.attribute | The attribute name for where the document time to live attribute. | String | `ttl` |
```

docs/api/Model.md
```diff
@@ -124,9 +119,11 @@ async function printTableRequest() {
 }
 ```
 
-## Model.get(hashKey[, settings][, callback])
+## Model.get(primaryKey[, settings][, callback])
```
fishcharlie, line 122:
```suggestion
## Model.get(key[, settings][, callback])
```

docs/api/Model.md
```diff
@@ -124,9 +119,11 @@ async function printTableRequest() {
...
+
+You can use Model.get to retrieve a item from DynamoDB. This method uses the `getItem` DynamoDB API call to retrieve the object.
 
-You can use Model.get to retrieve a document from DynamoDB. This method uses the `getItem` DynamoDB API call to retrieve the object.
+`keys` can be a string representing the hashKey/partition key or an object containing the hashKey/partition key AND rangeKey/sort key.
```
fishcharlie, line 126:
```suggestion
The `key` parameter can be a string representing the hashKey or an object containing the hashKey and rangeKey.
```

docs/api/Model.md
```diff
@@ -173,7 +170,7 @@ User.get(1, {"return": "request"}, (error, request) => {
 });
 ```
 
-In the event you have a rangeKey for your model, you can pass in an object for the `hashKey` parameter.
+In the event you have a rangeKey/sort key for your model, you can pass in an object for the `key` parameter which includes the range key/sort key & the hash key/partition key.
```
fishcharlie, line 173:
```suggestion
In the event you have a `rangeKey` for your model, you can pass in an object for the `key` parameter which includes the `hashKey` & `rangeKey`.
```

docs/api/Model.md
fishcharlie: Actually. This should be deleted. It's covered below. It's repetitive.

```suggestion
```

docs/api/Model.md
```diff
@@ -124,9 +119,11 @@ async function printTableRequest() {
 }
 ```
 
-## Model.get(hashKey[, settings][, callback])
+## Model.get(primaryKey[, settings][, callback])
+
+You can use Model.get to retrieve a item from DynamoDB. This method uses the `getItem` DynamoDB API call to retrieve the object.
 
-You can use Model.get to retrieve a document from DynamoDB. This method uses the `getItem` DynamoDB API call to retrieve the object.
+`keys` can be a string representing the hashKey/partition key or an object containing the hashKey/partition key AND rangeKey/sort key.
```
fishcharlie, line 127:
```suggestion
```

docs/api/Model.md
```diff
@@ -217,9 +214,11 @@ User.get({"id": 1}, (error, myUser) => {
 });
 ```
 
-## Model.batchGet(keys[, settings][, callback])
+## Model.batchGet(primaryKeys[, settings][, callback])
```
fishcharlie, line 217:
```suggestion
## Model.batchGet(keys[, settings][, callback])
```

docs/api/Model.md
```diff
@@ -124,9 +119,11 @@ async function printTableRequest() {
 }
 ```
 
-## Model.get(hashKey[, settings][, callback])
+## Model.get(primaryKey[, settings][, callback])
+
+You can use Model.get to retrieve a item from DynamoDB. This method uses the `getItem` DynamoDB API call to retrieve the object.
```
fishcharlie, line 124:
```suggestion
You can use Model.get to retrieve a document from DynamoDB. This method uses the `getItem` DynamoDB API call to retrieve the object.
```

docs/api/Model.md
```diff
@@ -217,9 +214,11 @@ User.get({"id": 1}, (error, myUser) => {
 });
 ```
 
-## Model.batchGet(keys[, settings][, callback])
+## Model.batchGet(primaryKeys[, settings][, callback])
 
-You can use Model.batchGet to retrieve multiple documents from DynamoDB. This method uses the `batchGetItem` DynamoDB API call to retrieve the object.
+You can use Model.batchGet to retrieve multiple items from DynamoDB. This method uses the `batchGetItem` DynamoDB API call to retrieve the object.
```
fishcharlie, line 219:
```suggestion
You can use Model.batchGet to retrieve multiple documents from DynamoDB. This method uses the `batchGetItem` DynamoDB API call to retrieve the object.
```

docs/api/Model.md
```diff
@@ -278,7 +277,7 @@ User.batchGet([1, 2], {"return": "request"}, (error, request) => {
 });
 ```
 
-In the event you have a rangeKey for your model, you can pass in an object for the `hashKey` parameter.
+In the event you have a range key/sort key for your model, you can pass in an object for the `key` parameter which includes the range key/sort key & the hash key/partition key.
```
fishcharlie, line 280:
```suggestion
In the event you have a `rangeKey` for your model, you can pass in an object for the `key` parameter which includes the `hashKey` & `rangeKey`.
```

docs/api/Model.md
```diff
@@ -324,7 +323,7 @@ User.batchGet([{"id": 1}, {"id": 2}], (error, myUsers) => {
 
 ## Model.create(document, [settings], [callback])
 
-This function lets you create a new document for a given model. This function is almost identical to creating a new document and calling `document.save`, with one key difference, this function will default to setting `overwrite` to false.
+This function lets you create a new item for a given model. This function is almost identical to creating a new item and calling `document.save`, with one key difference, this function will default to setting `overwrite` to false.
```
fishcharlie, line 326:
```suggestion
This function lets you create a new document for a given model. This function is almost identical to creating a new document and calling `document.save`, with one key difference, this function will default to setting `overwrite` to false.
```

docs/api/Model.md
```diff
@@ -351,7 +350,7 @@ User.create({"id": 1, "name": "Tim"}, (error, user) => {  // If a user with `id=
 
 ## Model.batchPut(documents, [settings], [callback])
 
-This saves documents to DynamoDB. This method uses the `batchWriteItem` DynamoDB API call to store your objects in the given table associated with the model. This method is overwriting, and will overwrite the data you currently have in place for the existing key for your table.
+This saves items to DynamoDB. This method uses the `batchWriteItem` DynamoDB API call to store your objects in the given table associated with the model. This method is overwriting, and will overwrite the data you currently have in place for the existing key for your table.
```
fishcharlie, line 353:
```suggestion
This saves documents to DynamoDB. This method uses the `batchWriteItem` DynamoDB API call to store your objects in the given table associated with the model. This method is overwriting, and will overwrite the data you currently have in place for the existing key for your table.
```

docs/api/Document.md
```diff
@@ -1,6 +1,9 @@
 # Document
 
-A document represents an item for a given model in DynamoDB. This item can created locally (meaning it's not yet saved in DynamoDB), or created from an item already stored in DynamoDB (ex. `Model.get`).
+A document (also called item in DynamoDB) represents an item for a given model in DynamoDB. This item can created locally (meaning it's not yet saved in DynamoDB), or created from an item already stored in DynamoDB (ex. `Model.get`).
```
fishcharlie, line 3: > (also called item in DynamoDB) represents an item

Kinda repetitive, don't you think? Read 3 more words and you know what it represents or means. Instead we add 5 more words? Just adding complexity here.

```suggestion
A document represents an item for a given model in DynamoDB. This item can created locally (meaning it's not yet saved in DynamoDB), or created from an item already stored in DynamoDB (ex. `Model.get`).
```

docs/api/Model.md
```diff
@@ -217,9 +214,11 @@ User.get({"id": 1}, (error, myUser) => {
 });
 ```
 
-## Model.batchGet(keys[, settings][, callback])
+## Model.batchGet(primaryKeys[, settings][, callback])
 
-You can use Model.batchGet to retrieve multiple documents from DynamoDB. This method uses the `batchGetItem` DynamoDB API call to retrieve the object.
+You can use Model.batchGet to retrieve multiple items from DynamoDB. This method uses the `batchGetItem` DynamoDB API call to retrieve the object.
+
+`primaryKeys` can be an array of strings representing the hashKey/partition key and/or an array of objects containing the hashKey/partition key AND rangeKey/sort key.
```
fishcharlie, line 221:
```suggestion
```

docs/api/Model.md
```diff
@@ -832,7 +837,7 @@ User.scanAll((err, models) => {});
 
 ## Model.methods.document.set(name, function)
 
-This function allows you to add a method to the model documents that you can call later. When Dynamoose calls your `function` parameter, `this` will be set to the underlying document. If an existing method exists with the given name, it will be overwritten, except if you are trying to replace an internal method, then this function will fail silently.
+This function allows you to add a method to the model items that you can call later. When Dynamoose calls your `function` parameter, `this` will be set to the underlying item. If an existing method exists with the given name, it will be overwritten, except if you are trying to replace an internal method, then this function will fail silently.
```
fishcharlie, line 840:
```suggestion
This function allows you to add a method to the model documents that you can call later. When Dynamoose calls your `function` parameter, `this` will be set to the underlying document. If an existing method exists with the given name, it will be overwritten, except if you are trying to replace an internal method, then this function will fail silently.
```

docs/api/Model.md
```diff
@@ -896,7 +901,7 @@ console.log("Set name");
 
 ## Model.methods.document.delete(name)
 
-This allows you to delete an existing method from the document. If no existing method is assigned for that name, the function will do nothing and no error will be thrown.
+This allows you to delete an existing method from the item. If no existing method is assigned for that name, the function will do nothing and no error will be thrown.
```
fishcharlie, line 904:
```suggestion
This allows you to delete an existing method from the document. If no existing method is assigned for that name, the function will do nothing and no error will be thrown.
```

docs/api/Schema.md
```diff
@@ -356,7 +356,7 @@ If you set `index` to `true`, it will create an index with all of the default se
 
 ### hashKey: boolean
 
-You can set this to true to overwrite what the `hashKey` for the Model will be. By default the `hashKey` will be the first key in the Schema object.
+You can set this to true to overwrite what the `hashKey` (partition key) for the Model will be. By default the `hashKey` will be the first key in the Schema object.
```
fishcharlie, line 359: This is where it should be made more clear about partition key. Same goes for `rangeKey` below.

```suggestion
You can set this to true to overwrite what the `hashKey` for the Model will be. By default the `hashKey` will be the first key in the Schema object. `hashKey` is commonly called a partition key in the AWS documentation.
```

docs/api/Model.md
MickL: I think it was when I started naming document to item. This can be reverted.

docs/api/Model.md
MickL: It is already part of General.md and i didnt see the require anywhere else. So i thought it would consistent to have it nowhere. For me schema would be the entry point.

docs/api/Model.md
MickL: I think it was just to show that a Model accepts options. Can be removed if you think it is not necessary. I know the function definition is in the title, but sometimes one overreads it in the title and then sees the example and thinks it has no config options.

MickL:
I updated the example how to create indexes to showcase the creation of a global index and a secondary index.

docs/api/Model.md
MickL: Reverted

MickL:
- I updated the example how to create indexes to showcase the creation of a global index and a secondary index.
- I committed all your suggestions and answered all comments (see unresolved conversations).
- I removed all repetitive notes of partition key / sort key.
- I reverted all changed from document to item

Hope everything is okay now.

docs/api/Model.md
MickL: You used `create: false` in a lot of examples so I did that, too for consistency.

MickL:
I double checked everything. Should be good now. @fishcharlie feel free to make changes directly in the PR if you want.

docs/api/Document.md
```diff
@@ -2,6 +2,9 @@
 
 A document represents an item for a given model in DynamoDB. This item can created locally (meaning it's not yet saved in DynamoDB), or created from an item already stored in DynamoDB (ex. `Model.get`).
 
+_Info: An item is similar to a row in a relational database or a document in MongoDB._
+
+
```
fishcharlie, line 7:
```suggestion
```

docs/api/Model.md
```diff
@@ -173,7 +168,7 @@ User.get(1, {"return": "request"}, (error, request) => {
 });
 ```
 
-In the event you have a rangeKey for your model, you can pass in an object for the `hashKey` parameter.
+In the event you have a `rangeKey` for your model, you can pass in an object for the `key` parameter which includes the `hashKey` & `rangeKey`.
```
fishcharlie, line 171: Think this was my bad. Want it to be consistent throughout.

```suggestion
In the event you have a rangeKey for your model, you can pass in an object for the `key` parameter which includes the hashKey & rangeKey.
```

docs/api/Model.md
```diff
@@ -399,9 +394,11 @@ await User.batchPut([
 });
 ```
 
-## Model.update(keyObj[, updateObj[, settings]],[ callback])
+## Model.update(key[, updateObj[, settings]],[ callback])
+
+This function lets you update an existing item in the database. You can either pass in one object combining both the hashKey (partition key) you wish to update along with the update object, or keep them separate by passing in two objects.
 
-This function lets you update an existing document in the database. You can either pass in one object combining both the hashKey you wish to update along with the update object, or keep them separate by passing in two objects.
+`key` can be a string representing the hashKey or an object containing the hashKey AND rangeKey.
```
fishcharlie, line 401:
```suggestion
`key` can be a string representing the hashKey or an object containing the hashKey & rangeKey.
```

docs/api/Model.md
```diff
@@ -481,9 +478,11 @@ await User.update({"id": 1}, {"name": "Bob", "$ADD": {"age": 1}});
 
 The `validate` Schema attribute property will only be run on `$SET` values. This is due to the fact that Dynamoose is unaware of what the existing value is in the database for `$ADD` properties.
 
-## Model.delete(hashKey[, settings][, callback])
+## Model.delete(key[, settings][, callback])
 
-You can use Model.delete to delete a document from DynamoDB. This method uses the `deleteItem` DynamoDB API call to delete the object.
+You can use Model.delete to delete a item from DynamoDB. This method uses the `deleteItem` DynamoDB API call to delete the object.
+
+`key` can be a string representing the hashKey or an object containing the hashKey AND rangeKey.
```
fishcharlie, line 485:
```suggestion
`key` can be a string representing the hashKey or an object containing the hashKey & rangeKey.
```

docs/api/Model.md
```diff
@@ -530,7 +529,7 @@ User.delete(1, {"return": "request"}, (error, request) => {
 });
 ```
 
-In the event you have a rangeKey for your model, you can pass in an object for the `hashKey` parameter.
+In the event you have a rangeKey for your model, you can pass in an object for the `key` parameter which includes the rangeKey & the hashKey.
```
fishcharlie, line 532:
```suggestion
In the event you have a rangeKey for your model, you can pass in an object for the `key` parameter which includes the rangeKey & hashKey.
```

docs/api/Model.md
```diff
@@ -576,7 +575,9 @@ User.delete({"id": 1}, (error) => {
 
 ## Model.batchDelete(keys[, settings][, callback])
 
-You can use Model.batchDelete to delete documents from DynamoDB. This method uses the `batchWriteItem` DynamoDB API call to delete the objects.
+You can use Model.batchDelete to delete items from DynamoDB. This method uses the `batchWriteItem` DynamoDB API call to delete the objects.
+
+`keys` can be an array of strings representing the hashKey and/or an array of objects containing the hashKey AND rangeKey.
```
fishcharlie, line 580:
```suggestion
`keys` can be an array of strings representing the hashKey and/or an array of objects containing the hashKey & rangeKey.
```

docs/api/Model.md
```diff
@@ -644,7 +645,7 @@ User.batchDelete([1, 2], {"return": "request"}, (error, request) => {
 });
 ```
 
-In the event you have a rangeKey for your model, you can pass in an object for the `hashKey` parameter.
+In the event you have a rangeKey for your model, you can pass in an object for the `key` parameter which includes the rangeKey & the hashKey.
```
fishcharlie, line 648:
```suggestion
In the event you have a rangeKey for your model, you can pass in an object for the `key` parameter which includes the rangeKey & hashKey.
```

docs/api/Query.md
```diff
@@ -4,15 +4,17 @@ Dynamoose provides the ability to query a model by using the `Model.query` funct
 
 ## Model.query([filter])
 
-This is the basic entry point to construct a query request. The filter property is optional and can either be an object or a string representing the key you which to first filter on. In the event you don't pass in any parameters and don't call any other methods on the query object it will query with no filters or options.
+This is the basic entry point to construct a query request. It requires to set at least the hashKey of the item(s). The filter property is optional and can either be an object or a string representing the key you which to first filter on. In the event you don't pass in any parameters and don't call any other methods on the query object it will query with no filters or options.
```
fishcharlie, line 7:
```suggestion
This is the basic entry point to construct a query request. When running a query you must specify at least the hashKey of the document(s). This can either be the hashKey of the table, or the hashKey of an index. The filter property is optional and can either be an object or a string representing the key you which to first filter on. In the event you don't pass in any parameters and don't call any other methods on the query object it will query with no filters or options.
```

docs/api/Query.md
```diff
@@ -4,15 +4,17 @@ Dynamoose provides the ability to query a model by using the `Model.query` funct
...
-Cat.query({"breed": {"contains": "Terrier"}}).exec() // will query all items and filter all items where the key `breed` contains `Terrier`
+Cat.query("breed").contains("Terrier").exec() // will query all items where the hashKey `breed` contains `Terrier`
+Cat.query({"breed": {"contains": "Terrier"}}).exec() // will query all items where the hashKey `breed` contains `Terrier`
 ```
 
 If you pass an object into `Model.query` the object for each key should contain the comparison type. For example, in the last example above, `contains` was our comparison type. This comparison type must match one of the comparison type functions listed on this page.
 
+Please note: `Model.query()` combines both the `KeyConditionExpression` and the `FilterExpression` from DynamoDB. If you query for an attribute that you defined as your hashKey or rangeKey DynamoDB will use `KeyConditionExpression`. This could be the most performant and cost efficient way to query for. If querying for attributes that are not defined as your hashKey or rangeKey DynamoDB might select more items at first and then filter the result which could have a bad impact on performance and costs.
```
fishcharlie, line 16:
```suggestion
`Model.query()` combines both the `KeyConditionExpression` and the `FilterExpression` from DynamoDB. If you query for an attribute that you defined as your hashKey or rangeKey DynamoDB will use `KeyConditionExpression`. This could be the most performant and cost efficient way to query for. If querying for attributes that are not defined as your hashKey or rangeKey DynamoDB might select more items at first and then filter the result which could have a bad impact on performance and costs.    
```

docs/api/Schema.md
```diff
@@ -332,24 +332,33 @@ If you set `index` to `true`, it will create an index with all of the default se
...
+    },
 	"email": {
 		"type": String,
 		"index": {
-			"name": "emailIndex",
-			"global": true
-		} // creates a global index with the name `emailIndex`
+          "name": "emailIndex",
+          "global": true
+        } // creates a global secondary index with the name `emailIndex` and hashKey `email`
```
fishcharlie, line 344: Same as above.

docs/api/Schema.md
```diff
@@ -332,24 +332,33 @@ If you set `index` to `true`, it will create an index with all of the default se
 
 ```js
 {
+    "id" : {
+      "hashKey": true,
+      "type": String,
+    },
```
fishcharlie, line 338: Probably have to format this to use tabs to ensure consistency.

docs/api/Schema.md
```diff
@@ -332,24 +332,33 @@ If you set `index` to `true`, it will create an index with all of the default se
...
+        "throughput": {"read": 5, "write": 10}
+      } // creates a local secondary index with the name `emailIndex`, hashKey `id`, rangeKey `email`
+    },
 	"email": {
-		"type": String,
-		"index": {
-			"name": "emailIndex",
-			"throughput": {"read": 5, "write": 10}
-		} // creates a global index with the name `emailIndex`
+		"type": String
```
fishcharlie, line 361: Same as above.

docs/api/Scan.md
```diff
@@ -1,6 +1,8 @@
 # Scan
 
-Dynamoose provides the ability to scan a model by using the `Model.scan` function. This function acts as a builder to construct your scan with the appropriate settings before executing it (`scan.exec`).
+Dynamoose provides the ability to scan a table (model) by using the `Model.scan` function. This function acts as a builder to construct your scan with the appropriate settings before executing it (`scan.exec`).
```
fishcharlie, line 3:
```suggestion
Dynamoose provides the ability to scan a model by using the `Model.scan` function. This function acts as a builder to construct your scan with the appropriate settings before executing it (`scan.exec`).
```

docs/api/Scan.md
```diff
@@ -1,6 +1,8 @@
...
+
+Please note: The Scan operation operates on your entire table. For tables of real size, this can quickly use up all of your Read Capacity. If you're using it in your application's critical path, it will be very slow in returning a response to your users. The best option is not never use `scan()` unless you know what you are doing!
```
fishcharlie, line 5: How do you feel about something like this? I feel like it's phrased in a way that is more factual, and less about specific application instances or consumer skill level.

```suggestion
Scan operations run on every document in your table or index. This means that all filtering is done after the documents are read, and therefore is not the most performant method. Due to that, if possible, it is highly encouraged to use [`Model.query`](Query.md) as opposed to `Model.scan`.
```

docs/api/Query.md
```diff
@@ -112,7 +114,7 @@ Cat.query("name").eq("Will").using("name-index"); // Run the query on the `name-
 
 ## query.all([delay[, max]])
 
-Unlike most other query functions that directly change the DynamoDB query request, this function is purely internal and unique to Dynamoose. This function sends continuous query requests upon receiving the response so long as the `lastKey` property exists on the response. This can be useful if you wish to get all the items from the table and don't want to worry about checking the `lastKey` property and sending a new query request yourself.
+If a query result is more than the limit of your DynamoDB table (before filtering!), DynamoDB paginates the results so you would have to send multiple requests. This function sends continuous query requests until all items have been received (as long as the `lastKey` property exists on the response). This can be useful if you wish to get all the items from the table and don't want to worry about checking the `lastKey` property and sending a new query request yourself.
```
fishcharlie, line 117: Still gotta figure out how to make this better. `the limit of your DynamoDB table` doesn't really make a lot of sense. Reading that makes me think there is a setting or option to adjust the limit of the table. Not quite sure how to make it better, will think about it for the next review.

Same likely goes for Scan.all.

```suggestion
If a query result is more than the limit of your DynamoDB table, DynamoDB paginates the results so you would have to send multiple requests. This function sends continuous query requests until all items have been received (as long as the `lastKey` property exists on the response). This can be useful if you wish to get all the items from the table and don't want to worry about checking the `lastKey` property and sending a new query request yourself.
```

fishcharlie:
@MickL I'm good with either one. Do you prefer me to commit directly to your branch? Especially since it's your `master` branch I do get hesitant about doing that on someone else's fork without discussing it first.

Thanks for all your work on this! It's really appreciated.

MickL:
@fishcharlie I committed all your suggestions. You can make changes directly on my branch no problem :)

MickL:
Updated .count() because for me it was not clear if it actually uses `Select: "COUNT"` and saves band width(which it does!) or just makes a normal query and then returns the count. Had to figure that out by using console.log query-params.

I basically copied what the [documentation says](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html) but repalaced the term "item" with "document".

MickL:
TODO @fishcharlie : Add information about `null` values somewhere

docs/api/Model.md
fishcharlie: Adding this back to only the Model and Schema initialization sections.

PR #800, opened 2020-04-12
Serialization

### LINKED ISSUE #797: [FEATURE] Proposal to add methods to serialize the output of documents prior to delivering to end user client
## Goal
Add serialization methods to output Items/Documents in one or more formats to aid in adding, removing, or modifying information prior to output to an end user or other internal methods.

Based on this package for serializing Sequelize Documents: https://github.com/hauru/sequelize-to-json 

#### Model.addSerializer(name, options)
Configuring a new serializer that can be called by name
* name - The name of this serializer
* options - A schema defining how to manipulate a given document  

#### Model.serialize(documents: Array, nameOrOptions: String/Object/Array)
Runs an array of documents through a given serializer by name or provided schema
* documents - And array of documents belonging to this model
* nameOrOptions - One of either a name of an existing configured serializer, an array containing the names
 of fields that should be returned (equivalent to options.include), or an options object detailing more complex operations (see below)

#### Document.serialize(nameOrOptions: String/Object/Array)
Runs the given document through the named serializer or provided schema
* nameOrOptions - One of either a name of an existing configured serializer, an array containing the names
 of fields that should be returned (equivalent to options.include), or an options object detailing more complex operations (see below)

### Options

| Field | Type | Description |
| ----- | ---- | ----------- |
| include | array | Field names which should be included in the serialized output of the document |
| exclude | array | Field names which should **not** be included in the serialized output of the document |
| postSerialize | function | A function which receives the fully serialized object and the original document to make last minute changes to the output |

### Example

```javascript
const document = myModel.create({
  email: 'hello@example.com',
  username: 'bobdole',
  firstName: 'Bob',
  lastName: 'Dole',
  password: 'testing123',
  stripeCustomerId: 'cust_m1234567890',
  status: 'ACTIVE'
});

const options = {
  include: ['email', 'username', 'firstName', 'lastName'],
  exclude: ['password', 'stripeCustomerId', 'status'],
  postSerialize: (serialized, original) => {
    serialized.isActive = original.status === 'ACTIVE';
    return serialized;
  }  
}

myModel.addSerializer('mySerializer', options);

const serialized = document.serialize('mySerializer');

console.log(serialized);

/*
  Returns:

  {
    email: 'hello@example.com',
    username: 'bobdole',
    firstName: 'Bob',
    lastName: 'Dole',
    isActive: true
  }
*/

```

ablankenship10:
### Summary:

Implement Serialization Class. WIP, needs connection to Model + Document. Details in TODO comments

#### Schema
```
// Code here
```

#### Model
```
// Code here
```

#### General
```
// Code here
```

### GitHub linked issue:

Closes #797 

### Other information:
Details on how to implement into the Model and Document classes in TODO comments in the code.

### Type (select 1):
- [ ] Bug fix
- [X] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [X] No
- [ ] I'm not sure

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

### Other:
- [ ] I have searched through the GitHub pull requests to ensure this PR has not already been submitted
  - [ ] `npm test`
  - [ ] `npm run lint`
- [ ] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoosejs/dynamoose/blob/master/LICENSE)
- [ ] All of my commits and commit messages are detailed, explain what changes were made, and are easy to follow and understand


ablankenship10:
Should be in a decent place that is usable. A few items to discuss before moving any further:

1. Is error handling acceptable? Is everything being checked to an adequate degree before being allowed to proceed, or should there be more or less checks? Do we want the library to throw errors or fail in a different manor?

2. For documentation, do you want a file dedicated to serialization or do you just want the functions described in the respective Model + Document files?

Anything else you need from me to proceed?

fishcharlie:
@ablankenship10 

1. Will respond to this individually in review.
2. What do you think? I personally think same file since that's how `Model.methods` has it.

lib/Document.js
```diff
@@ -398,6 +403,7 @@ function DocumentCarrier(model) {
 			Document[key] = Document.Model[key].bind(Document.Model);
 		}
 	});
+	Document.serializer = model.serializer;
```
fishcharlie, line 406: We should probably put this under the `Document.prototype.serialize = function(...args) {` method above.

lib/Serializer.js
```diff
@@ -0,0 +1,118 @@
+class Serializer {
```
fishcharlie, line 1: How do we feel about creating a folder called `Serializer`. Then having this file be called `Serializer/index.js`. Then having a file called `Serializer/utils.js`. That way we can separate it out a bit more.

lib/Serializer.js
```diff
@@ -0,0 +1,118 @@
...
+			if (options.modify && typeof options.modify === "function") {
+				if (!options.include && !options.exclude) {
+					serialized = {...document};
+				}
+				serialized = options.modify(serialized, document);
+			}
+			return serialized;
+		} catch (error) {
+			// Failing quietly and defaulting to dumping the whole object may not be safest idea, lest we expose sensitive data.
+			return {};
```
fishcharlie, line 72: How do we feel about this? What errors are we expecting here? Would it be better to just throw the error?

test/Serializer.js
```diff
@@ -0,0 +1,251 @@
...
+const {Model} = dynamoose;
+const Serializer = require("../lib/Serializer");
+
+describe("Serializer", () => {
+	let User;
+	beforeEach(() => {
+		User = new Model("User",
+			{id: Number, name: String, email: String, phone: String, passwordHash: String, status: String},
+			{"create": false, "waitForActive": false}
+		);
```
fishcharlie, line 15: I know it's a lot on one line, but I'd prefer to maintain consistency throughout.

```suggestion
		User = new Model("User", {id: Number, name: String, email: String, phone: String, passwordHash: String, status: String}, {"create": false, "waitForActive": false});
```

lib/Serializer.js
```diff
@@ -0,0 +1,118 @@
+class Serializer {
+	constructor() {
+		this._serializers = {_default: _defaultSerializer};
+		this._defaultSerializer = "_default";
```
fishcharlie, line 4: I really want to move to private properties. Happen to know what version of Node.js supports that? This just feels kinda messy to me. Lots of other cases in the code base where we could do that as well.

test/Serializer.js
```diff
@@ -0,0 +1,251 @@
...
+
+	it("Should verify Model and Document have a serializer instance and expected methods bound", () => {
+		const doc = new User({"id": 1, name: "User"});
+		expect(User.serializer).to.be.an.instanceof(Serializer);
+		expect(doc.serialize).to.be.a("function");
+		expect(User.serializeMany).to.be.a("function");
+	});
+
+	it("Should add some serializers to the Models serializer instance", () => {
+		addSerializers();
```
fishcharlie, line 29: Seems like this happens on almost every test. How do we feel about adding this to a `beforeEach`?

test/Serializer.js
```diff
@@ -0,0 +1,251 @@
...
+	it("Should run the document through a serializer configured with an array (include)", () => {
+		addSerializers();
+		const docs = createDocuments();
+		const result = docs[0].serialize("contactInfoOnly");
+		expect(result).to.have.property("name");
+		expect(result).to.have.property("email");
+		expect(result).to.have.property("phone");
+		expect(result).to.not.have.property("id");
+		expect(result).to.not.have.property("passwordHash");
+		expect(result).to.not.have.property("status");
```
fishcharlie, line 66: Something like this maybe?

```suggestion
		expect(result).to.eql(utils.object.pick(docs[0], ["name", "email", "phone"]));
```

fishcharlie: Will have to do it throughout.

test/Serializer.js
```diff
@@ -0,0 +1,251 @@
...
+			id: 2,
+			name: "User Two",
+			email: "usertwo@example.com",
+			phone: "0123456789",
+			passwordHash: "KnbuZWU0RNPPXhgPNMxovLtUiuMN4I6i",
+			status: "active"
+		});
+
+		return docs;
+	};
```
fishcharlie, line 250: I think we should move these to the top. And ideally I think these should probably be `beforeEach` hooks. And we could have a higher scoped variable called `docs`. Again, just to maintain consistency throughout. Not sure if the way I wrote the other tests is the best way, but seems beneficial to maintain consistency.

test/Serializer.js
```diff
@@ -0,0 +1,251 @@
...
+		addSerializers();
+		const docs = createDocuments();
+		const result = docs[0].serialize("nonExistingSerializer");
+		expect(Object.keys(result)).to.be.an("array").length(0);
+	});
+
+	it("Should throw errors on invalid usage", () => {
+		expect(() => {
+			User.serializer.add({invalidUsage: "Should fail"}, ["name", "email"]);
+		}).to.throw("Field name is required and should be of type string");
```
fishcharlie, line 198: How do we feel about making these one line?

```suggestion
		expect(() => User.serializer.add({invalidUsage: "Should fail"}, ["name", "email"])).to.throw("Field name is required and should be of type string");
```

lib/Serializer.js
```diff
@@ -0,0 +1,118 @@
...
+const cleanAndValidateDocumentsArray = documentsArray => {
+	if (!documentsArray || !Array.isArray(documentsArray)) {
+		throw new Error("documentsArray must be an array of document objects");
+	}
+	return documentsArray.filter(doc => typeof doc.serialize === "function");
+};
+
+const _defaultSerializer = {
+	modify: (s, o) => {
+		return {...o};
```
fishcharlie, line 114: I think we should name these a bit better. Not sure if I'm correct on the first one.

```suggestion
	modify: (settings, object) => {
		return {...object};
```

lib/Serializer.js
```diff
@@ -0,0 +1,118 @@
...
+		return _serialized;
+	}, serialized);
+
+const excludeHandler = (document, excludeRules, serialized) =>
+	excludeRules.reduce((_serialized, key) => {
+		if (Object.prototype.hasOwnProperty.call(_serialized, key)) {
+			delete _serialized[key];
+		}
+		return _serialized;
+	}, serialized);
```
fishcharlie, line 91: Strange that this didn't throw an ESLint error. Might have to look into which lint error to turn on for that. Totally a strict style guide thing for me haha.

```suggestion
const excludeHandler = (document, excludeRules, serialized) => {
	excludeRules.reduce((_serialized, key) => {
		if (Object.prototype.hasOwnProperty.call(_serialized, key)) {
			delete _serialized[key];
		}
		return _serialized;
	}, serialized);
};
```

lib/Serializer.js
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+const includeHandler = (document, includeRules, serialized) =>
+	includeRules.reduce((_serialized, key) => {
+		if (Object.prototype.hasOwnProperty.call(document, key)) {
+			_serialized[key] = document[key];
+		}
+		return _serialized;
+	}, serialized);
```
fishcharlie, line 83: Same as below.

```suggestion
const includeHandler = (document, includeRules, serialized) => {
	includeRules.reduce((_serialized, key) => {
		if (Object.prototype.hasOwnProperty.call(document, key)) {
			_serialized[key] = document[key];
		}
		return _serialized;
	}, serialized);
};
```

lib/Document.js
ablankenship10: This isn't actually necessary, just the injection of the Document.serialize() function. Will fix this.

lib/Serializer.js
ablankenship10: The only case I can forsee is just sending completely wrong data types. Yea throwing an error is probably acceptable, though maybe a specific type of error so its more easily handleable

lib/Serializer.js
ablankenship10: 12.16.1 is the latest with private properties. Which is the current LTS and used by Lambda.

lib/Serializer.js
ablankenship10: modify(serialized, original) would be correct. Not opposed just taking up minimal space for a small default method.

lib/Serializer.js
ablankenship10: Does this actually return the result with the brackets? Would think it would just be undefined unless you add `return` also

ablankenship10:
> 2. What do you think? I personally think same file since that's how `Model.methods` has it.

Yea sounds good to me!

lib/Serializer.js
fishcharlie: @ablankenship10 Ahh. No you are correct. Gotta add `return` as well. My bad.

lib/Serializer.js
fishcharlie: @ablankenship10 I do agree, however I think it makes it easier for new contributors to submit PRs with clearer naming. Especially for non obvious methods like this. Generic naming for something like `Array.forEach` or something seems more acceptable. But one of my goals recently is to make the code base as approachable as possible (obviously I failed at that with the craziness going on with binding methods to Model/Document 🤣).

lib/Serializer.js
fishcharlie: @ablankenship10 Happen to know if AWS Lambda allows for creating a new Lambda function with a Node.js version below that?

lib/Serializer.js
fishcharlie: @ablankenship10 That sounds like a great solution to me. I think it'll confuse users more if it returns an empty object on error.

lib/Serializer.js
ablankenship10: Lambda currently supports version 10 and 12 only.

lib/Serializer.js
ablankenship10: Im not sure that's very necessary as its a pretty small class in itself. The injection of the methods on the other object is a bit weird perhaps but being only 3 lines or less is pretty manageable to hook into everything, and really its just for ease of access to the user instead of going to the Model.serializer object itself every time.

fishcharlie:
@ablankenship10 Are you using this branch in your projects/services/applications? Would love to merge `master` into it, but there are breaking changes between this branch and `master`.

ablankenship10:
Not yet you're good!

fishcharlie:
Just merged `master` into this branch (b4d4167). This commit has breaking changes.

lib/Serializer.js
fishcharlie: Fixed in 1faea335b85769c2e9e07d14459601c385a0f824

test/Serializer.js
fishcharlie: Fixed in e1edde6ca579de801db6063a5dd387cd374cd797

test/Serializer.js
fishcharlie: Fixed in 8ea90a9c1380a02ea8172a4300abfe9751f3743a

lib/Serializer.js
fishcharlie: Fixed in cabc47162f7fbef33f4bfed7b9999812573b7b3c

test/Serializer.js
fishcharlie: Fixed in 3ed4e6af6d0bf403bf997942918c7cf31140153f

test/Serializer.js
fishcharlie: Done in 6a718133db0e7c471a338e8b3074938c77b50342

lib/Serializer.js
fishcharlie: Fixed in 3dd621a04d5ec3c93e97a1ea6f90a538603c5801

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+export class Serializer {
+	#serializers: any;
+	#defaultSerializer: string;
+
+	constructor() {
+		this.#serializers = {_default: defaultSerializer};
+		this.#defaultSerializer = "_default";
+	}
+
+	add(name: string, options): void {
```
fishcharlie, line 59: We have to add a TypeScript interface file for options. I have to go back and reread the original proposal and dig into what is allowed to be passed into `options`.

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+
+	remove(name: string): void {
+		validateName(name);
+		if (Object.keys(this.#serializers).includes(name)) {
+			delete this.#serializers[name];
+		}
+		if (this.#defaultSerializer === name) {
+			this.#defaultSerializer = "_default";
```
fishcharlie, line 78: This should be refactored to use `setDefault`

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+	}
+};
+
+const cleanAndValidateDocumentsArray = (documentsArray: ModelType<Document>[]): ModelType<Document>[] => {
+	if (!documentsArray || !Array.isArray(documentsArray)) {
+		throw new Error("documentsArray must be an array of document objects");
+	}
+
+	return documentsArray.filter((doc) => typeof doc.serialize === "function");
+};
```
fishcharlie, line 48: Not sure what the purpose of this function is. Is it just to filter out any plain objects that are not Documents? @ablankenship10

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+
+const excludeHandler = (document: ModelType<Document>, excludeRules: string[], serialized: ObjectType): ObjectType => {
+	return excludeRules.reduce((serialized: ObjectType, key: string) => {
+		if (Object.keys(serialized).includes(key)) {
+			delete serialized[key];
+		}
+
+		return serialized;
+	}, serialized);
+};
```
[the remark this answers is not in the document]
fishcharlie: I think there also might be room here to use `utils.object` methods here. Seems like these are broader functions that shouldn't be limited to Serializer.

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+		validateName(name);
+		if (Object.keys(this.#serializers).includes(name)) {
+			delete this.#serializers[name];
+		}
+		if (this.#defaultSerializer === name) {
+			this.#defaultSerializer = "_default";
+		}
+	}
+
+	_serializeMany(documentsArray: ModelType<Document>[] = [], nameOrOptions): ObjectType[] {
```
fishcharlie, line 82: Wanna figure out if there is a way to use private methods for this and `_serialize`. And make it only accessible through Dynamoose. Since it won't be accessed outside of this library.

test/Serializer.js
```diff
@@ -0,0 +1,118 @@
+const chaiAsPromised = require("chai-as-promised");
+const chai = require("chai");
+chai.use(chaiAsPromised);
+const {expect} = chai;
+const dynamoose = require("../dist");
+const {model} = dynamoose;
+const {Serializer} = require("../dist/Serializer");
+const utils = require("../dist/utils");
+
+describe("Serializer", () => {
```
fishcharlie, line 10: This needs to be improved a lot. I think we need to move a lot of this into the `Model` and `Document` tests.

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+
+	_serialize(document: ModelType<Document>, nameOrOptions = this.#defaultSerializer): ObjectType {
+		const inputType = typeof nameOrOptions;
+		let options;
+
+		if (inputType === "string") {
+			options = this.#serializers[nameOrOptions];
+		} else if (Array.isArray(nameOrOptions) || inputType === "object") {
+			options = nameOrOptions;
+		}
```
fishcharlie, line 95: In what situation would neither of these be run @ablankenship10?

Could we just do the following?

```js
if (inputType === "string") {
    options = this.#serializers[nameOrOptions];
} else {
    options = nameOrOptions;
}
```

Otherwise, we need a test for the `else` case where neither of these get run. And I'm not sure what that would look like.

test/Serializer.js
```diff
@@ -0,0 +1,118 @@
...
+				passwordHash: "5dxFDpyKeEKiVUSp9I6dQ1DGd3CFH5Jk",
+				status: "not_active"
+			},
+			{
+				id: 2,
+				name: "User Two",
+				email: "usertwo@example.com",
+				phone: "0123456789",
+				passwordHash: "KnbuZWU0RNPPXhgPNMxovLtUiuMN4I6i",
+				status: "active"
```
fishcharlie, line 54: All of the properties here need quotes around them.

test/Serializer.js
```diff
@@ -0,0 +1,118 @@
...
+		].map((a) => new User(a));
+	});
+	afterEach(() => {
+		docs = [];
+	});
+
+	it("Should verify Model and Document have a serializer instance and expected methods bound", () => {
+		expect(User.serializer).to.be.an.instanceof(Serializer);
+		expect(User.serializeMany).to.be.a("function");
+		expect(new User({"id": 1, name: "User"}).serialize).to.be.a("function");
```
fishcharlie, line 65:
```suggestion
		expect(new User({"id": 1, "name": "User"}).serialize).to.be.a("function");
```

test/Serializer.js
```diff
@@ -0,0 +1,118 @@
...
+	});
+
+	it("Should throw error when trying to serialize using non existant serializer", () => {
+		expect(() => docs[0].serialize("nonExistingSerializer")).throw("Field options is required and should be an object or array");
+	});
+
+	it("Should throw errors on invalid usage", () => {
+		expect(() => User.serializer.add({invalidUsage: "Should fail"}, ["name", "email"])).to.throw("Field name is required and should be of type string");
+		expect(() => User.serializer.add("broken", "invalidOptionsUsage")).to.throw("Field options is required and should be an object or array");
+		expect(() => User.serializeMany({notAnArray: "ofDocuments"})).to.throw("documentsArray must be an array of document objects");
```
fishcharlie, line 116: These object keys all need quotes around them.

test/Serializer.js
```diff
@@ -0,0 +1,118 @@
...
+		User.serializer.add("redundant", {include: ["email", "phone", "doesntExist"], exclude: ["phone", "doesntExist"]});
+		User.serializer.add("isActiveNoStatus", {
+			exclude: ["status"],
+			modify: (serialized, original) => {
+				serialized.isActive = original.status === "active";
+				return serialized;
+			}
+		});
+		User.serializer.add("isActive", {
+			modify: (serialized, original) => {
```
fishcharlie, line 31: All of these property names need quotes around them.

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+		if (Array.isArray(options)) {
+			return includeHandler(document, options, {});
+		}
+
+		let serialized: ObjectType = {};
+		if (options.include) {
+			serialized = includeHandler(document, options.include, serialized);
+		}
+		if (options.exclude) {
+			if (!options.include) {
```
fishcharlie, line 107: In what case would you want to pass in both `exclude` & `include`? I think we should put a limit where it only works with one or the other.

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+				serialized = {...document};
+			}
+			serialized = excludeHandler(document, options.exclude, serialized);
+		}
+		if (options.modify && typeof options.modify === "function") {
+			if (!options.include && !options.exclude) {
+				serialized = {...document};
```
fishcharlie, line 114: I think we should remove this and set it as the default value of this variable

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+			return includeHandler(document, options, {});
+		}
+
+		let serialized: ObjectType = {};
+		if (options.include) {
+			serialized = includeHandler(document, options.include, serialized);
+		}
+		if (options.exclude) {
+			if (!options.include) {
+				serialized = {...document};
```
fishcharlie, line 108: I think we should remove this and set it as the default value of this variable

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
...
+
+const validateOptions = (options): void => {
+	if (!options || !(Array.isArray(options) || typeof options === "object")) {
+		throw new Error("Field options is required and should be an object or array");
+	}
+};
+
+const cleanAndValidateDocumentsArray = (documentsArray: ModelType<Document>[]): ModelType<Document>[] => {
+	if (!documentsArray || !Array.isArray(documentsArray)) {
+		throw new Error("documentsArray must be an array of document objects");
```
fishcharlie, line 44: All of the Error's in this file should be refactored to be CustomError's

lib/Serializer.ts
```diff
@@ -0,0 +1,120 @@
+import { ObjectType, ModelType } from "./General";
+import { Document } from "./Document";
+
+const defaultSerializer = {
+	"modify": (serialized: ObjectType, original: ObjectType): ObjectType => {
+		return {...original};
+	}
```
fishcharlie, line 7:
```suggestion
	"modify": (serialized: ObjectType, original: ObjectType): ObjectType => ({...original})
```

test/Serializer.js
fishcharlie: Also there are certain things not being tested. We need to increase code coverage back to 100%

fishcharlie: Maybe need to figure out how to test private properties in classes

lib/Serializer.ts
ablankenship10: So this is potentially future proofing for an additional feature (found in the library I based this work on). Basically they allowed you to use wildcard tokens that you could like include all foreign keys (maybe say `@keys` in the include or exclude arrays that does the appropriate action for all hash+range key values in the document). So if you included all keys but then also wanted to exclude a certain one you could do that. You're right that its somewhat pointless for the current iteration but I think not really an issue to leave as is.

fishcharlie: @ablankenship10 Something like having include be `friends.**` (to include all nested attributes in `friends`) and exclude be `friends.*.age` (to exclude that one specific property)?

lib/Serializer.ts
ablankenship10: Basically in the case you use doc.serialize() without providing a serializer name, SerializerOptions type (array or object), and your don't have a defaultSerializer set, options would remain null and thus throw an error at the validateOptions() line because its not configured properly for use. Another option could be to just return the document as  is instead of failing with an error.

lib/Serializer.ts
ablankenship10: I don't think the current code supports nested documents. I wouldn't necessarily handle it like this as you would start having to configure values for serializers of other Models and it gets messy that way. The way I've done it before with the other library is using the modify function, if the current document being serialized contains another "joined" document, to set its value to the serialized.otherDocument = otherDocument.serialize() and run its own serializers configured in its own model.

lib/Serializer.ts
ablankenship10: If im following your suggestion correctly, the reason the code is this way is if you didn't configure one of includes or excludes and only had a modify function, it would operate on the complete document  values. Otherwise continue to use the serialized values that have already ran.

lib/Serializer.ts
ablankenship10: Same as above, if includes was not set in your options, meaning to include everything and only exclude, you would need the default to be all document values.

lib/Serializer.ts
fishcharlie: Ahh. I think I see what you mean. I still think there is a way to refactor this to have it work.

lib/Serializer.ts
fishcharlie: It does support nested objects tho. So you could have an attribute that is `friends.age` or `friends.name` which would be a nested object within the document.

lib/Serializer.ts
fishcharlie: Fixed in e9af7f4636853211090c4238161ea3e430fd5710

lib/Serializer.ts
fishcharlie: Fixed in 76c139d6808cefc95a7d6b66ef1981a405936301

lib/Serializer.ts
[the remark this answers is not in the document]
fishcharlie: Fixed in cf87578a52137f808fe96694a7e4d5d557fedfb7 & f499b58fa8ecd385c956d581efcf222598609e8c

lib/Serializer.ts
fishcharlie: Done in bbc1db27f9fd9cb2f93327491f31d821a946a0a2

lib/Serializer.ts
fishcharlie: @ablankenship10 And in what case would you not have a defaultSerializer set? Just trying to figure out how to write a test for this.

test/Serializer.js
fishcharlie: Fixed in 3a933d0cc04ce72766c8fa43d6bc7a4f19c47fd9

test/Serializer.js
fishcharlie: Fixed in 3a933d0cc04ce72766c8fa43d6bc7a4f19c47fd9

test/Serializer.js
fishcharlie: Fixed in 3a933d0cc04ce72766c8fa43d6bc7a4f19c47fd9

test/Serializer.js
fishcharlie: Fixed in 3a933d0cc04ce72766c8fa43d6bc7a4f19c47fd9

lib/Serializer.ts
fishcharlie: Fixed in 810e4680653bed69202410b0629a9eaf9ba9351b

lib/Serializer.ts
ablankenship10: If you attempted to use document.serialize() before actually configuring one (actually the included default serializer (at the top of the file) SHOULD always at least be used. This is mainly for handling a case of user error though it may not be possible to hit. Originally the tests I wrote did have complete coverage though so I think I managed to hit it somewhere.

fishcharlie: @ablankenship10 Got it. Will look into this more after I work on the tests. If it still can't be hit I'll refactor to use else.

I think part of this is also that I changed the default serializer to be a private property. So I don't think it's even accessible for the user to modify anymore.

ablankenship10: Yea that one is just there so if you don't do anything at all but use serialize() it at least returns the full document. Once you add at least 1 serializer, that one should be used as default unless its removed. Ideally if you only have 1, or a main one, you never have to actually call it by name when calling serialize(). The only purpose of the named ones is in the case where you want something different than the default (i.e. returning a user their own User object with their own account settings configured, vs another users object where that information is not needed/private)

test/Serializer.js
fishcharlie: Fixed in 5102af5a74e7b98c671fdfd423ded589faea6a57

lib/Serializer.ts
fishcharlie: I can't figure out a way to do this quickly, so gonna leave it as is. It'll be a project for another time.

test/Model.js
```diff
@@ -3748,6 +3748,238 @@ describe("Model", () => {
...
+				});
+			});
+		});
+
+		describe("model.serialize", () => {
+			it("Should be a function", () => {
+				expect(new User().serialize).to.be.a("function");
+			});
+
+			serializeTests.forEach((test) => {
```
fishcharlie, line 3949: I really wish this could be part of `test/Document.js`, but I can't figure out a clean way to make it DRY while still doing that.

Any ideas anyone?

fishcharlie:
After some discussions in Slack I think there were a few concerns about this pull request. Mainly surrounding if this pull request should even be merged or implemented. I have attempted to summarize the concerns here along with my replies.

1. Why not use [`_.pick`](https://lodash.com/docs/4.17.15#pick)?
    - This pull request also has functionality to modify properties or add additional properties during the serialization process. This is known as the `modify` function.
2. Why not make this a plugin?
    - Plugin support is not yet available in v2 (#851). Once plugin support is added, it's very likely this will be refactored out into a different plugin package.
3. Why not have this be a standalone package?
    - In the future, it would be nice to automatically serialize documents when retrieving them. For example when you run `Model.get` having a setting where the result from that is automatically serialized. This would be impossible to do as a standalone package due to having to interact with Dynamoose directly.
    - The syntax of this solution would also be different than the original goal and inspiration provided in the original issue #797.

I will be keeping this pull request open until at least Tuesday May 19th 2020. If you have additional issues or comments you would like to make about this pull request, please submit a review on this pull request with your comments. However it is important to note that as this pull request is complete and has been open for over a month, there is a high bar for overhauling this pull request at this point.

If no additional comments are made, I will merge this pull request as soon as I'm able to (after the date listed above), and after @ablankenship10 approves the pull request (and marks it as ready (removing draft status)).

ablankenship10:
I think it looks pretty clean and provides a convenient method of modifying how the document will be output in different ways based on various criteria. I'll pull the code tonight and try it out in my demo project for completeness.

dolsem:
Yeah, so my two cents is that ideally this would be a standalone package, and I would avoid serializing automatically on `Model.get()`. There's a way to write it as a wrapper that will provide the same API as discussed in #797, adding methods by subclassing/extending defined models. That all being said,
> it is important to note that as this pull request is complete and has been open for over a month, there is a high bar for overhauling this pull request at this point

is a fair point. So whatever you guys decide.

On a different note, @fishcharlie there was one more point about the behavior of `serializeMany()`. Defining it as a static method on the model seems unclear to me, because nothing prevents the array to include items of type that is different from the type of the underlying model. This could be a global method instead, but then, it might be as simple as `.map(item => item.serialize())`.

lib/Serializer.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export interface SerializerOptions {
+	include?: string[];
+	exclude?: string[];
+	modify?: (serialized: ObjectType, original: ObjectType) => ObjectType;
+}
+
+export class Serializer {
+	#serializers: {[key: string]: SerializerOptions};
+	#defaultSerializer: string;
```
dolsem, line 14: I'm just curious, why not TypeScript's `protected`/`private` for this? It accomplishes the same goal for the API, but leaves some leeway for people who _really_ want to access these. You never know all people's requirements beforehand, and I don't see any harm in enforcing this on the type-checker level instead.

lib/Serializer.ts
```diff
@@ -0,0 +1,119 @@
...
+			},
+			{
+				"if": Boolean(options.exclude),
+				"function": (serialized: ObjectType): ObjectType => utils.object.delete(serialized, options.exclude)
+			},
+			{
+				"if": Boolean(options.modify),
+				"function": (serialized: ObjectType): ObjectType => options.modify(serialized, document)
+			}
+		].filter((item) => item.if).reduce((serialized: ObjectType, item) => item.function(serialized), {...document});
```
dolsem, line 117: I like how easily extensible this approach is, pretty clean and very cool.

lib/Serializer.ts
fishcharlie: @dolsem I think this falls under undocumented things shouldn't be used. I'd love to do this more throughout the code to try to enforce it more. Once plugin support is added, some of those things will probably be reversed and will be more open. But I'd like to get to the point where internal things are **impossible** to access and are marked somehow in the codebase. Anything that is modified that isn't internal or private, should be considered a breaking change. At that point we'd be able to remove that documentation note from the website. But modifying internal things would be allowed and wouldn't be considered a breaking change.

https://dynamoosejs.com/other/FAQ#can-i-use-an-undocumented-property-class-method-or-function-in-dynamoose

fishcharlie: It would also be great at some point to ensure that tests conform to that. So anytime you have to modify or delete a test, that is considered a breaking change. However adding tests is completely allowed.

We aren't quite at that point yet, and there are some cases where the tests are just wrong. So bug fixes require changes to the tests.

So the goal and idea of this is to work towards enforcing more backwards compatibility and ensuring there are less breaking changes (or at least that they are marked and detected correctly).

dolsem: But why prevent people from accessing Dynamoose's internals if they really want to? They might have good reasons for it, and it will be their responsibility to deal with breaking changes since these things are technically not supposed to be accessed. Just like there's reflection in Java, I think there should be a way in JS/TS to access internals if the user can't work around that.

lib/Serializer.ts
ablankenship10: Agreed this is pretty cool!

lib/Serializer.ts
fishcharlie: @dolsem I assume when using 3rd party packages, that when printing out something, that I'm able to access and use everything it displays. In this case if you do `console.log(model.serializer)` it prints out basically an empty object. However, if we were to just set these to TS `private` for JS users there is no signal that these are private properties.

We could add `_` as a prefix. However I'm not so sure that is the best indication anymore.

See my first bullet point here: https://github.com/dynamoose/dynamoose/issues/900#issuecomment-627652795.

So I think there are users who are using internal properties in ways they shouldn't be.

Now if it gets to the point where users wish to access this, or feel like there is something they want to do that they can't. Then we'll cross that bridge when we get to it. But I think the pros of preventing access outweigh the drawbacks. Especially in this case I can't think of any conceivable reason why you'd need to access these things. Maybe there is, but if there is we should create a standard API and have a better way to do it that is documented.

In that comment above, it was a clear indication of the user wanting to do something that wasn't yet supported. Passing a schema into another schema. Which is something I'm working on implementing. But normally when users reach for internal properties, it's an indication that Dynamoose could do something better or isn't doing something correctly. Which should be fixed as opposed to just allowing access to internal things IMO.

fishcharlie:
@ablankenship10 I can merge this after you review & test and mark it as ready for review.

ablankenship10:
> @ablankenship10 I can merge this after you review & test and mark it as ready for review.

Sweet, I'll try and get a test running this morning if I have time. Thanks!

lib/Serializer.ts
dolsem: @fishcharlie yeah, you're right. Forgot about JS for a second.

ablankenship10:
I was able to use as expected in my local testing, good work!

PR #821, opened 2020-04-18
Set with Model.update

### LINKED ISSUE #819: Schema attribute `set` should be executed on `Model.update()`, too
Currently having a `set` function on a schema's attribute it gets executed only on `Model.create()` but not on `Model.updated()`. IMO it should get executed on both functions.

fishcharlie:
This PR fixes an issue where `set` wouldn't be called with `Model.update`.

Closes #819


MickL:
Wohoooo 🥳

PR #831, opened 2020-04-20
Project Updates

fishcharlie:


fishcharlie:
I know this is bad practice to merge without checks passing. But Coveralls is throwing 502 errors (even tho status page is green), and I'm so fed up with Coveralls at this point... wish GitHub would build a solution for code coverage.

Anyways, gonna YOLO merge since tests are passing locally, and the only file that was modified since last commit was README.md, and all tests continue to pass with 100% coverage locally.

PR #846, opened 2020-04-30
throughput was not optional

MickL:


MickL:
There seems to be an error. I tried to fix that but I dont get the code. @fishcharlie Maybe you can take a look?

hardyscc:
@fishcharlie @MickL On the DynamoDB side, we only need to set once `BillingMode: PAY_PER_REQUEST` on the table level, all the subsequent secondary indexes will be auto inherited, therefore better make it optional.

so 
```
	throughput: number | {read: number; write: number};
```
should be
```
	throughput?: "ON_DEMAND" | number | {read: number; write: number};
```
I am happy to help on this PR, if you guys agree with it.

fishcharlie:
> @fishcharlie @MickL On the DynamoDB side, we only need to set once `BillingMode: PAY_PER_REQUEST` on the table level, all the subsequent secondary indexes will be auto inherited, therefore better make it optional.
> 
> so
> 
> ```
> 	throughput: number | {read: number; write: number};
> ```
> 
> should be
> 
> ```
> 	throughput?: "ON_DEMAND" | number | {read: number; write: number};
> ```
> 
> I am happy to help on this PR, if you guys agree with it.

At first glance. This sounds right. Would love to see a PR. Thanks @hardyscc!!

hardyscc:
@MickL please refer to #853

MickL:
Closed in favor of #853

PR #847, opened 2020-04-30
Fix import of aws-sdk

MickL:
The import you used throws an error:

Bildschirmfoto 2020-04-30 um 17 17 50

### DIFF
--- a/lib/aws/sdk.ts
+++ b/lib/aws/sdk.ts
@@ -1,3 +1,3 @@
-import AWS from "aws-sdk";
+import * as AWS from "aws-sdk";
 
 export = AWS;
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,6 @@
 	"compilerOptions": {
 		"module": "commonjs",
 		"moduleResolution": "node",
-		"esModuleInterop": true,
 		"lib": ["es2015"],
 		"target": "es6",
 		"outDir": "dist",


hardyscc:
Add the following will solve the problem

tsconfig.json
```
{
  "compilerOptions": {
    ...
    "esModuleInterop": true,
  },
}
```

I think should not force the user to enable it, better change the import as follow

```
import * as AWS from 'aws-sdk'
```

fishcharlie:
> I think should not force the user to enable it, better change the import as follow

@hardyscc Basically what this PR does, correct? Or something else?

---

@MickL Not sure how to achieve this. But would love to have some type of test or something to ensure this doesn't regress in the future. I'm not sure if there is a flag we could enable in the `tsconfig.json` file to throw an error on things like this? Or if there is another way we could catch this in our CI suite? Any ideas around this?

hardyscc:
@fishcharlie just remove `"esModuleInterop": true,` on your tsconfig.json, then fix all the import problem, let me try make a PR.

MickL:
Yes please not force the user to have set esModuleInterop. It is not set in most projects (like Nest for example) and even tho default imports are forbidden in a lot of tslint / eslint configs.

fishcharlie:
@hardyscc Look forward to seeing that PR! Thanks for that!!

@MickL I do think it's important to have this fail during our TS build process or tests. Otherwise this problem could regress. It sounds like @hardyscc's solution will do that and fail the TS build in this event. Which would allow us to catch this earlier in the process. We wanna get to a point where we catch these things **before** a release, and not after. In this case we caught it after. Would like to have a solution that helps us catch these problems earlier.

MickL:
If you remove the flag it does throw an error, dont build and so no need for extra tests?

hardyscc:
@fishcharlie i found that your current testcase has a fail
```
1437 passing (3s)
  4 pending
  1 failing

  1) utils.empty_function
       Should equal an empty function:

      AssertionError: expected '()=>{cov_rv3i2iwf2().f[0]++;}' to deeply equal '()=>{}'
      + expected - actual

      -()=>{cov_rv3i2iwf2().f[0]++;}
      +()=>{}
      
      at Context.<anonymous> (test/utils/empty_function.js:10:110)
      at processImmediate (internal/timers.js:456:21)
      at process.topLevelDomainCallback (domain.js:137:15)
```

MickL:
I removed the option from tsconfig

hardyscc:
@MickL just fixed under #849

fishcharlie:
> i found that your current testcase has a fail
> 
> ```
> 1437 passing (3s)
>   4 pending
>   1 failing
> 
>   1) utils.empty_function
>        Should equal an empty function:
> 
>       AssertionError: expected '()=>{cov_rv3i2iwf2().f[0]++;}' to deeply equal '()=>{}'
>       + expected - actual
> 
>       -()=>{cov_rv3i2iwf2().f[0]++;}
>       +()=>{}
>       
>       at Context.<anonymous> (test/utils/empty_function.js:10:110)
>       at processImmediate (internal/timers.js:456:21)
>       at process.topLevelDomainCallback (domain.js:137:15)
> ```

@hardyscc I hate that test... 😆. In `test/utils/empty_function.js` try changing line 10 to the following:

```
		expect(utils.empty_function.toString().replace(/\s/gu, "").replace(/cov_.{9,10}\(\)\.f\[0\]\+\+;/gu, "")).to.eql("()=>{}");
```

MickL:
Closed in favor of #849

PR #860, opened 2020-05-03
query.sort

### LINKED ISSUE #848: [FEATURE] Specify .query order (ScanIndexForward)
A query result is ordered by the sortKey (rangeKey). DynamoDB gives the option to query it ascending or descending with [ScanIndexForward](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-ScanIndexForward) but Dynamoose does not have that option.

fishcharlie:
Closes #848.

### DIFF
--- a/docs/docs/guide/Query.md
+++ b/docs/docs/guide/Query.md
@@ -112,6 +112,22 @@ This causes the query to be run on a specific index as opposed to the default ta
 Cat.query("name").eq("Will").using("name-index"); // Run the query on the `name-index` index
 ```
 
+## query.sort(order)
+
+This function sorts the documents you receive back by the rangeKey. By default, if not provided, it will sort in ascending order.
+
+The order parameter must be a string either equal to `ascending` or `descending`.
+
+```js
+Cat.query("name").eq("Will").sort("ascending");
+```
+
+```js
+Cat.query("name").eq("Will").sort("descending");
+```
+
+Under the hood this sets the `ScanIndexForward` property when making the request to DynamoDB. This ensures sorting is done on the database side to optimize results.
+
 ## query.all([delay[, max]])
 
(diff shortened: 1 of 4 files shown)


fishcharlie:
@IMBurbank & @MickL would you both like to review/test this PR and let me know what you think?

MickL:
The code looks good. Ascending does nothing so personally I would remove it. Will test it out on monday!

fishcharlie:
@MickL Just based on questions I’ve seen from doing open source and my experience, it would be confusing to have descending and not ascending. One of the goals of Dynamoose is to hide some of the complexity of AWS-SDK. So as opposed to having to go and research why there is only descending, it’s provided. One of my first questions if it didn’t have ascending would be, why doesn’t it have ascending? And the answer would be, because of how AWS works. And in this case, that isn’t a question I want to answer, and being more explicit doesn’t hurt and adds clarity. Also helps so that users don’t have to understand the minor details of the underlying API.

I was about to be more explicit and set it either way in the request. But then I decided that it would be a better idea to save the few bytes of bandwidth, and didn’t provide value to be explicit with AWS. Being explicit with the user is better tho.

It might also be useful in the future 😉. Who knows.

fishcharlie:
@MickL Let me know how testing goes!!

MickL:
I tested and it works as expected: 

- `.sort('descending')` sets `ScanIndexForward` to `false`
- `.sort('asscending')` does not set `ScanIndexForward` at all

There is an error in `nestjs-dynamoose` because of deprecated interfaces, documented here: https://github.com/hardyscc/nestjs-dynamoose/issues/76

IMBurbank:
I'm planning to test a bit later this morning.

MickL:
"morning" is relative 😜

IMBurbank:
I'm working through a few other changes to see if I can get a migration to v2 working properly (reverted v2 model updates), but this change looks good. Both sorts worked for me.

fishcharlie:
@MickL @IMBurbank Just as a status update for this PR. I'm planning on merging this hopefully by the end of this weekend at the latest and releasing this as part of v2.2.0 which should be out by the end of this weekend as well. Just been kinda busy lately. Apologizes for the delay and I hope to get this out soon.

PR #868, opened 2020-05-04
Model.waitForActive.check should have been optional

MickL:
@fishcharlie After throughput this is the second attribute that should have been optional but was not. If you have time could you please check all interfaces if there are more things that should be optional but are not?

Btw. I dont know if this `Partial` interface is a good idea. Why use `Partial<ModelOptions>` if already `ModelOptions` can set what attributes are optional and which are required?

### DIFF
--- a/lib/Model/index.ts
+++ b/lib/Model/index.ts
@@ -15,7 +15,7 @@ import {DynamoDB, AWSError} from "aws-sdk";
 // Defaults
 interface ModelWaitForActiveSettings {
 	enabled: boolean;
-	check: {timeout: number; frequency: number};
+	check?: {timeout: number; frequency: number};
 }
 export interface ModelExpiresSettings {
 	ttl: number;


fishcharlie:
> If you have time could you please check all interfaces if there are more things that should be optional but are not?

This will come with time. I appreciate you making these PRs. It's super helpful. However, I don't use TypeScript in any of my personal projects. It's extremely hard to catch every little thing like this.

I am trying to work on some ideas that will help improve this process and catch these things earlier. But TypeScript Support is in beta, so I expect a lot of these little glitches. As mentioned on the website, creating PRs or issues is the best way to help here.

> Btw. I dont know if this Partial interface is a good idea. Why use Partial<ModelOptions> if already ModelOptions can set what attributes are optional and which are required?

My thought process here is that the final ModelOptions that we use must have all required properties. For example, the default properties, must conform to ModelOptions. However, when a user passes in options, they aren't required to pass in everything. But the end result after those two get merged, is of type ModelOptions with everything being required.

The goal is for `ModelOptions` to all be required, so that the default options can conform to it and are all required. I know `expires` is an exception to that. Need to fix that at some point.

MickL:
Could you please merge if everything is ok?

Btw if you have different interfaces for user-inputs and internal usage then I would simply create two interfaces for that instead of this partial thing.

fishcharlie:
@MickL 

> Could you please merge if everything is ok?

Of course! I want to clone this down and test one thing tho but haven't gotten enough time to yet 😞. Will try to get around to it soon.

> Btw if you have different interfaces for user-inputs and internal usage then I would simply create two interfaces for that instead of this partial thing.

Why's that? That would just leads to a lot of duplicated code. Why do you think duplicated code is better than `partial`? And just so you know "this partial thing" is part of TypeScript. https://www.typescriptlang.org/docs/handbook/utility-types.html#partialt

MickL:
An interface is not actually code it is just a description. And it is not duplicate: When user input interface and internal interface are different then there should be two interfaces imo.

fishcharlie:
@MickL 

> An interface is not actually code it is just a description.

I'd argue it's not executable code, but is still code. However this is getting into semantics and overall is not worth it in my opinion.

> And it is not duplicate: When user input interface and internal interface are different then there should be two interfaces imo.

I think it's important to realize tho. In the future when we add a new setting or something to the model. The process currently is:

1. Add it to the interface
2. TypeScript will throw an error about how the default settings don't conform to the interface
3. Fix the error by adding the default setting value
4. Use the new setting

If we separate them into two interfaces, this process changes to:

1. Add it to the internal interface
2. TypeScript will throw an error about how the default settings don't conform to the internal interface
3. Fix the error by adding the default setting value
4. Use the new setting
5. Add it to the user facing interface

However, it's **extremely** easy to forget step 5 in this case. There is no test, contract, warning, error, etc. to ensure the relationship between those two interfaces. What if you update one interface and forget to update the other?

Using `Partial`, doesn't have that problem. It's one thing that you update, and if you forget to update the default settings, TypeScript will let you know.

fishcharlie:
@MickL So one question I have about this. If you look at lines 149 or 151 it has `model.options.waitForActive.check.frequency` or `model.options.waitForActive.check.timeout`.

But `check` is now optional. With the new optional features in TS/JS it should likely be `waitForActive.check?.frequency` or something along those lines. If I hover over `check` in VSCode it shows as an optional property.

In other type safe languages as far as I'm aware, it is supposed to catch things like this for you to prevent runtime errors of trying to access a property from an optional object. Like Swift would be an example of catching that beforehand.

What is even more strange, is in a [TypeScript playground](https://www.typescriptlang.org/v2/en/play?#code/JYOwLgpgTgZghgYwgAgKoGdrIN4CgCQIcAthAPwBcOB+MwU6YVjUoA5gNw0A2cjzYViE4EAvl1G5cCAPYhGyAK6YoAQSoYsAXhzjpchcugAhDSuQ68+AERFS1qlZt0GYB8mvGZAI2sAaGmteRndrAGViYDAAC2sxXD19eRluCAA6bhk2AAojNTS7dJdGAEouWWTUjKzclWMCknTgsDKgA) it throws the expected error.

Do you have any ideas why lines 149 & 151 aren't throwing errors since it's now optional?

MickL:
`model.options.waitForActive.check  && model.options.waitForActive.check.frequency === 0 ? await utils.set_immediate_promise() : await utils.timeout(model.options.waitForActive.check.frequency); frequency === 0 ? await utils.set_immediate_promise() : await utils.timeout(model.options.waitForActive.check.frequency);`

fishcharlie:
@MickL That all makes sense. You have to check to ensure the optional property exists before using it. That isn't really my question tho. My question is, on those lines I mentioned, we aren't doing that currently. And with your change, you make `check` optional. So why does TS not catch this (like it does in the playground)? Seems like with your change this should throw a TS build error (due to us not checking that the optional exists).

---

As for your other question, two reasons:

1. Sometimes the linter was throwing a warning about not setting return types for functions. Looks like even after removing this one it works. Don't fully understand that rule yet.
2. The code you mentioned in your second part also is not the same as the first part. The second version will start executing the promise function immediately. Whereas the first one will delay executing the promise function that wrapper function is called. So those two versions are not the same.

MickL:
In the playground it is red for me. TypeScript cant catch that anyway it just helps you preventing this error.

MickL:
Then you can write: `return () => new Promise`

fishcharlie:
@MickL Ahhhh. I found out what the problem is. `strictNullChecks` was set to `false` (by default) in the `tsconfig.json` for Dynamoose.

So it wasn't throwing errors in these cases (like the playground was).

Here is what I think the proposal for how to move forward is:

1. Create a new PR with the following things:
    1. Add `"strictNullChecks": true` to `tsconfig.json`
    2. Fix the build errors that occur after setting that option
    3. Ensure that code coverage remains at 100% while doing this. Might require adding more tests to test those optional cases.
2. Create a new PR that will replace this PR with the following things:
    1. Add `export type DeepPartial<T> = {[P in keyof T]?: DeepPartial<T[P]>};` to `General.ts`
    2. Change `export type ModelOptionsOptional = Partial<ModelOptions>;` to `export type ModelOptionsOptional = DeepPartial<ModelOptions>;`
    3. Make `expires` non optional in `ModelOptions`, and fix errors that occur from this
3. After both PRs get merged, mark this pull request as a duplicate and close

---

> Then you can write: `return () => new Promise`

Again. See my first item I mentioned. In certain cases there is a lint warning/error when you don't set the return type in functions. In this case (for some reason that I don't yet understand) that isn't the case. I want to better understand that lint warning/error about specifying return types for functions, and why it doesn't apply here.

MickL:
> After both PRs get merged, mark this pull request as a duplicate and close

You could also simply push into this PR, too.

> In certain cases there is a lint warning/error when you don't set the return type in functions

I have never had an error when you dont set a return type. The TypeScript compiler is very smart and automatically sets return types:

```
const myVar = "hello";

const myFn = () => {
  return 2;
}

let num: number = myVar; // Error
let txt: string = myFn(); // Error
```

Also note that everything the compiler complains about is configurable.

fishcharlie:
@MickL **Huge** difference between the `The TypeScript compiler` and the linter.

Here is the rule tho:

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md

After willing skimming it, I have no idea why the linter isn't throwing warnings/errors. I have `max-warnings` set to 0, and that rule is on.

MickL:
Is there any progress on the `waitForActive.check` thing?

fishcharlie:
@MickL No updates. Feel free to work through the steps I provided above.

MickL:
Im closing this PR as it is not as easy as I thought. Opened an issue tho: https://github.com/dynamoose/dynamoose/issues/911

PR #885, opened 2020-05-09
Adding access to the original value for set function in Schema settings

### LINKED ISSUE #879: [FEATURE] Add access to the current value of an attribute in a custom attribute setter
### Summary:
In Dynamoose v1.11 (and Mongoose), the value of `this` in an attribute set function is assigned the value of the entire model. This provided a convenient place to salt passwords when they change. It allowed me to short circuit out before having to run a CPU intensive salting algorithm. Also it tied in nicely with passport's local strategy and keeps password handling at a low level in the application.

#### Schema
```js
const UserSchema = new Schema({
  // rest of the user model
  password: {
    type: String,
    required: true,
    set: async function set(value) {
      if (this.password === value) {
        return value;
      }

      const salt = await bcrypt.genSalt(AUTH.SALT_ROUNDS);
      const hashedPassword = await bcrypt.hash(value, salt);
      return hashedPassword;
    },
  },
});
```

However, I'm conscious of the evolution that Dynamoose has taken with v2 and I like and agree with the clear separation of thinking between Schemas and Models. So to keep an attribute scoped within itself and keep the nice separation of concerns which helps keep schema designs focused, I'd like to propose these alternatives.

### Proposed alternatives
I like these because they will not impact existing users of the framework.

#### Add a second parameter to the set callback that is assigned the current value (before setting). Similar to the API of React lifecycle events. Also doesn't necessitate the use of a scoped function.
```js
const UserSchema = new Schema({
  // rest of the user model
  password: {
    type: String,
    required: true,
    set: async (newValue, currentValue) => {
      if (newValue === currentValue) {
        return newValue;
      }

      const salt = await bcrypt.genSalt(AUTH.SALT_ROUNDS);
      const hashedPassword = await bcrypt.hash(newValue, salt);
      return hashedPassword;
    },
  },
});
```

#### Assign the current value (before setting) to `this`
```js
const UserSchema = new Schema({
  // rest of the user model
  password: {
    type: String,
    required: true,
    set: async function set(value) {
      if (this === value) {
        return value;
      }

      const salt = await bcrypt.genSalt(AUTH.SALT_ROUNDS);
      const hashedPassword = await bcrypt.hash(this, salt);
      return hashedPassword;
    },
  },
});
```

It might be worth implementing both approaches and leave it up to library consumers to decide what makes sense to them because I feel they're both valid and align with the library design ideas:

- the first approach of `(newValue, currentValue)` is purely functional.
- `this` in a model is scoped to the model (makes sense). `this` having the value of the attribute also makes semantic sense.
- both approaches are scoped to the attribute, not the model.

But if I had to pick one, I would pick the first approach as it doesn't require a scoped function, which in my opinion is a slight win on the code readability and cognitive load front.

### Environment:

Operating System: `MacOS`
Operating System Version: `10.14.6 Mojave`
Node.js version (`node -v`): `12.14.1`
NPM version: (`npm -v`): `6.13.4`
Dynamoose version: `2.1.2`

### Other:
- [x] I have searched through the GitHub issues (including closed issues) and pull requests to ensure this feature has not already been suggested before

fishcharlie:
Closes #879

### DIFF
--- a/docs/docs/guide/Schema.md
+++ b/docs/docs/guide/Schema.md
@@ -326,6 +326,17 @@ You can use a set function on an attribute to be run whenever saving a document 
 }
 ```
 
+Unlike `get` this method will additionally pass in the original value as the second parameter (if avaiable). Internally Dynamoose uses the [`document.original()`](Document#documentoriginal) method to access the original value. This means that using [`Model.batchPut`](Model#modelbatchputdocuments-settings-callback), [`Model.update`](Model#modelupdatekey-updateobj-settings-callback) or any other document save method that does not have access to [`document.original()`](Document#documentoriginal) this second parameter will be `undefined`.
+
+```js
+{
+	"name": {
+		"type": String,
+		"set": (newValue, oldValue) => `${newValue.charAt(0).toUpperCase()}${newValue.slice(1)}-${oldValue.charAt(0).toUpperCase()}${oldValue.slice(1)}` // Prepend the newValue to the oldValue (split by a `-`) and capitalize first letter of each when saving to database
+	}
+}
+```
+
 ### index: boolean | object | array
 
 You can define indexes on properties to be created or updated upon model initialization. If you pass in an array for the value of this setting it must be an array of index objects. By default no indexes are specified on the attribute.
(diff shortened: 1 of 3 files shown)


fishcharlie:
@ebrearley Mind testing & reviewing this pull request for me?

ebrearley:
Thanks @fishcharlie for the super quick turnaround on this one! Tested this locally and works a treat. Also, appreciate the detailed description in the documentation.

PR #887, opened 2020-05-09
Model.get Attributes

### LINKED ISSUE #881: [FEATURE] Using `Model.getItem()` return only specific attributes (ProjectionExpression)
In DynamoDB there is `ProjectionExpression` (formerly `AttributesToGet`). I couldnt find this in the Dynamoose docs. For .query there is .attributes([]) but not for .getItem()?

Reference: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_GetItem.html

P.S. Could you please add to the docs which request parameter [query.attributes(attributes)](https://dynamoosejs.com/guide/Query#queryattributesattributes) is using? `AttributesToGet` or `ProjectionExpression` or a custom solution? If I read the docs I want to be sure "Oh yes this function is using `ProjectionExpression`, thats exactly what I was looking for!", but currently I have to try it out and use the Dynamoose.logger to verify. The same may go for other functions.

fishcharlie:
Closes #881


fishcharlie:
@MickL Wanna test & review this?

Issue #896, opened 2020-05-12
[FEATURE] Add linter


MickL:
This project may use a linter like eslint or tslint and maybe a code formatter. Also the Gitlab CI pipeline should check for code style.

This issue is meant for discussion and keep progress.

MickL:
Personally i hate Prettier just because it is not configurable. It is opinionated and I dont like a lot of how it styles.

I prefer to use tslint or eslint instead which are both configurable and do archive 90% what prettier does. Currently Angular sticks to tslint even tho it is deprecated because eslint is just not there where tslint is. They also do provide pretty good rules and styles for TypeScript.

I would suggest to:

- use tslint
- copy the Angular tslint rules (and remove what is Angular specific)
- Stick to the [Angular Styleguide](https://angular.io/guide/styleguide) (except what is Angular specific)

pfulop:
I can also take this one since I would love my PR to be linted instead of me going over the formatting manually.

Since tslint is deprecated I can use eslint for this and AFAIK there are @typescript-eslint/parser and @typescript-eslint/eslint-plugin that should achieve what tslint did (I have a project where I went through the migration process).

MickL:
Yes there are plugins but I recently switched vom eslint back to tslint. I totally understand why Angular sticks to it and would not recommend using eslint for ts right now.

fishcharlie:
TSLint has been deprecated in favor of ESLint (https://www.npmjs.com/package/tslint). I will not be merging any pull requests that migrate to using TSLint.

MickL:
eslint TypeScript support is not feature complete. tslint is stable and well tested over the years. Here is a roadmap for typescript-eslint: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md

Anyway I dont mind if we go for eslint

fishcharlie:
@MickL Honestly, I've never heard someone argue for using a deprecated package or API before. That is a first for me. Especially when we are already using the replacement or recommended alternative. Seems strange to be arguing to take a step back...

I am kinda against adding Prettier to the project. It's very opinionated and I tend to disagree with the decisions it makes.

Also, please keep in mind. ESLint already exists in the project. Running `npm run lint` will run ESLint, and it gets run as part of the CI process. So I'm not really sure what the purpose of this issue is.

PR #890, opened 2020-05-10
Strongly typed models

pfulop:
### Summary:
This PR is a result of #889 question. After looking around and trying to find a solution for creating a dynamic interface based on the Schema, and have documents conform to it, I must sadly say that typescript is not there yet to achieve this.

But the basic functionality of strongly typed models is still achievable and this Library was almost there when it comes to type definitions.

So for starters:

_What I want to achieve is:_  
https://medium.com/@tomanagle/strongly-typed-models-with-mongoose-and-typescript-7bc2f7197722

_How I want to achieve it:_
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a2bc1d868d81733a8969236655fa600bd3651a7b/types/mongoose/index.d.ts#L213

~~**Why is this PART 1?**~~
~~I could go all over the project and start correcting (mostly) return types to have correctly typed models and documents, but instead, I've chosen a strategy for solving my problems. This way I can see the library behaving correctly based on real usage. More I will use it, more types I can add.~~

#### Model
```typescript
interface Stat extends Document {
  id: string;
  tasks: number;
  resolvedTasks: number;
  revenue: number;
  payedOut: number;
  users: number;
}

const Stat = dynamoose.model<Stat>(
  'Stat',
  new dynamoose.Schema({
    id: String,
    tasks: Number,
    resolvedTasks: Number,
    revenue: Number,
    payedOut: Number,
    users: Number,
  }),
  { create: false },
);
```

#### General
```typescript
const stat = new Stat({ id: '1', tasks: 0, resolvedTasks: 0, revenue: 0, payedOut: 0, users: 0 });
const stat = await Stat.get('1');
```

### GitHub linked issue:
Closes #889 
#---

### Other information:
Currently constructing and retrieving documents trough get will infer the correct type.

You may also see that mongoose is using any in their `new` signature but I instead chose to use `Partial<T>`  instead so there can be type checks to some extent when creating new Doc.

**IMPORTANT Export**
I had to change export from the index of the library. I am not sure why `export =` was used (feel free to fill me in on this). I needed to export _Document_ since custom interfaces HAVE TO extend _Document_ (and that was also a big issue I had when creating custom interfaces). If it was exported trough `export = { Document }`, typescript wouldn't allow an interface to extend _Document_ since it would treat it as a value and not as a type.

### Type (select 1):
- [ ] Bug fix
- [X] Feature implementation
- [ ] Documentation improvement
- [ ] Testing improvement

[...]

PR #898, opened 2020-05-12
New ESLint rules

pfulop:
### Summary:
This PR adds more lint rules so uniform code style is more easily achievable. It also ads pre-hook to disallow commit if there is a lint error on staged files and also hook to run tests before pushing.

I also run initial lint on the codebase.

### GitHub linked issue:
#896 

### Type (select 1):
- [ ] Bug fix
- [X] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [X] No
- [ ] I'm not sure

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### Other:
- [X] I have read through and followed the Contributing Guidelines
  - [X] `npm test`
  - [X] `npm run lint`
- [X] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoose/dynamoose/blob/master/LICENSE)


.eslintrc.js
```diff
@@ -1,51 +1,41 @@
...
+    "@typescript-eslint/interface-name-prefix": "off",
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-unused-vars": "off",
+    "@typescript-eslint/camelcase": "off",
+    "@typescript-eslint/no-this-alias": "off",
+    "@typescript-eslint/no-var-requires": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "object-curly-spacing": ["error", "never"],
+    "space-before-function-paren": ["error", "never"],
```
fishcharlie, line 27: I think this should be always.

```suggestion
    "space-before-function-paren": ["error", "always"],
```

.eslintrc.js
```diff
@@ -1,51 +1,41 @@
...
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/eslint-recommended",
+    "plugin:@typescript-eslint/recommended",
+  ],
+  rules: {
+    "@typescript-eslint/interface-name-prefix": "off",
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-unused-vars": "off",
```
fishcharlie, line 21: Why did we turn these off?

.eslintrc.js
```diff
@@ -1,51 +1,41 @@
...
+    "@typescript-eslint/camelcase": "off",
+    "@typescript-eslint/no-this-alias": "off",
+    "@typescript-eslint/no-var-requires": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "object-curly-spacing": ["error", "never"],
```
fishcharlie, line 26: I actually do like the look of spaces before and after the import objects and deconstructing assignments. I don't think I like it in array's tho.

What happens if we set this to always? Kinda just wanna see how it looks.

.eslintrc.js
```diff
@@ -1,51 +1,41 @@
...
+    "space-before-function-paren": ["error", "never"],
+    "space-before-blocks": ["error", "always"],
+    "no-multi-spaces": "error",
+    "space-in-parens": "error",
+    "@typescript-eslint/comma-spacing": "error",
```
fishcharlie, line 31: Let's group all the TS rules together

package.json
```diff
@@ -88,5 +90,16 @@
...
+      "pre-commit": "lint-staged",
+      "pre-push": "npm test"
+    }
+  },
+  "lint-staged": {
+    "*.{js,ts}": [
+      "eslint --fix"
+    ]
+  }
+}
```
fishcharlie, line 105:
```suggestion
}

```

publish/index.js
```diff
@@ -16,6 +16,42 @@ const os = require("os");
 const npmFetch = require("npm-registry-fetch");
 let package = require("../package.json");
 
+
+
+async function checkCleanWorkingDir() {
```
fishcharlie, line 21: Any reason why these got moved?

package.json
```diff
@@ -14,6 +14,8 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "eslint": "^6.8.0",
+    "husky": "^4.2.5",
+    "lint-staged": "^10.2.2",
```
fishcharlie, line 18: Curious to hear your thoughts behind this. We used husky in the past, and personally I wasn't a big fan of it. But wanna hear your perspective on it.

package.json
```diff
@@ -88,5 +90,16 @@
...
+  ],
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
+      "pre-push": "npm test"
+    }
+  },
+  "lint-staged": {
+    "*.{js,ts}": [
+      "eslint --fix"
```
fishcharlie, line 102: If we do decide to keep this why not just change this to `npm run lint:fix` to keep it all in one place?

package.json
```diff
@@ -88,5 +90,16 @@
...
-}
+  ],
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
+      "pre-push": "npm test"
+    }
+  },
+  "lint-staged": {
+    "*.{js,ts}": [
```
fishcharlie, line 101: Is there a reason we have to define this? Why can't it be part of the `npm run lint` script? Why do we need to define it in two places?

.eslintrc.js
pfulop: `@typescript-eslint/no-explicit-any` is I believe from the previous eslint. Everything else
is based on [this](https://github.com/nestjs/nest/blob/master/.eslintrc.js) as suggested in https://github.com/dynamoose/dynamoose/issues/896. 

From my standpoint of view:

- interface-name-prefix: I really don't see reason in prefixing interfaces with I
- explicit-function-return-type: this is usually just an extra overhead for the developer from my experience Mainly considering typescript is really good with inferring return types now.
- no-explicit-any: I am personally not a fan of this but I think turning this on would slow down and make this PR big

.eslintrc.js
pfulop: I also like it in object and such, I based this rules on the comments from https://github.com/dynamoose/dynamoose/pull/890 but now I see there were two ways to interpret the comments and I probably chose wrong,

package.json
pfulop: Husky can be a pain in the ass but I find it mostly helping in any kind of project. Most of the issues such as these (https://github.com/dynamoose/dynamoose/pull/890) won't happen with husky anymore (unless somebody skips commit prechecks). 

My thought is to always let tooling take any burden out of the developer's mind. I don't like to think about styling and usually forget about it when I am too focused on some issues. Then I submit PR and instead of moving forward with it I need somebody to correct me on my styling issues, which is just a slowdown and kinda annoying one. So if I can prevent it with anything I will. The argument would be to format files on save (which I also like to do) but for this project, my editor for some reason doesn't format according to the eslint rules.

TL;DR: I like to have consistent peace of mind for all developers that before they create PR their changes follow code styling of the project without much effort from their side.

package.json
pfulop: I tried and it will also check package.json for instance.  Lint-staged supplies files to be checked and it will override `--ext .ts,.js `parameter for eslint.

pfulop:
> How do we feel about turning on all for ESLint instead of recommended?

This currently produces a huge amount of eslint errors that are not auto-fixable. Also, in my personal opinion, it's a bit excessive.

.eslintrc.js
fishcharlie: You do have `@typescript-eslint/no-explicit-any` twice in here tho.

`interface-name-prefix` also didn't cause any problems before. Like we didn't have that disabled in the previous version and everything was working. What changed between last version and this version that required setting that to off?

`explicit-function-return-type`. I think that should be a separate pull request, and I'm not so sure I'd approve it. I don't see the harm in it as of right now, and as a newbie to TS, it does help get the syntax down and get in the habit of declaring return types.

`no-explicit-any`. Agreed. Separate PR would be good. However, it's duplicated twice in the rules. So we need to remove one of them.

.eslintrc.js
```diff
@@ -1,51 +1,42 @@
...
+    "object-curly-spacing": ["error", "always"],
+    "space-before-function-paren": ["error", "always"],
+    "space-before-blocks": ["error", "always"],
+    "no-multi-spaces": "error",
+    "space-in-parens": "error",
+    indent: ["error", "tab"],
+    "linebreak-style": ["error", "unix"],
+    quotes: ["error", "double"],
+	semi: ["error", "always"],
+	"quote-props": ["error", "always"],
```
fishcharlie, line 36: Formatting got messed up here. This is also something that ESLint should catch and throw an error about. So before fixing it. Please make sure that the linter currently throws an error here.

package.json
fishcharlie: @pfulop And how about `lint-staged`? Never heard of that library before.

package.json
```diff
@@ -88,5 +90,16 @@
...
   "files": [
     "dist"
-  ]
+  ],
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
+      "pre-push": "npm test"
+    }
+  },
```
fishcharlie, line 99: Personally. I prefer to have no configuration in the `package.json` file. It just adds complexity to understanding it. Let's either refactor this out into a different `.huskyrc.js` or `husky.config.js` file. I don't have a preference as to which we use. But also using a JS file will allow for comments, which might be helpful in the future.

package.json
```diff
@@ -88,5 +90,16 @@
   },
   "files": [
     "dist"
-  ]
+  ],
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
```
fishcharlie, line 96: So if this gets run before commit. It runs `npm run lint:fix`, however, does it actually stage those files? Or would it run the lint, fix the files, commit, then still have the lint changes in the working copy?

fishcharlie:
Also, we got merge conflicts.

package.json
pfulop: `lint-staged` is a library that runs the specified command (in this case linting) only on staged files, therefore, your PR will have only stuff that you changed (because if linter would run globally and somebody would push without checks, you might style fix his file and it would end up in your PR).

`lint-staged` will also stage files that got fixed by linter that was run by husky. Maybe that's why you didn't like husky because it was used without lint-staged.

.eslintrc.js
pfulop: Added .eslintrc.js to be explicitly included since by default it's ignored by eslint.

pfulop: `interface-name-prefix` yeah I guess it's not needed in this project since interface names are following typescript recommendations.

`explicit-function-return-type` by default it's just a warning. I would let it go through this PR turned off and then maybe create another one for turning it into error if you really think it will help you. Otherwise, in my opinion, it's too much pain for a (mostly none) little gain.

.eslintignore
```diff
@@ -1,3 +1,4 @@
 docs
 dist
 publish/node_modules
+!.eslintrc.js
```
fishcharlie, line 4: New line at end of files.

```suggestion
!.eslintrc.js

```

.eslintrc.js
```diff
@@ -1,51 +1,40 @@
...
-			}
-		}
+	"plugins": ["@typescript-eslint/eslint-plugin"],
+	"extends": [
+		"eslint:recommended",
+		"plugin:@typescript-eslint/eslint-recommended",
+		"plugin:@typescript-eslint/recommended",
 	],
+	"rules": {
+		"@typescript-eslint/explicit-function-return-type": "off",
```
fishcharlie, line 18: Gonna make the call to remove this. We can discuss adding it back later.

```suggestion
```

.huskyrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  hooks: {
```
fishcharlie, line 2: Hmm. This should have failed the linter since it doesn't have quotes around it. Any ideas why it didn't?

```suggestion
  "hooks": {
```

package.json
fishcharlie: This still needs to be done.

pfulop:
I had to go back to eslint command inside lint-staged instead of npm run lint:fix since that one doesn't accept files as an input.

fishcharlie:
> I had to go back to eslint command inside lint-staged instead of npm run lint:fix since that one doesn't accept files as an input.

@pfulop Ahh. We might need to add `--` in front of it or something. Do you have `Allow edits from maintainers` enabled? If so, mind if I push directly to your branch so I can finish this up?

pfulop:
I've sent you an invite (not sure where Allow edits from maintainers setting is). Feel free to push what you need into the branch

fishcharlie:
@pfulop Wanna review #913 & #914? Basically took what you did in this pull request, spilt it out into two different pull requests, and made a few more minor changes. But **really** heavily based on this pull request.

Thank you for your hard work on this. It would have taken me much longer to do those pull requests if I wasn't able to copy and paste from this one.

PR #914, opened 2020-05-23
Adding Husky Hooks

fishcharlie:


fishcharlie:
I do have some hesitations to this pull request. For example, I wasn't able to get it to work in [Tower](https://www.git-tower.com/mac) so at least for my workflow this is completely useless.

fishcharlie:
Considering closing this due to https://github.com/typicode/husky/issues/639#issuecomment-573875583. Don't really want to introduce a tool into Dynamoose that doesn't even work. There has been a **lot** of discussion on that issue but no comment from the owner (@typicode) since January 14th.

Anything that is introduced into Dynamoose adds complexity and is another thing for me to maintain. Which I'm not opposed to, but it's gotta provide benefits that outweigh the drawbacks. I use Tower for most of my Git commands. So the reality of this pull request is it'll increase the complexity in the project for me to maintain, while providing no additional benefits to my workflow. That might sound too self centered, but the reality is that this project isn't at the point of having enough maintainers to really warrant this. If I hear from a lot of other people that this is something they want, I'd be happy to reconsider and would be willing to take this on and help try to maintain it. But until then, it doesn't improve my workflow at all, and would just be one more dependency to maintain (two actually).

Does anyone in the community have any more thoughts on this? Really don't want to introduce a bad tool into the project. Especially one that isn't even getting support or updates from the owner.

cc/ @pfulop

PR #915, opened 2020-05-23
Website icons

fishcharlie:


fishcharlie:
Need to update the spacing between the GitHub Icon and the Dark Mode toggle so it's consistent with the spacing between the GitHub Icon and NPM Icon.

Issue #889, opened 2020-05-09
How to create fully typed model?


pfulop:
### Summary:
Currently, when creating a new model the model type is simply Document. I believed the same approach can be taken as with mongoose and just supply interface when creating new model.

```
export default mongoose.model<IUser>('User', UserSchema);
```

But with dynamoose doing so will result in an error:

> Type 'Stat' does not satisfy the constraint 'Document'.

With mongoose, it is easy to simply have your interface extend mongoose's Document interface, however, I can't seem to find a way to access Document interface in dynamoose as it seems to not be exported.

What is the recommended route to create fully typed Models?

### Other:
- [x] I have read through the Dynamoose documentation before posting this issue
- [x] I have searched through the GitHub issues (including closed issues) and pull requests to ensure this question has not already been raised before
- [x] I have searched the internet and Stack Overflow to ensure this question hasn't been raised or answered before
- [ ] I have filled out all fields above
- [x] I am running the latest version of Dynamoose

fishcharlie:
@pfulop What do you mean by "fully typed model"? Where document properties must conform to the model? For example:

```js
const schema = new dynamoose.Schema({
    "id": String,
    "name": String
});
const User = dynamoose.model("User", schema);

const user = new User({
    "id": 1,
    "age": 18 // This would throw a TypeScript Type Error since `age` does not exist in the Schema so therefore this document doesn't conform to User
});
```

Is that what you mean by "fully typed model"? Or something else?

pfulop:
@fishcharlie Yes that's what I mean. But currently, the types for User are:

`const User: Document & Model<Document> & ModelDocumentConstructor<Document>`

Which translates to

`const User: new (object: {
    [key: string]: any;
}) => Document`

Which means any kind of key will satisfy the constrains.

fishcharlie:
@pfulop Yes. That would be **amazing**. I'd love that feature. It'd be great if it could generate a dynamic interface (or something, not sure if that term is correct) based on the Schema, and have documents conform to it.

Sadly at this point, I'm unsure how to achieve that in TypeScript. I'm still pretty new to TypeScript, so really not sure how to achieve that with my current knowledge.

TypeScript Support/Typings is also in beta for Dynamoose, so it's functionality is limited at this stage.

If you want to take a shot at making a pull request to implement this, that would be great! Otherwise, it's on my radar and something I think would be super cool, but not something that is likely to be done anytime soon without support from the community.

pfulop:
[...]

PR #925, opened 2020-05-29
Strong typed models (original from @pfulop)

hardyscc:
### Summary:

> This PR is a result of #889 question. After looking around and trying to find a solution for creating a dynamic interface based on the Schema, and have documents conform to it, I must sadly say that typescript is not there yet to achieve this.
> 
> But the basic functionality of strongly typed models is still achievable and this Library was almost there when it comes to type definitions.
> 
> So for starters:
> 
> What I want to achieve is:
> https://medium.com/@tomanagle/strongly-typed-models-with-mongoose-and-typescript-7bc2f7197722
> 
> How I want to achieve it:
> https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a2bc1d868d81733a8969236655fa600bd3651a7b/types/mongoose/index.d.ts#L213
> 

I also add support to resolve this issue hardyscc/nestjs-dynamoose#76
and tested the type with this project https://github.com/hardyscc/aws-nestjs-starter

#### Schema
```ts
interface Stat extends Document {
  id: string;
  tasks: number;
  resolvedTasks: number;
  revenue: number;
  payedOut: number;
  users: number;
}

const Stat = dynamoose.model<Stat>(
  'Stat',
  new dynamoose.Schema({
    id: String,
    tasks: Number,
    resolvedTasks: Number,
    revenue: Number,
    payedOut: Number,
    users: Number,
  }),
  { create: false },
);
```

#### Model
```
// Code here
```

#### General
```ts
const stat = new Stat({ id: '1', tasks: 0, resolvedTasks: 0, revenue: 0, payedOut: 0, users: 0 });
const stat = await Stat.get('1');
```

### GitHub linked issue:

#---

### Type (select 1):
- [ ] Bug fix
- [x] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:
  - [ ] `npm test`
  - [ ] `npm run lint`

### DIFF
--- a/lib/Document.ts
+++ b/lib/Document.ts
@@ -307,6 +307,8 @@ Document.objectFromSchema = async function (object: any, model: Model<Document>,
 				value.forEach((subValue, index: number, array: any[]) => {
 					if (index === 0 || typeof subValue !== typeof array[0]) {
 						checkTypeFunction([`${key}.${index}`, subValue]);
+					} else if (keysToDelete.includes(`${key}.0`) && typeof subValue === typeof array[0]) {
+						keysToDelete.push(`${key}.${index}`);
 					}
 				});
 				validParents.push({key});
--- a/test/unit/Document.js
+++ b/test/unit/Document.js
@@ -2263,6 +2263,11 @@ describe("Document", () => {
 				"input": [{"id": 1, "name": ""}, {"type": "toDynamo", "modifiers": ["set"], "required": true}],
 				"error": new Error.ValidationError("name is a required property but has no value when trying to save document")
 			},
+			{
+				"schema": {"id": Number, "friends": {"type": Array}},
+				"input": [{"id": 1, "friends": ["a", "b", "c"]}, {"type": "toDynamo"}],
+				"output": {"id": 1, "friends": []}
+			},
 
 			// Multiple Schemas
 			{


fishcharlie:
@hardyscc Is this ready for me to review? Or you still working on it?

hardyscc:
@fishcharlie Yes please ^^

hardyscc:
@fishcharlie just merged from dynamoose/dynamoose#master

hardyscc:
@fishcharlie Please review it thanks

package.json
```diff
@@ -19,7 +19,7 @@
     "typescript": "^3.8.3"
   },
   "scripts": {
-    "prepare": "npm run build:clean; npm run build",
+    "prepare": "(npm run build:clean || true) && npm run build",
```
fishcharlie, line 22: What is the purpose for `|| true`? I don't think `npm run build:clean` can ever fail. I'm not the best at some of these commands tho. Especially the difference between `&&`, `;`, etc.

lib/Condition.ts
```diff
@@ -83,18 +104,20 @@ export class Condition {
...
 	attribute: (key: string) => Condition;
 	eq: (value: any) => Condition;
-	lt: (value: number) => Condition;
-	le: (value: number) => Condition;
-	gt: (value: number) => Condition;
-	ge: (value: number) => Condition;
+	lt: (value: any) => Condition;
+	le: (value: any) => Condition;
+	gt: (value: any) => Condition;
+	ge: (value: any) => Condition;
```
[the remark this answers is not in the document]
hardyscc: this changes is to support Date (string) compare, can change to `number | string`, keep it any to be more flexible

package.json
hardyscc: This is aim to add windows support for `npm install hardyscc/dynamoose --save`

Please see: 
https://medium.com/@chillypenguin/running-node-npm-scripts-sequentially-on-windows-8737dc24da1f

fishcharlie: @hardyscc Ahh got it. It makes sense for `&&` but not `||`. What happens if we change it to something like the following? Just since I don't think `npm run build:clean` can ever fail (and if it does, we have much bigger problems).

```suggestion
    "prepare": "npm run build:clean && npm run build",
```

lib/Model/index.ts
```diff
@@ -367,18 +367,18 @@ export class Model<T extends DocumentCarrier> {
...
+	batchGet(keys: K[]): Promise<ModelBatchGetDocumentsResponse<T>>;
+	batchGet(keys: K[], callback: CallbackType<ModelBatchGetDocumentsResponse<T>, AWSError>): void;
+	batchGet(keys: K[], settings: ModelBatchGetSettings & {"return": "documents"}): Promise<ModelBatchGetDocumentsResponse<T>>;
+	batchGet(keys: K[], settings: ModelBatchGetSettings & {"return": "documents"}, callback: CallbackType<ModelBatchGetDocumentsResponse<T>, AWSError>): void;
+	batchGet(keys: K[], settings: ModelBatchGetSettings & {"return": "request"}): DynamoDB.BatchGetItemInput;
+	batchGet(keys: K[], settings: ModelBatchGetSettings & {"return": "request"}, callback: CallbackType<DynamoDB.BatchGetItemInput, AWSError>): void;
+	batchGet (keys: K[], settings?: ModelBatchGetSettings | CallbackType<ModelBatchGetDocumentsResponse<T>, AWSError> | CallbackType<DynamoDB.BatchGetItemInput, AWSError>, callback?: CallbackType<ModelBatchGetDocumentsResponse<T>, AWSError> | CallbackType<DynamoDB.BatchGetItemInput, AWSError>): void | DynamoDB.BatchGetItemInput | Promise<ModelBatchGetDocumentsResponse<T>> {
```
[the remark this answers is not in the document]
hardyscc: if K is not given it will default to InputKey
```ts
export class Model<T, K = InputKey> 
```

the reason I made this is to support following typing
```ts
export interface UserKey {
  id: string;
}

export interface User extends UserKey {
  name: string;
  email?: string;
}

@Injectable()
export class UserService {
  constructor(
    @InjectModel('User')
    private userModel: Model<User, UserKey>,
  ) {}

  create(user: User) {
    return this.userModel.create(user);
  }

  update(key: UserKey, user: Partial<User>) {
    return this.userModel.update(key, user);
  }
}
```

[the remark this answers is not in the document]
fishcharlie: @hardyscc The worst thing about generics is it gets so confusing what each of them means 😆. Is `Injectable` and `InjectModel` part of TypeScript?

I also think we need documentation around all of this. One of the goals as of v2 is to have every possible thing be documented. So if this is now possible, I think it needs a lot of documentation around it. Along with any other use cases that this now supports.

lib/Condition.ts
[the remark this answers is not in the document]
fishcharlie: @hardyscc Ahh got it. Makes sense.

fishcharlie:
@hardyscc Take a look at #948. That is a PR to get testing done for TypeScript typings. If you wanna review that, then I can merge and we can pull it into here to get some tests going to ensure this stuff doesn't have any regressions.

lib/Model/index.ts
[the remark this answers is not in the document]
hardyscc: It is actually nestjs server framework decorator, you don't need to care about, let me remove them.

package.json
hardyscc: Yes sure can change to that,
this is just in case clean fail, the latter still run.

hardyscc:
Should i merge in the master first, since now it is have some behind

lib/Model/index.ts
```diff
@@ -86,22 +86,22 @@ async function getTableDetails (model: Model<DocumentCarrier>, settings: {allowE
 
 	return model.latestTableDetails;
 }
-async function createTableRequest (model: Model<DocumentCarrier>): Promise<DynamoDB.CreateTableInput> {
+async function createTableRequest (model: Model<any, any>): Promise<DynamoDB.CreateTableInput> {
```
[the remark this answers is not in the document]
hardyscc: actually it is very hard to make the generic typing to also work well with explicit type, so use 'any' is the trick. i will try further review them agin

fishcharlie:
> I agree to combine this two PRs.
> however, i think we should i merge in the master first, since now it have some behinds

@hardyscc Up to you. Will likely have to remerge in after #948 as well tho. Order doesn't matter.

hardyscc:
just merged in #948, can start write some test case for type.

hardyscc:
@fishcharlie I think it is not necessary to fulfils the use case for nestjs-dynamoose, I can just maintain another copy over there, this means only small change is necessary to fix all your current typing issue, if you think that is ok, I can remake this PR.

fishcharlie:
@hardyscc 

> I think it is not necessary to fulfils the use case for nestjs-dynamoose, I can just maintain another copy over there

Not quite sure what you mean by that

hardyscc:
@fishcharlie I just don't want to apply too much changes at once, perhaps just do the phase 1 first, which only solve the query chain problem, this fulfil most of the related issue.

hardyscc:
@fishcharlie this PR is ready for review, which only fix the query chain issue, 
```ts
this.model.scan().exec();
```
Please aware most of the changes are outdated, so please revisit again.
I will create another PR for next phase. lets keep each of the PK small for easy discussion :)

lib/index.ts
```diff
@@ -13,8 +13,8 @@ import ModelStore = require("./ModelStore");
 import {ModelType} from "./General";
 
 const model = <T extends Document>(name: string, schema?: Schema | SchemaDefinition | (Schema | SchemaDefinition)[], options: ModelOptionsOptional = {}): ModelType<T> => {
```
fishcharlie, line 15: Is this `<T extends Document>` even used anymore?

lib/DocumentRetriever.ts
```diff
@@ -332,15 +305,17 @@ DocumentRetriever.prototype.all = function (this: DocumentRetriever, delay = 0,
...
+export interface Scan<T> extends DocumentRetriever, BasicOperators<Scan<T>> {
+	exec(): Promise<ScanResponse<T>>;
+	exec(callback: CallbackType<ScanResponse<T>, AWSError>): void;
+}
 
-export class Scan extends DocumentRetriever {
-	exec(): Promise<ScanResponse<Document[]>>;
-	exec(callback: CallbackType<ScanResponse<Document[]>, AWSError>): void;
-	exec (callback?: CallbackType<ScanResponse<Document[]>, AWSError>): Promise<ScanResponse<Document[]>> | void {
+export class Scan<T> extends DocumentRetriever {
```
fishcharlie, line 313: Is there a reason for having an interface and a class? Isn't there a way to combine these into 1 thing? Or?

hardyscc:
this test case shouldn't fail, please help retry.

test/types/Model.ts
```diff
@@ -9,3 +10,30 @@ const shouldFailWithNumberAsName = dynamoose.model(1);
...
+	})
+);
+
+UserModel.create({"id": "1", "name": "john", "age": 25});
+
+UserModel.get({"id": "1"});
+
+UserModel.query("name").eq("john").exec();
+
+UserModel.scan().exec();
```
fishcharlie, line 39: New line at end of file. See #950.

```suggestion
UserModel.scan().exec();

```

test/types/Model.ts
```diff
@@ -9,3 +10,30 @@ const shouldFailWithNumberAsName = dynamoose.model(1);
 const shouldFailWithBooleanAsName = dynamoose.model(true);
 
 const shouldSucceedWithOnlyPassingInName = dynamoose.model("User");
+
+interface User extends Document {
+	id: string;
+	name: string;
+}
+
```
fishcharlie, line 18:
```suggestion
```

test/types/Model.ts
```diff
@@ -9,3 +10,30 @@ const shouldFailWithNumberAsName = dynamoose.model(1);
...
+const UserModel = dynamoose.model<User>(
```
fishcharlie, line 19: Let's change this to be `TypedModel`. That way it's more clear what is going on.

```suggestion
const TypedModel = dynamoose.model<User>(
```

test/types/Model.ts
```diff
@@ -9,3 +10,30 @@ const shouldFailWithNumberAsName = dynamoose.model(1);
...
+			"type": String,
+			"index": {
+				"global": true
+			}
+		},
+		"age": Number
+	})
+);
+
+UserModel.create({"id": "1", "name": "john", "age": 25});
```
fishcharlie, line 33: I think that all of these should have test names for them. For example.

```suggestion
const shouldCreateSuccessfully = TypedModel.create({"id": "1", "name": "john", "age": 25});
```

fishcharlie: I think we should also have a failing test here. Shouldn't the following fail?

```js
const shouldNotFailWithNotAllAttributesPassedIn = TypedModel.create({"id": "1"});
// @ts-expect-error
const shouldFailWithInvalidAttributes = TypedModel.create({"id": "1", "random": "string"});
```

test/types/Model.ts
```diff
@@ -9,3 +10,30 @@ const shouldFailWithNumberAsName = dynamoose.model(1);
...
+
+UserModel.get({"id": "1"});
+
+UserModel.query("name").eq("john").exec();
```
fishcharlie, line 37: I'm thinking this along with `scan` should be separated out into a different file.

test/types/Model.ts
```diff
@@ -9,3 +10,30 @@ const shouldFailWithNumberAsName = dynamoose.model(1);
 const shouldFailWithBooleanAsName = dynamoose.model(true);
 
 const shouldSucceedWithOnlyPassingInName = dynamoose.model("User");
+
+interface User extends Document {
```
fishcharlie, line 14: Trying to figure out how to best format this. Maybe something like the following? To make it clear it's a new section.

```suggestion
// Typed Models

interface User extends Document {
```

lib/ModelStore.ts
```diff
@@ -2,9 +2,9 @@ import CustomError = require("./Error");
 import {Model} from "./Model";
 import {Document} from "./Document";
 
-let models: {[name: string]: Model<Document>} = {};
+let models = {};
```
fishcharlie, line 5: I think this should be typed somehow. The current type is `{}`. Which leads to an issue where something like `models["number"] = 1;` is allowed, when it shouldn't be.

test/types/Model.ts
hardyscc: currently create() still don't take the generic type, my plan is in next phase, since there are too much changes.

test/types/Model.ts
hardyscc: yes ok, since this is not the finally, we can always go back and review them again.

lib/ModelStore.ts
hardyscc: ok i will put it back

test/types/Model.ts
fishcharlie: @hardyscc Does anything take the generic type yet? Or is this PR just preparation for that?

hardyscc:
ok committed, currently create() still don't take the generic type yet, my plan is in next phase, since I think there are too much changes. anyway would u like to put it in this PR too?

fishcharlie:
@hardyscc Ok. This is getting too complicated. I don't understand what this PR is doing anymore. I also don't think the original description is accurate anymore.

I just created #954 & #955 to do 2 of the things this PR does. Those seemed to be to be separate from what the original goal of this PR was.

So here is how I propose we move forward:

1. You review #954 & #955 
2. We merge those two PRs
3. We close this PR, and you reset your branch back to `dynamoose/master`
4. You implement strongly typed models. And have it be as large of a PR as you want, **but** sticking to strongly typed models. If there is any code that isn't related to strongly typed models, have it be a different PR. I'm not opposed to large PRs, unless they aren't focused on the goal and task they set out to do. That is why I broke it out into #954 & #955 since those weren't relevant to this PR at all.

For step 4, it needs to be complete. Tests, code changes, and documentation. And it shouldn't be half done. Each PR should be complete to the point where it achieves it's intended goal. We've run into problems previously where a half done PR will get merged, and the author of the PR never comes back to complete it. Which leads to more work for me, that I might not have the bandwidth to do. So PRs do need to be complete.

hardyscc:
sounds good.

fredvollmer:
@fishcharlie @hardyscc Was this work ever taken on in another PR? If it would be helpful, I'd love to get involved in this work.

hardyscc:
@fredvollmer It is already been merged under #983

you can try out on master branch
```
npm install dynamoose/dynamoose#master
```

PR #941, opened 2020-06-08
Update to use preview branch

hross:
I fixed the stale bot closing date logic but wanted to do a targeted release/test with known repos that have issue before I release a new version. Do you mind helping me test the logic and confirming it works for you?


hross:
I merged the changes to latest based on some other testing so this is not needed anymore.

PR #955, opened 2020-06-20
Cleaning up duplicated code for Condition & DocumentRetriever

fishcharlie:


hardyscc:
@fishcharlie Im waiting for this to merge :)

lib/DocumentRetriever.ts
```diff
@@ -16,7 +16,10 @@ interface DocumentRetrieverTypeInformation {
 	type: DocumentRetrieverTypes;
 	pastTense: string;
 }
+
 // DocumentRetriever is used for both Scan and Query since a lot of the code is shared between the two
+// type DocumentRetriever = BasicOperators;
+interface DocumentRetriever extends BasicOperators {} // eslint-disable-line @typescript-eslint/no-empty-interface
```
fishcharlie, line 22: @hardyscc I'd like to fix this before merging. I had a solution that was better, but it had a problem since `settings` is used on both DocumentRetriever and Condition.

Hopefully I'll find time to fix this soon. But that is why it hasn't been merged yet.

PR #985, opened 2020-08-03
Nested arrays with multiple data types

bobbyrc:
### Summary:
When creating a schema with an attribute containing a nested array with multiple data types, Dynamoose incorrectly threw an error thinking that the array was actually a nested index.
This small fix checks to see if the nested property is an array before throwing an error about nested indexes.

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:
  - [x] `npm test`
  - [x] `npm run lint`


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dynamoose",
-  "version": "2.3.0",
+  "version": "2.3.1",
```
fishcharlie, line 3: This needs to be reverted. This (along with `package-lock.json` version) will be bumped automatically as part of our automated release process.

```suggestion
  "version": "2.3.0",
```

lib/Schema.ts
```diff
@@ -582,7 +582,8 @@ export class Schema {
 		}
 
 		this.attributes().forEach((key) => {
-			if (key.includes(".") && this.getAttributeSettingValue("index", key)) {
+			const attributeSettingValue = this.getAttributeSettingValue("index", key);
+			if (key.includes(".") && attributeSettingValue && !Array.isArray(attributeSettingValue)) {
```
fishcharlie, line 586: So the problem with this is that in the event you have an index with multiple types, it won't throw an error. For example, the following schema should throw an error `Index must be at root object and not nested in object or array.`.

```js
new dynamoose.Schema({
    "id": String,
    "obj": {
        "type": Object,
        "schema": {
            "data": [
                {"type": Array, "schema": [String]},
                {"type": String, "index": true}
            ]
        }
    }
});
```

However, with the current code, it doesn't look like it'll throw an error. I think the solution is something like:

```suggestion
			const attributeSettingValue = this.getAttributeSettingValue("index", key);
			if (key.includes(".") && (Array.isArray(attributeSettingValue) ? attributeSettingValue.some((singleValue) => Boolean(singleValue)) : attributeSettingValue)) {
```

That way if it's an array, we are checking to see if `index` exists on any of those types.

test/unit/Schema.js
```diff
@@ -76,6 +76,10 @@ describe("Schema", () => {
 		expect(() => new dynamoose.Schema({"id": String, "friend": {"type": Object, "schema": {"name": {"type": String, "index": {"global": true}}}}})).to.throw("Index must be at root object and not nested in object or array.");
 	});
 
+	it("Should not throw error if passing multiple data types for a nested array attribute", () => {
+		expect(() => new dynamoose.Schema({"id": String, "friends": {"type": Object, "schema": {"names": [{"type": Array, "schema": [String]}, {"type": Array, "schema": [Number]}]}}})).to.not.throw();
+	});
+
```
fishcharlie, line 82: Above I gave another example of a case where it should fail. We should add that as a test, and make sure it throws an error. But we should make it fit this pattern (using the name `friends`, etc.).

We also need to confirm that that test fails without the change I mentioned above, and passes with the change I mentioned above.

test/unit/Schema.js
```diff
@@ -76,6 +76,10 @@ describe("Schema", () => {
 		expect(() => new dynamoose.Schema({"id": String, "friend": {"type": Object, "schema": {"name": {"type": String, "index": {"global": true}}}}})).to.throw("Index must be at root object and not nested in object or array.");
 	});
 
+	it("Should not throw error if passing multiple data types for a nested array attribute", () => {
```
fishcharlie, line 79: Just to confirm, this test fails without the changes you already made, correct?

package.json
bobbyrc: Ah gotcha. Sorry if I overlooked that somewhere!

test/unit/Schema.js
bobbyrc: Yep, this test currently fails without the array check and passes with it.

lib/Schema.ts
bobbyrc: For sure. Apologizes for missing this, it was difficult to jump into dynamoose and get my bearings quickly. There's a ton of complexity everywhere throughout the code. Impressed seeing the work as I was looking through.

test/unit/Schema.js
bobbyrc: Added a test to account for this case!

lib/Schema.ts
fishcharlie: @bobbyrc No problem at all! It's not easy jumping into a new codebase. No need at all to apologize for that. Truly appreciate your help on this!! It only gets easier from here haha.

bobbyrc:
Commenting to remove stale tag

fishcharlie:
Made a lot of progress on this PR. Added some more tests which revealed some potential bugs. Currently it looks like Dynamoose is only respecting the first entry in the array of entries. So if in your schema you allow an array of strings and an array of numbers, and pass in an array of numbers, it will throw an error (when it should be accepting that).

The only thing to do to complete this PR is to fix it so those tests pass, and I wanna take one more scan of the code changes after we complete that.

---

I've kinda traced the problem to the following line:

https://github.com/dynamoose/dynamoose/blob/0202bdaa5dbc5558a7c363536896731e7da5e3b3/lib/Schema.ts#L405

In some cases, that is returning `false`, when it should be returning `true`. In other cases it correctly returns `false`.

Digging further. The `getValueTypeCheckResult` method, `key` is equal to `friends.names.1`. Which does not seem right. That causes `typeDetails` to be a string type.

image

In reality I expect this to be an array of array types. Like the following.

image

One *possible* solution would be to change the key from `${fullKey}.${subKey}` to `fullKey`. Problem with that, is I *think* that will cause a lot of regressions. Not positive on that tho.

I forget exactly how this logic all works, and all the different edge cases we need to account for.

I'm not sure why this is causing problems now either. All of this code was tested to work with multiple types. I think it could be the fact that we have an array we are passing in as the property value in the document, and an array of possible types. And when you do `friends.names.1` or `friends.names.0` or whatever, it's not sure if it should be referring to the value, or the type. But there are reasons why I doubt that. Like the fact that it's obvious what should be going on, and what it should be referring to. However it is possible that it's just a bug in the code that hasn't surfaced until now.

Point being, I'm not really sure at this point what is going on.

fishcharlie:
Disregard what I said. This is fixed in 0b703d87d093193caecf17780b42bb445d17705f.

@bobbyrc Could you review (and possibly test) the changes I made and let me know what you think?

fishcharlie (approved):

fishcharlie (approved):

PR #988, opened 2020-08-05
Update returns and schema settings saveUnknown

felix-sg:
### Summary:
Hello, We had an issue when using the Document.update feature after updating to version 2.3.
The problem was we had our models set to saveUnknown:true. So when we call the update function, we expect it to return the objects whole.

After some research in in the Document Class I found out that the function `Document.objectFromSchema => checkTypeFunction` was not using the schema setting saveUnknown correctly as the value in the settings was undefined.

I hope you will understand and consider my edits.
Thanks from ShareGroop

### Type (select 1):
- [X] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [ ] No
- [X] I'm not sure

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### Other:
- [X] I have read through and followed the Contributing Guidelines
- [X] I have ensured the following commands are successful from the root of the project directory
  - [X] `npm test`
  - [X] `npm run lint`
- [X] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoose/dynamoose/blob/master/LICENSE)

### DIFF
--- a/lib/Document.ts
+++ b/lib/Document.ts
@@ -315,7 +315,7 @@ Document.objectFromSchema = async function (object: any, model: Model<Document>,
 			}
 		} else {
 			// Check saveUnknown
-			if (!settings.saveUnknown || !utils.dynamoose.wildcard_allowed_check(schema.getSettingValue("saveUnknown"), key)) {
+			if (settings.saveUnknown === false || !utils.dynamoose.wildcard_allowed_check(schema.getSettingValue("saveUnknown"), key)) {
 				keysToDelete.push(key);
 			}
 		}
--- a/test/unit/Document.js
+++ b/test/unit/Document.js
@@ -2558,6 +2558,11 @@ describe("Document", () => {
 				"output": {"id": 1, "data": "hello world"}
 			},
 			{
+				"input": [{"id": 1, "data": "hello world"}, {"customTypesDynamo": true, "checkExpiredItem": true, "type": "fromDynamo"}],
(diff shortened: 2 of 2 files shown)


fishcharlie:
@felix-sg Was this issue occurring on v2.2.1?

felix-sg:
> @felix-sg Was this issue occurring on v2.2.1?

We were on version one previously. I can't tell about version 2.2

fishcharlie:
@felix-sg First of all, apologizes in this taking so long for me to review.

I'm also assuming you mean `Model.update` in your description there.

This isn't really how I envisioned `objectFromSchema` to work. As you can see with the rest of the conditionals in that function they are just checking for a truthy value. If the bug is specifically related to `Model.update`, that is where the change needs to take place.

```js
{"name": "$SET", "operator": " = ", "objectFromSchemaSettings": {"validate": true, "enum": true, "forceDefault": true, "required": "nested", "modifiers": ["set"]}},
```

Maybe by adding `"saveUnknown": true` to `objectFromSchemaSettings` in that object?? That is on line 650 of `lib/Model/index.ts`.

The test that you wrote should also be moved into a `Model.update` test, and the test you wrote should fail as currently written.

Let me know if you have any questions regarding this. I hope that makes sense for what the goal is here. I do think it's more complicated than it possibly needs to be, but the goal is to make it flexible to where you can define these different settings depending on use case, since that was a problem I believe we had in v1.

Issue #1049, opened 2020-11-25
TypeError: key.split is not a function


softmarshmallow:
Schema
``` ts

export const NestedLayer = new dynamoose.Schema({
    nodeId: String,
    key: String,
    index: Number,
    name: String,
    sdkVersion: String,
    node: String,
    type: {
        "type": String,
        "enum": ["INSTANCE", "GROUP", "VANILLA"]
    },
    layers: {
        type: Array,
        required: false,
        schema: [{
            type: Object,
            schema: dynamoose.THIS as any
        }]
    },
    componentId: String,
    width: Number,
    height: Number,
}, {
    saveUnknown: true,
})

export const SceneScheam = new dynamoose.Schema({
    id: String,
    projectId: String,
    fileId: String,
    nodeId: String,
    sdkVersion: String,
    //DesignPlatformType
    designPlatform: {
        "type": String,
        "enum": ["com.figma.Desktop", "com.bohemiancoding.sketch3", "xyz.bridged.bridged"]
    },
    cachedPreview: String,
    // SceneType
    sceneType: {
        "type": String,
        "enum": ["SCREEN", "COMPONENT", "DOCS"]
    },
    route: String,
    path: String,
    name: String,
    description: String,
    tags: {
        "type": Set,
        "schema": [String]
    },
    alias: String,
    variant: String,
    layers: {
        type: Array,
        schema: [NestedLayer as any]
    },
    width: Number,
    height: Number

}, {
    saveUnknown: true
})

export const Scene = dynamoose.model(TABLE, SceneScheam);


```

```
TypeError: key.split is not a function
    at /Users/-/Documents/GitHub/services/node_modules/dynamoose/lib/utils/dynamoose/wildcard_allowed_check.ts:4:25
    at Array.find (<anonymous>)
    at Object.module.exports [as wildcard_allowed_check] (/Users/-/Documents/GitHub/services/node_modules/dynamoose/lib/utils/dynamoose/wildcard_allowed_check.ts:3:30)
    at checkTypeFunction (/Users/-/Documents/GitHub/services/node_modules/dynamoose/lib/Document.ts:329:50)
    at Array.map (<anonymous>)
    at Function.<anonymous> (/Users/-/Documents/GitHub/services/node_modules/dynamoose/lib/Document.ts:334:111)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/-/Documents/GitHub/services/node_modules/dynamoose/dist/Document.js:5:58)
[...]

PR #1050, opened 2020-11-25
Update wildcard_allowed_check.ts

softmarshmallow:
### Summary:

Fixes https://github.com/dynamoose/dynamoose/issues/1049

### GitHub linked issue:

#---
https://github.com/dynamoose/dynamoose/issues/1049

### Type (select 1):
- [ V ] Bug fix

- [ ] 🚨 YES 🚨
- [ V ] No
- [ ] I'm not sure

- [ V ] Yes
- [ ] No

- [ V ] Yes
- [ ] No

### Other:
- [ V ] I have read through and followed the Contributing Guidelines
- [ V ] I have searched through the GitHub pull requests to ensure this PR has not already been submitted
- [ V ] I have updated the Dynamoose documentation (if required) given the changes I made
- [ V ] I have added/updated the Dynamoose test cases (if required) given the changes I made
- [ V ] I have ensured the following commands are successful from the root of the project directory
  - [ V ] `npm test`
  - [ V ] `npm run lint`
- [ V ] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoose/dynamoose/blob/master/LICENSE)
- [ V ] All of my commits and commit messages are detailed, explain what changes were made, and are easy to follow and understand
- [ V ] I have filled out all fields above

### DIFF
--- a/lib/utils/dynamoose/wildcard_allowed_check.ts
+++ b/lib/utils/dynamoose/wildcard_allowed_check.ts
@@ -1,6 +1,10 @@
 export = (saveUnknown: string[] | boolean, checkKey: string, settings = {"splitString": ".", "prefixesDisallowed": true}): boolean => {
 	if (Array.isArray(saveUnknown)) {
 		return Boolean(saveUnknown.find((key) => {
+			// explicit type checking for typescript "split is not a function" error.
+		    	if(typeof key !== 'string'){
+		                return true
+			}
 			const keyParts = key.split(settings.splitString);
 			const checkKeyParts = checkKey.split(settings.splitString);
 			let index = 0, keyPart = keyParts[0];
--- a/lib/utils/dynamoose/wildcard_allowed_check.ts
+++ b/lib/utils/dynamoose/wildcard_allowed_check.ts
@@ -2,8 +2,8 @@ export = (saveUnknown: string[] | boolean, checkKey: string, settings = {"splitS
 	if (Array.isArray(saveUnknown)) {
 		return Boolean(saveUnknown.find((key) => {
 			// explicit type checking for typescript "split is not a function" error.
-			if (typeof key !== 'string'){
-				return true
+			if (typeof key !== "string") {
+				return true;
 			}
 
 			const keyParts = key.split(settings.splitString);
--- a/lib/utils/dynamoose/wildcard_allowed_check.ts
+++ b/lib/utils/dynamoose/wildcard_allowed_check.ts
@@ -2,9 +2,10 @@ export = (saveUnknown: string[] | boolean, checkKey: string, settings = {"splitS
 	if (Array.isArray(saveUnknown)) {
 		return Boolean(saveUnknown.find((key) => {
 			// explicit type checking for typescript "split is not a function" error.
-		    	if(typeof key !== 'string'){
-		                return true
+			if (typeof key !== 'string'){
+				return true
 			}
+
 			const keyParts = key.split(settings.splitString);
 			const checkKeyParts = checkKey.split(settings.splitString);
 			let index = 0, keyPart = keyParts[0];


fishcharlie (changes requested):
Thanks for the contribution! Few things.

1. For the problem you are having, what does `key` equal and what is the type?
2. We need tests for this PR.

softmarshmallow:
@fishcharlie Merge it or don't. i modified one single line, which does the thing. there is reported issue, you may ignore if you want I'm using the modified version anyway.

fishcharlie:
> @fishcharlie Merge it or don't. i modified one single line, which does the thing. there is reported issue, you may ignore if you want I'm using the modified version anyway.

@softmarshmallow Ok. I've never seen someone contribute a PR to a project, then complain that they won't make perfectly reasonable requested changes. That is a new one for me.

**First of all** I want to make it very clear that I **TRULY** appreciate your contribution here. I appreciate everyone contributing to Dynamoose.

That being said, you made some false accusations in your comment that I do not appreciate and I don't feel are appropriate. I apologize in advance if this comes across as firm, but these things need to be said for the sake of software quality, which I care about for this project.

Part of collaborative coding is peer reviews. It helps everyone grow as developers, and helps everyone learn. Dismissing PRs and making the comments you did, is only hurting yourself.

> Merge it or don't

I will not merge it as is. I have an obligation to the Dynamoose community to not merge code that doesn't meet the projects quality guidelines.

Let's pretend I merge this, and it breaks everyones code because in some cases `key` is not a string, but should return false. I'm the one responsible at that point. And when people complain that Dynamoose broke their projects, I'm the one who is gonna have to deal with it. Do you want to take over that responsibility? That would be even more work than just making the few changes I asked for.

What I'm asking is not unreasonable. Add some tests, and in the issue you reported, add a simple `console.log(key, typeof key);` and report back the results, so I can make sure this change is good, and there isn't a larger problem here. Those two things are not asking very much.

> i modified one single line

Guess what? Code can fail with just 1 single line. Even from the best developers. That is why peer reviews, automated tests, etc. exist. I know there are billions of examples of this, but in 2017 a **single** typo broke the internet: https://www.theverge.com/2017/3/2/14792442/amazon-s3-outage-cause-typo-internet-server.

I don't believe the number of lines is relevant here.

> which does the thing

Again, I **REALLY appreciate** that. I'm glad this fixes a bug. And I really want to merge this. I really do.

That being said, I wanna make sure of two things:

1. There is not a larger problem at play here (which is why I need the information about what `key` is set to and the type)
2. There won't be regressions in the future and this code is covered in tests (which is why I'm requesting you add tests to this PR)

I believe there truly could be a larger issue at play here. Because I wrote the majority of this code. And I know that `wildcard_allowed_check.ts` takes in the `saveUnknown` setting that you pass into the Schema. And that should only ever be an array of strings, or a boolean (see the [documentation](https://dynamoosejs.com/guide/Schema) for information about that).

So the fact that there would be an array, with an item inside that isn't a string, is a huge problem. Just bypassing it by returning `true`, doesn't seem to be fixing the root issue here. This code you added should **NEVER** get run. But it is for you. So I believe there is a larger problem here.

I could be wrong. But based on my knowledge of the code, I think that could be the case. Which is why I'm asking for those 2 things.

> you may ignore if you want

I'm not ignoring it. If I was ignoring it, you wouldn't even have a review on this PR. I gave you feedback, it's your job as the author of this PR to address the feedback and we keep working towards the best possible version to merge in.

I also am slightly offended by this comment. I care a lot about Dynamoose. I never **want** to ignore it. **Every** single bug in Dynamoose also exists in my own code. I don't think anyone wants bugs in their code. So I never want to ignore an issue. And even suggesting that I do really takes away from all the hard work I've put into this project that I care so much about.

> I'm using the modified version anyway

Great. That is your choice. It's open source. You are welcome to do as you wish so long as you follow the license, GitHub TOS, local laws, etc.

Keep in mind as I said, I think there might be a bigger issue at play here tho. So if you would prefer to not have my help in tracking down that larger issue, that is fine. But just be aware, I think there is a chance you are running code that has other problems and side-effects from that fix. Not 100% sure on that, but if I were you, I'd want to verify and make sure I'm running good code.

Finally, please be aware that I don't provide support for forked or custom versions of Dynamoose. So you are on your own when it comes to that. When the issue template says `Dynamoose version` that has to be a version released on NPM on the official Dynamoose package.

---

Feel free to address the feedback I gave you, and add tests to increase the quality of your PR, OR you can close the PR. Either way, I'm **not** ignoring your PR, and I'm **not** merging code that doesn't have tests.

I'm here to help you. If you need help with ideas for tests and how to do it, please let me know, and I'd gladly assist. But not caring about quality then getting your PR merged is not an option and not going to happen.

You might not believe tests are important, but I do. I think it's important for the quality of software. And maybe you don't care about the quality of the software and code you write, but I do care about it for this project.

---

I hope you will consider making the requested changes and working with me to get this fixed for all the Dynamoose users. We are a team here. Instead of dismissing my review and making comments about how I can ignore it if I want, please work with me to get some high quality code merged into Dynamoose and get this fixed.

softmarshmallow:
Thanks for such kind reply. Your code, you'll know it the best. Take a deep look, Please fix this. in recurse modeling that cuases fail to run. And making the type sure with "if" more than just letting it crash, I'm not sure which will break the internet more. Closing.

PR #1141, opened 2021-03-10
Returned saved document

### LINKED ISSUE #1135: [BUG] Model.create returns the Partial provided - not the record stored in the database
### Summary:

when calling `Model.create()`  - an instance of `Document` is created [with the provided partial.](https://github.com/dynamoose/dynamoose/blob/v2.7.0/lib/Model/index.ts#L934). This new document's `.save()` method is immediately called. This asynchronous function goes through the process of taking this partial and creating a document for it, and then it returns the instance of the document of [which it was called.](https://github.com/dynamoose/dynamoose/blob/e2ec6eaf726cce741da101883617bbfd2c21dd98/lib/Document.ts#L188-L193)

This document was created with the provided partial - and not the record added to dynamodb. This can cause problems when using default values in the model, as they will not exist on the partial that is returned. This behavior is slightly different than what is depicted in the [documentation.](https://dynamoosejs.com/guide/Model#modelcreatedocument-settings-callback)

#### Schema
```
import { Schema } from 'dynamoose';
import { nanoid } from 'nanoid';

const userSchema = new dynamoose.Schema({
  id: {
    hashKey: true,
    type: String,
    default: () => nanoid(), 
  },
  createdAt: {
    type: Date,
    required: true,
    default: new Date(),
  },
  lastLogIn: {
    type: Date,
    required: true,
    default: new Date(),
  },
  username: {
    type: String,
    required: true,
    index: {
      name: 'usernameIndex',
      global: true,
    },
  },
});
```

#### Model
```
const UserModel= dynamoose.model('User', userSchema);
```

#### General
```
const partial = {
  username: 'someusername',
  password: 'myhashedpassword',
  someUselessKey: 'someUselessValue',
};

const newUser = await UserModel.create(partial);

// new User is an instance of Document with only the properties provided in partial
console.log(newUser);
console.log(newUser.id); // undefined
console.log(newUser.createdAt); // undefined
console.log(newUser.someUselessKey); // 'someUselessValue'
```

The instance returned has all the properties (and only the properties) of the partial used for construction.

The default and calculated values defined in the model would exist as the return value of of calling `Model.create()`. 

### Environment:

Operating System: Windows 10
Operating System Version: Version	10.0.19042 Build 19042
Node.js version (`node -v`): 14.15.1
NPM version: (`npm -v`): 6.14.8
Dynamoose version: 2.7.0

It seems like this can be avoided by passing the `request` option to the create setting and then resolving the promise in your code, or just by simply querying for the document you just created with the partial data provided.

I found stale #929 which appears to be related.

### Other:

fishcharlie:
### Summary:

This PR changes the behavior to return the saved document instead of document passed in for `Model.create` and `document.save`.

### GitHub linked issue:

Closes #1135

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [ ] No
- [x] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:
  - [x] `npm test`
  - [x] `npm run lint`

### DIFF
--- a/lib/Document.ts
+++ b/lib/Document.ts
@@ -140,8 +140,11 @@ export class Document {
 			settings = {};
 		}
 
+		let savedItem;
+
 		const localSettings: DocumentSaveSettings = settings;
 		const paramsPromise = this.toDynamo({"defaults": true, "validate": true, "required": true, "enum": true, "forceDefault": true, "combine": true, "saveUnknown": true, "customTypesDynamo": true, "updateTimestamps": true, "modifiers": ["set"]}).then((item) => {
+			savedItem = item;
 			let putItemObj: DynamoDB.PutItemInput = {
 				"Item": item,
 				"TableName": this.model.name
@@ -182,13 +185,22 @@ export class Document {
 		if (callback) {
 			const localCallback: CallbackType<Document, AWSError> = callback as CallbackType<Document, AWSError>;
 			promise.then(() => {
-				this[internalProperties].storedInDynamo = true; localCallback(null, this);
+				this[internalProperties].storedInDynamo = true;
+
+				const returnDocument = new this.model.Document(savedItem as any);
+				returnDocument[internalProperties].storedInDynamo = true;
+
+				localCallback(null, returnDocument);
 			}).catch((error) => callback(error));
 		} else {
 			return (async (): Promise<Document> => {
 				await promise;
 				this[internalProperties].storedInDynamo = true;
-				return this;
+
+				const returnDocument = new this.model.Document(savedItem as any);
+				returnDocument[internalProperties].storedInDynamo = true;
+
+				return returnDocument;
 			})();
 		}
 	}
--- a/test/unit/Document.js
+++ b/test/unit/Document.js
@@ -170,6 +170,16 @@ describe("Document", () => {
 					expect(result).to.eql(user);
 				});
 
+				it("Should return correct result after saving with defaults", async () => {
+					putItemFunction = () => Promise.resolve();
+
+					User = dynamoose.model("User", {"id": Number, "name": String, "defaultValue": {"type": String, "default": "Hello World"}});
+					user = new User({"id": 1, "name": "Charlie"});
+
+					const result = await callType.func(user).bind(user)();
+					expect(result.toJSON()).to.eql({"id": 1, "name": "Charlie", "defaultValue": "Hello World"});
+				});
+
 				it("Should return request if return request is set as setting", async () => {
 					const result = await callType.func(user).bind(user)({"return": "request"});
 					expect(putParams).to.eql([]);
--- a/test/unit/Model.js
+++ b/test/unit/Model.js
@@ -1878,6 +1878,15 @@ describe("Model", () => {
 		];
 		functionCallTypes.forEach((callType) => {
 			describe(callType.name, () => {
+				it("Should return correct result after saving with defaults", async () => {
+					createItemFunction = () => Promise.resolve();
+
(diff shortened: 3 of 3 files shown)


lib/Document.ts
```diff
@@ -182,13 +185,22 @@ export class Document {
 		if (callback) {
```
heythisispaul, line 185: Is there a reason you're not able to make `.save()` async at the top? It'd help with readability and flow if you could just `await` and keep everything at the same scope. This could probably be simplified down to something like:

```typescript
try {
  await promise;
  this[internalProperties].storedInDynamo = true;
  
  const returnDocument = new this.model.Document(savedItem as any);
  returnDocument[internalProperties].storedInDynamo = true;
  if (callback) {
      const localCallback: CallbackType<Document, AWSError> = callback as CallbackType<Document, AWSError>;
      localCallback(null, returnDocument);
      return;
    }
  return returnDocument;
} catch (error) {
  if (callback) {
    callback(error);
    return;
  } 
  throw error;
}
```

heythisispaul:
@fishcharlie thanks for responding to my issue report so quickly, really appreciate the work you're doing here.

You mentioned on #1135 that you didn't like this implementation, what don't you like about it?

From what I can tell, it doesn't seem like a breaking change since this behavior mimics what was recorded in the docs, and it's still adhering to the pre-existing return type of the `.save()` method, so I don't think it's changing your external API. If you wanted to be double sure, you could just make this implementation the effect of passing a `resultFromDynamo` or something flag on the `settings` param. If that setting is not passed keep the existing behavior.

lib/Document.ts
fishcharlie: @heythisispaul You can't `await` unless you are within an `async` function. I forget **exactly** what the rational was behind not making these functions `async` but I think there was a purpose behind it at the time.

It is a good point tho and it exists in most Dynamoose functions.

I think it might be a good topic to bring up for v3. It'd be interesting to know how many users are actually using callbacks in Dynamoose. And would users be willing to use [`util.callbackify`](https://nodejs.org/dist/latest-v14.x/docs/api/util.html#util_util_callbackify_original) as an alternative if we drop native support for callbacks.

fishcharlie:
@heythisispaul 

> thanks for responding to my issue report so quickly, really appreciate the work you're doing here

Of course! Thanks for the appreciation, really awesome to know you like Dynamoose.

I ran into this problem myself right before I created this PR, so that always helps with the speed of these things.

> You mentioned on #1135 that you didn't like this implementation, what don't you like about it?

Just feels messy to me. I was playing around originally with the idea of having a setting that gets passed into `this.toDynamo` that will mutate the object (`this`). So that when defaults are added or properties removed, it would mutate the object itself. That way you aren't having to hold onto two versions of this object, and manage both.

That turned out to be more complicated than I expected. So I went down this path.

I think I'm gonna merge it as is tho. It does the job, and any improvements can come later.

---

Did you get a chance to test this to make sure it resolves your issue?

lib/Document.ts
heythisispaul: > You can't await unless you are within an async function. 

Yeah, sorry I think ultimately I was asking why this whole function could/would not be declared as `async` in it's definition. 

Now that I look at this again, it would mean that this function would change the implementation slightly since `void` would no longer be a valid return value, only `Promise<void>`.

I think it's fair to consider dropping callback support as you mentioned, but yeah that's a v3 deal.

heythisispaul:
@fishcharlie Yeah I see what you're saying, now that I look through this a little deeper. Juggling with both version can lead to some extra overhead. I think all in all it's still a valid solution.

> Did you get a chance to test this to make sure it resolves your issue?

Just did! Looks good, thanks.  👍

PR #1180, opened 2021-04-15
Fix type for waitForActive option on dynamoose.Model

rajivshah3:
### Summary:

The documentation for the `waitForActive` option on `dynamoose.Model` says this:

> Settings for how DynamoDB should handle waiting for the table to be active before enabling actions to be run on the table. This property can also be set to `false` to easily disable the behavior of waiting for the table to be active. For production environments we recommend setting this value to `false`.

If I understand correctly, I think this should be possible:
```ts
const modelOptions = {
	create: false,
	waitForActive: false,
}

const myModel = dynamoose.model<Model1>("model1", mySchema, modelOptions);
```

However, TypeScript gives me an error:

> Argument of type '{ create: boolean; waitForActive: boolean; }' is not assignable to parameter of type `'DeepPartial<ModelOptions>'`.
  Types of property 'waitForActive' are incompatible.
    Type 'boolean' has no properties in common with type `'DeepPartial<ModelWaitForActiveSettings>'`. ts(2345)

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:
  - [x] `npm test`
  - [x] `npm run lint`

### DIFF
--- a/lib/Model/index.ts
+++ b/lib/Model/index.ts
@@ -37,7 +37,7 @@ export interface ModelOptions {
 	throughput: "ON_DEMAND" | number | {read: number; write: number};
 	prefix: string;
 	suffix: string;
-	waitForActive: false | ModelWaitForActiveSettings;
+	waitForActive: boolean | ModelWaitForActiveSettings;
 	update: boolean | ModelUpdateOptions[];
 	populate: string | string[] | boolean;
 	expires: number | ModelExpiresSettings;
--- a/lib/Model/index.ts
+++ b/lib/Model/index.ts
@@ -317,7 +317,7 @@ export class Model<T extends DocumentCarrier = AnyDocument> {
 			setupFlow.push(() => createTable(this));
 		}
 		// Wait for Active
-		if (typeof this.options.waitForActive !== "boolean" && (this.options.waitForActive || {}).enabled) {
+		if ((this.options.waitForActive || {}).enabled) {
 			setupFlow.push(() => waitForActive(this, false));
 		}
 		// Update Time To Live
--- a/lib/Model/index.ts
+++ b/lib/Model/index.ts
@@ -317,7 +317,7 @@ export class Model<T extends DocumentCarrier = AnyDocument> {
 			setupFlow.push(() => createTable(this));
 		}
 		// Wait for Active
-		if ((this.options.waitForActive || {}).enabled) {
+		if (this.options.waitForActive === true || (this.options.waitForActive || {}).enabled) {
 			setupFlow.push(() => waitForActive(this, false));
 		}
 		// Update Time To Live
--- a/lib/Model/index.ts
+++ b/lib/Model/index.ts
@@ -37,7 +37,7 @@ export interface ModelOptions {
 	throughput: "ON_DEMAND" | number | {read: number; write: number};
 	prefix: string;
 	suffix: string;
-	waitForActive: ModelWaitForActiveSettings;
+	waitForActive: false | ModelWaitForActiveSettings;
 	update: boolean | ModelUpdateOptions[];
 	populate: string | string[] | boolean;
 	expires: number | ModelExpiresSettings;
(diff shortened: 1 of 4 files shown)


lib/Model/index.ts
```diff
@@ -163,23 +163,25 @@ function waitForActive (model: Model<DocumentCarrier>, forceRefreshOnFirstAttemp
 	return (): Promise<void> => new Promise((resolve, reject) => {
 		const start = Date.now();
 		async function check (count: number): Promise<void> {
-			try {
-				// Normally we'd want to do `dynamodb.waitFor` here, but since it doesn't work with tables that are being updated we can't use it in this case
-				const tableDetails = (await getTableDetails(model, {"forceRefresh": forceRefreshOnFirstAttempt === true ? forceRefreshOnFirstAttempt : count > 0})).Table;
-				if (tableDetails.TableStatus === "ACTIVE" && (tableDetails.GlobalSecondaryIndexes ?? []).every((val) => val.IndexStatus === "ACTIVE")) {
-					return resolve();
+			if (typeof model.options.waitForActive !== "boolean") {
```
fishcharlie, line 166: If this is just a TypeScript type error, I'm not sure these code level changes are needed.

lib/Model/index.ts
```diff
@@ -37,7 +37,7 @@ export interface ModelOptions {
 	throughput: "ON_DEMAND" | number | {read: number; write: number};
 	prefix: string;
 	suffix: string;
-	waitForActive: ModelWaitForActiveSettings;
+	waitForActive: false | ModelWaitForActiveSettings;
```
fishcharlie, line 40: Not sure if this should be a boolean instead. I forget what the behavior is for `true`.

lib/Model/index.ts
```diff
@@ -315,7 +317,7 @@ export class Model<T extends DocumentCarrier = AnyDocument> {
 			setupFlow.push(() => createTable(this));
 		}
 		// Wait for Active
-		if ((this.options.waitForActive || {}).enabled) {
+		if (typeof this.options.waitForActive !== "boolean" && (this.options.waitForActive || {}).enabled) {
```
fishcharlie, line 320: Same as above.

fishcharlie (changes requested):
Couple of things:

1. Code coverage declined (likely due to the code changes). We need to fix that to either add tests, or remove those changes.
2. We have a `test/types` folder that houses TypeScript tests. We probably need to create some tests in there to verify this works and doesn't have regressions in the future.
3. In the Model documentation `waitForActive` is of type `Object`. But in the description it talks about `false`. We should change the documentation as well to represent the correct behavior here.

rajivshah3:
Hi, I'll make those changes in a few days when I have some more time

fishcharlie:
Thanks @rajivshah3!

lib/Model/index.ts
rajivshah3: I thought about that, but I think `true` would have the same behavior as `false` in this case because of this line:

https://github.com/dynamoose/dynamoose/blob/40443acc6f7a07681592f294d3893902c3652e34/lib/Model/index.ts#L318

Because `this.options.waitForActive.enabled` would be `undefined` if I understand correctly

lib/Model/index.ts
rajivshah3: This part was just to make TypeScript happy, because if I remove it then I get these errors:

```
lib/Model/index.ts:177:33 - error TS2339: Property 'check' does not exist on type 'false | ModelWaitForActiveSettings'.
  Property 'check' does not exist on type 'false'.

177                     model.options.waitForActive.check.frequency === 0 ? await utils.set_immediate_promise() : await utils.timeout(model.options.waitForActive.check.frequency);
                                                    ~~~~~

lib/Model/index.ts:177:143 - error TS2339: Property 'check' does not exist on type 'false | ModelWaitForActiveSettings'.
  Property 'check' does not exist on type 'false'.

177                     model.options.waitForActive.check.frequency === 0 ? await utils.set_immediate_promise() : await utils.timeout(model.options.waitForActive.check.frequency);
                                                                                                                                                                  ~~~~~

lib/Model/index.ts:179:58 - error TS2339: Property 'check' does not exist on type 'false | ModelWaitForActiveSettings'.
  Property 'check' does not exist on type 'false'.

179             if (Date.now() - start >= model.options.waitForActive.check.timeout) {
                                                                      ~~~~~                                                                    
```

rajivshah3:
Hi @fishcharlie , sorry for the delay. I've replied to your comments and am ready for another review

lib/Model/index.ts
rajivshah3: I think I can revert this change given the behavior I mentioned in https://github.com/dynamoose/dynamoose/pull/1180#discussion_r631234611. I originally thought there would be an error if it tried to access an `enabled` property on `true`

rajivshah3: Reverted in 49cca6b3 but I don't think it'll show up on this PR until it's reopened

rajivshah3:
I tried writing a unit test to bring the code coverage back to 100% but I'm not sure if it's correct 🤔

lib/Model/index.ts
fishcharlie: @rajivshah3 In that case tho, the documentation needs to be updated to say `false` instead of true. Currently the documentation say it's accepts a boolean, but this says it only accepts `false`.

We probably need to fix that line you linked as well and add some tests.

fishcharlie (changes requested):

lib/Model/index.ts
fishcharlie: 566c42b8ca7807553574e1e802cd811dcf1bff5b

fishcharlie:
@rajivshah3 I just pushed a few changes to your branch. Would you mind reviewing it and let me know what you think?

rajivshah3:
> In that case tho, the documentation needs to be updated to say false instead of true. Currently the documentation say it's accepts a boolean, but this says it only accepts false.

Ooh ok, I see now. I think I was misreading the documentation before and thought that only `false` or an object were allowed


> Would you mind reviewing it and let me know what you think?

Looks good to me! Thanks so much for the help

fishcharlie (approved):

fishcharlie:
@rajivshah3 Awesome. Release will hopefully be coming soon. I really apologize for how long it took me to get this merged in. Personal stuff has just been consuming a lot of my free time recently.

Thanks so much for your contribution!!

rajivshah3:
Thanks! No worries and sorry for the delay on my side as well

lib/Model/index.ts
fishcharlie: @rajivshah3 Looks like this actually isn't correct. Because if `model.options.waitForActive` is a boolean (like this PR is implementing), nothing will happen. We don't have any else condition here to handle this.

fishcharlie: Put up a PR for this, but it's causing errors with other tests: #1271. No idea why.

rajivshah3: Ah yikes, sorry about that. I think I may have made this change under the impression that `waitForActive` would either be `false` or a `object` since that was the inference I made from the docs at that time.

fishcharlie: @rajivshah3 No problem. Mind reviewing/testing that PR I linked? It should be good to go once Coveralls comes back online from their downtime.

Issue #1020, opened 2020-09-30
[FEATURE] Where `defaultReturnValues` is in version 2


ealanisg:
### Summary:

When updating a model I'm getting back all the document whereas it is not needed for my purposes as I could set in the request `"ReturnValues": "NONE",` but it seems that dynamoose is leaving it as default `"ReturnValues": "ALL_NEW"`

I see that the option was available in version 1 but looks like being removed for version 2
https://github.com/dynamoose/dynamoose/blob/2286b41d0554c770d1d733117711c52dcf33df31/CHANGELOG.md#version-140

### Reference:
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html

### Environment:
Dynamoose version: 2.3.0

### Other:
- [X] I have read through the Dynamoose documentation before posting this issue
- [X] I have searched through the GitHub issues (including closed issues) and pull requests to ensure this question has not already been raised before
- [X] I have searched the internet and Stack Overflow to ensure this question hasn't been raised or answered before
- [X] I have filled out all fields above
- [X] I am running the latest version of Dynamoose

camiloaz:
Any update on this?

Nimesh-K-Makwana:
Having the same issue. Even after setting the defaultReturnValues = "NONE", still it takes "ALL_NEW" as default. any update how to solve this issue?

fishcharlie:
I think this was fixed in #1193. Therefore I'm closing this issue. Please let me know if this is incorrect.

PR #1193, opened 2021-05-02
Add returnValues option to Model.update settings

Canas:
### Summary:
Currently there is no way to set `ReturnValues` in `Model.update` params as `ALL_NEW` is hardcoded in `updateItemParams`.

To address this, a `returnValues` option is added to `Model.update` local settings, which is able to overried `ReturnValues` while leaving `ALL_NEW` as the default there are no regressions.

#### General
```typescript
// to use just add returnValues to the settings object
MyModel.update({id: 1}, {name: 'Tom'}, { returnValues: "NONE" });
```

### GitHub linked issue: https://github.com/dynamoose/dynamoose/issues/1020

---Closes

### Type (select 1):
- [ ] Bug fix
- [x] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:
  - [x] `npm test`
  - [x] `npm run lint`

### DIFF
--- a/docs/docs/guide/Model.md
+++ b/docs/docs/guide/Model.md
@@ -471,7 +471,7 @@ You can also pass in a `settings` object parameter to define extra settings for 
 |------|-------------|------|---------|
 | return | What the function should return. Can be `document`, or `request`. In the event this is set to `request` the request Dynamoose will make to DynamoDB will be returned, and no request to DynamoDB will be made. | String | `document` |
 | condition | This is an optional instance of a Condition for the update. | [dynamoose.Condition](Condition.md) | `null`
-| returnValues | Set which documents to return after the update. Valid values are: `NONE`, `ALL_OLD`, `UPDATED_OLD`, `ALL_NEW` `UPDATED_NEW` | String | `ALL_NEW`
+| returnValues | Set which documents to return after the update. This setting will be passed into the DynamoDB `ReturnValues` parameter. | String | `ALL_NEW`
 
 There are two different methods for specifying what you'd like to edit in the document. The first is you can just pass in the attribute name as the key, and the new value as the value. This will set the given attribute to the new value.
 
--- a/docs/docs/guide/Model.md
+++ b/docs/docs/guide/Model.md
@@ -471,6 +471,7 @@ You can also pass in a `settings` object parameter to define extra settings for 
 |------|-------------|------|---------|
 | return | What the function should return. Can be `document`, or `request`. In the event this is set to `request` the request Dynamoose will make to DynamoDB will be returned, and no request to DynamoDB will be made. | String | `document` |
 | condition | This is an optional instance of a Condition for the update. | [dynamoose.Condition](Condition.md) | `null`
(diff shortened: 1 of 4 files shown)


docs/docs/guide/Model.md
```diff
@@ -471,6 +471,7 @@ You can also pass in a `settings` object parameter to define extra settings for
 |------|-------------|------|---------|
 | return | What the function should return. Can be `document`, or `request`. In the event this is set to `request` the request Dynamoose will make to DynamoDB will be returned, and no request to DynamoDB will be made. | String | `document` |
 | condition | This is an optional instance of a Condition for the update. | [dynamoose.Condition](Condition.md) | `null`
+| returnValues | Set which documents to return after the update. Valid values are: `NONE`, `ALL_OLD`, `UPDATED_OLD`, `ALL_NEW` `UPDATED_NEW` | String | `ALL_NEW`
```
fishcharlie, line 474:
```suggestion
| returnValues | Set which documents to return after the update. This setting will be passed into the DynamoDB `ReturnValues` parameter. | String | `ALL_NEW`
```

Couple things here. Making it clear how it'll be used in relation to DynamoDB, and removing the valid DynamoDB options. If DynamoDB were to ever update their values, this documentation would be outdated. I tend to like to keep our documentation as focused on Dynamoose as possible. There are a lot of drawbacks to repeating the DynamoDB documentation in my opinion.

lib/Model/index.ts
```diff
@@ -239,6 +239,7 @@ interface ModelBatchPutSettings {
 interface ModelUpdateSettings {
 	return?: "document" | "request";
 	condition?: Condition;
+	returnValues?: DynamoDB.ReturnValue;
```
fishcharlie, line 242: Super clever. Part of me really loves this. Less to maintain, and passing that responsibility onto DynamoDB.

My only feeling is that this type of concept of exposing DynamoDB types within a Dynamoose type isn't done at all before this. So I'm slightly hesitant to do this. But I can't think of any good argument for why it could be a bad thing. If AWS were to break it somehow, we have much bigger issues.

fishcharlie (changes requested):
This looks great! One documentation change, and one more thing.

Can we add a few tests to `test/types` for this? That way the ModelUpdateSettings change you made will have less likelihood for regression.

fishcharlie:
Thanks for your contribution @Canas! Really appreciate it.

Canas:
> Can we add a few tests to test/types for this? That way the ModelUpdateSettings change you made will have less likelihood for regression.

Thanks for the suggestion, I originally didn't do this because under the hood `DynamoDB.ReturnValue` is just `string`, they don't enforce constant values. Nevertheless, it might change as you said so I added the tests.

Wondering if should I add this link along your doc suggestion directly;
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html#DDB-UpdateItem-request-ReturnValues

Canas:
Is there anything missing in the PR @fishcharlie ?

fishcharlie:
@Canas I don't believe so. I was out of town all last week. I'll try to do a final review soon and get this merged in.

Issue #851, opened 2020-05-01
Plugin Support


fishcharlie:
# Plugin Support

## Overview

In v1 of Dynamoose we had beta support for plugins. This issue is to explore bringing them back and to discuss possible implementation options. I'm really looking for comments about this. What would you like to see in Dynamoose plugin support? What were your thoughts/feelings about v1 plugins? Do you have any suggestions about how to clean up the code, make it more DRY and improve it from v1 plugins?

## Existing v1 Plugins

Although v2 plugins will have breaking changes. It is a goal to be able to support the use cases for these plugins from v1:

- https://github.com/dolsem/dynamoose-plugin-optimistic-locking
- https://github.com/mogwai/dynamoose-caching

## Goals

- Have a cleaner, simpler, and DRYer implementation and API than v1
- Have support for custom types (this is something that Mongoose does, where plugins can expose a type(s) and users can use that custom type(s) in their own schemas, however we did not do this in v1, would like to see this done in v2)
- Better definition about what should be considered a plugin vs part of the core library

## Challenges

https://github.com/dynamoose/dynamoose/blob/9e5fb1605e99da2d5550cfa896d224c09aaa3cc8/lib/Model.js#L618-L643

24 lines to support plugins in v1 for what should be a 1 line task. It adds a **lot** of bloat to Dynamoose. None of that code was DRY, and having to emit on everything felt really messy. I'm very against going back to that old way of doing it. Adding that much code and bloat just to support plugins does not seem worth it to me. I would love to support plugins tho. It just needs to be done in a way that is reasonable, clean, and elegant.

Kinda same thing with logging support in v2. We only log AWS requests and responses currently in v2. So 1 line of code became 4 because of logging support.

https://github.com/dynamoose/dynamoose/blob/ec9379affcd6f7a59509fcaf6ced680110001089/lib/aws/ddb/internal.ts#L29-L32

Which is one of the reasons why I haven't expanded the logging support very much in v2. I realize that there aren't too many good solutions to get around this stuff, but both of those are just so messy. If we are gonna add this to the library, I want it to be clean and elegant.

So really looking for advice, suggestions, and feedback here.

## References

- #325 (original v1 implementation PR)
- https://mongoosejs.com/docs/plugins.html

dolsem:
[...]

PR #1196, opened 2021-05-03
fix: Include the reasons for dynamodb transaction cancellation exception

KennethWKZ:
### Summary:
Extract more error details for `dynamoose.transaction` when it throw transaction cancellation exception from `aws-sdk`.

#### General
```
	const func = ddb()[method](params);

	// Retrieve Dynamodb Transactions Cancellation Reasons Error if method related to transaction
	if (method === "transactGetItems" || method === "transactWriteItems") {
		func.on("extractError", ({error, httpResponse}) => {
			if (error) {
				const {CancellationReasons} = JSON.parse(httpResponse.body.toString());
				error.CancellationReasons = CancellationReasons;
			}
		});
	}

	log({"level": "debug", "category": `aws:dynamodb:${method}:request`, "message": JSON.stringify(params, null, 4), "payload": {"request": params}});
	const result = await func.promise();
	log({"level": "debug", "category": `aws:dynamodb:${method}:response`, "message": typeof result === "undefined" ? "undefined" : JSON.stringify(result, null, 4), "payload": {"response": result}});
	return result;
```

When it throw transaction cancellation errors, it will show cancellationReasons in the error object as well:
```
{
      "message": "Transaction cancelled, please refer cancellation reasons for specific reasons [None, ConditionalCheckFailed]",
      "code": "TransactionCanceledException",
      "time": "2021-05-03T02:24:12.020Z",
      "requestId": "HIFE278IO8GV1G7UJ3ONKDUU33VV4KQNSO5AEMVJF66Q9ASUAAJG",
      "cancellationReasons": [
        {
          "Code": "None"
        },
        {
          "Code": "ConditionalCheckFailed",
          "Message": "The conditional request failed"
        }
      ],
      "statusCode": 400,
      "retryable": false,
      "retryDelay": 43.816932240631616
}
```

### Type (select 1):
- [ ] Bug fix
- [x] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:
  - [x] `npm test`
  - [x] `npm run lint`


lib/aws/ddb/internal.ts
```diff
@@ -26,10 +26,22 @@ async function main (method: "transactGetItems", params: DynamoDB.TransactGetIte
...
+          if (error) {
+            const { CancellationReasons } = JSON.parse(httpResponse.body.toString());
+            error.cancellationReasons = CancellationReasons;
+          }
+        } catch (err) {
+            throw err;
+        }
+    });
+
+    log({ "level": "debug", "category": `aws:dynamodb:${method}:request`, "message": JSON.stringify(params, null, 4), "payload": { "request": params } });
```
fishcharlie, line 41: This line shouldn't have been modified

lib/aws/ddb/internal.ts
```diff
@@ -26,10 +26,22 @@ async function main (method: "transactGetItems", params: DynamoDB.TransactGetIte
...
+    const result = await func.promise();
+    log({ "level": "debug", "category": `aws:dynamodb:${method}:response`, "message": typeof result === "undefined" ? "undefined" : JSON.stringify(result, null, 4), "payload": { "response": result } });
+    return result;
```
fishcharlie, line 44: These lines shouldn't have been modified.

lib/aws/ddb/internal.ts
```diff
@@ -26,10 +26,22 @@ async function main (method: "transactGetItems", params: DynamoDB.TransactGetIte
...
+    func.on("extractError", ({ error, httpResponse }) => {
+        try {
+          if (error) {
+            const { CancellationReasons } = JSON.parse(httpResponse.body.toString());
+            error.cancellationReasons = CancellationReasons;
+          }
+        } catch (err) {
+            throw err;
+        }
+    });
```
fishcharlie, line 39: First of all **LOVE** how this is contained in only one place in the code. That is really awesome. When I first thought about this issue, I didn't think to put it in here. Really clever, and **really love it**. So great job!

However, I can't decide if I'm a fan of having this get called on every request made to AWS. It seems like it adds a lot of overhead and could impose performance risks.

That being said, I'm not sure of a better solution. Ideally we would implement this functionality as a plugin. And the plugin would handle it, so that it isn't being run on every request. But plugin support isn't yet available in Dynamoose. So not really sure how to handle this.

Do you have any thoughts?

fishcharlie (changes requested):
Thanks for your contribution! Few things tho.

1. Doesn't pass lint checks
2. Doesn't pass automated tests
3. Doesn't add any tests as required in our [contributing guidelines](https://github.com/dynamoose/dynamoose/blob/master/CONTRIBUTING.md)
4. Also left some code level comments

---

For the future, in the PR template it asks you if you have run `npm test` & `npm run lint` and to verify that they are successful. Please do me a favor in the future and make sure to run those commands yourself and be honest on the PR template. I know it can be a hassle, but it's really important and helps save everyone time on these type of things, and truly does make the code better. So I'd really appreciate that.

lib/aws/ddb/internal.ts
KennethWKZ: Fixed.

lib/aws/ddb/internal.ts
KennethWKZ: Fixed lint

KennethWKZ (commented):
Fixed

KennethWKZ:
@fishcharlie I added an if condition only extract error when method is transactGetItems & transactWriteItems, so will reduce the overload from other methods

lib/aws/ddb/internal.ts
fishcharlie: @KennethWKZ Got it. Interesting that aws-sdk v3 fixed this. Makes me question if we should even merge this in. aws-sdk v3 is already implemented in Dynamoose v3 alpha.

I'm honestly unsure if v2 is going to have any more releases. It might, but just unsure at this point. Really wanting to shift focus to v3.

I'm going to keep this PR open for now, and if we do end up making another v2 release, we can merge it in. Otherwise, I'd suggest using your own fork in the meantime.

KennethWKZ: ok thanks! @fishcharlie

fishcharlie:
@KennethWKZ It looks like we are still getting test errors here.

KennethWKZ:
> @KennethWKZ It looks like we are still getting test errors here.

@fishcharlie The failed test is not related to my PR, it was failed at this commit id - 7e7e76a82da3fd0a27ce3349e61c56a974558f32

KennethWKZ:
> @KennethWKZ It looks like we are still getting test errors here.

Fixed the unit tests but not sure how to test the stream function - func.on('extractError')

Need some help on this

andrewda:
Just pushed two tests for `func.on('extractError')` -- @KennethWKZ please take a look and see if those make sense.

fishcharlie:
First of all, I **really** appreciate this contribution @KennethWKZ. I know you likely spent a lot of time on this, and I always appreciate contributions from new people.

However, sadly, I'm going to have to close this PR. It pains me to do so, but when truly assessing everything, it's a decision that has to be made.

There are a few reasons for this decision which I'll cover below:

1. Dynamoose v3 is in progress, and uses the new AWS-SDK v3. So the odds of this behavior completely changing for that, are incredibly high. So merging this PR in, would require work and time to implement for v3.
2. I'm not convinced that this is truly a Dynamoose problem. To me this is something that AWS-SDK should fix themselves. I don't think a bandaid workaround fix like this in Dynamoose is beneficial. The correct course of action is for AWS-SDK to expose this information in their errors.
3. Even all of that being said, this feels much more like a plugin as opposed to something that should be a core part of Dynamoose. (see #851).
4. This issue is not widely reported, and solves a very one off use case.

I'm really sorry to have to close this PR. But with where Dynamoose is currently, the drawbacks to merging this in outweigh the benefits.

PR #1222, opened 2021-06-17
Fix logic of canUseIndexOfTable

andrewda:
### Summary:

This change fixes some odd behavior with local secondary indexes, and global indexes that use the same hash key as the table index. Prior to this change, if I queried tried to query a local secondary index or an index with the same hash key as the table, `canUseIndexOfTable` would return `true`, meaning the  table index would be used even when a better index exists.

This PR resolves a few of these logic issues in `canUseIndexOfTable` and adds some tests + documentation to make it more clear what is happening. Specifically, the "big" update is:

```diff
-if (rangeKeyOfTable) {
-    return Object.entries(chart).some(([fieldName]) => fieldName !== hashKeyInQuery[0]);
+if (rangeKeyOfTable && chart[rangeKeyOfTable]) {
+    return true;
}
```

The previous code basically said "if the hash key and any other key exists in the query, use the table index." This updates that logic to be "if the the hash key and the range key exist together in the query, use the table index." Otherwise, a better index may exist.

See the provided test for more details. On the current `master` branch, this test fails like so:

```diff
 {
   "ExpressionAttributeNames": {
-    "#a1": "favoriteNumber"
     "#qha": "id"
+    "#qra": "favoriteNumber"
   }
   "ExpressionAttributeValues": {
     ":qhv": {
       "N": "7"
     }
-    ":v1": {
+    ":qrv": {
       "N": "2"
     }
   }
-  "FilterExpression": "#a1 = :v1"
-  "KeyConditionExpression": "#qha = :qhv"
+  "IndexName": "myLSI"
+  "KeyConditionExpression": "#qha = :qhv AND #qra = :qrv"
   "TableName": "Cat"
 }
```

In other words, the LSI is not used for the query, despite it and the hash key being the only two parameters passed.

Please let me know of any further test cases you'd like to see, if any.

### Code Sample

```ts
export const ExampleModel: ModelDefinition = {
  name: 'example',
  schema: new Schema({
    game: { type: String, hashKey: true },
    id: { type: String, rangeKey: true },

    timestamp: { type: Number, index: { name: 'GameTimeIndex' } },
  }),
};

// ...

const response = await this.exampleModel
  .query()
  .where('game')
  .eq('test')
  .and()
  .where('timestamp')
  .ge(0)
  .sort(SortOrder.descending)
  .limit(50)
  .getRequest();
```

Before this change, the following request is made:

```ts
{
  FilterExpression: '#a1 >= :v1',
  ExpressionAttributeNames: { '#a1': 'timestamp', '#qha': 'game' },
  ExpressionAttributeValues: { ':v1': { N: '0' }, ':qhv': { S: 'test' } },
  TableName: 'example',
  Limit: 50,
  KeyConditionExpression: '#qha = :qhv',
  ScanIndexForward: false
}
```

After this change, we get this request:

```ts
{
  ExpressionAttributeNames: { '#qha': 'game', '#qra': 'timestamp' },
  ExpressionAttributeValues: { ':qhv': { S: 'test' }, ':qrv': { N: '0' } },
  TableName: 'example',
  Limit: 50,
  IndexName: 'GameTimeIndex',
  KeyConditionExpression: '#qha = :qhv AND #qra >= :qrv',
  ScanIndexForward: false
}
```

Now it correctly uses the `GameTimeIndex` -- much better!

### Other information:

Note that due to the nature of this change, it is likely to be breaking (or at least change functionality) for at least a few users. If someone were, for some reason, dependent on the query _not_ using the correct index, this would break that logic.

### Type:
- [x] Bug fix
- [ ] Feature implementation
- [x] Documentation improvement
- [x] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:
  - [x] `npm test`
  - [x] `npm run lint`


test/unit/DocumentRetriever.js
```diff
@@ -1,5 +1,6 @@
 const {expect} = require("chai");
-const DocumentRetriever = require("../../dist/DocumentRetriever");
+const rewire = require("rewire");
+const DocumentRetriever = rewire("../../dist/DocumentRetriever");
```
fishcharlie, line 3: I think we should remove the dependency on `rewire` and move the necessary functions into the `utils` folder.

fishcharlie (changes requested):
Just one suggestion. Other than that looks good.

Also. Would love to hear your thoughts on why you believe this is a breaking change. If we do come to the conclusion this is a breaking change, it'll need to be merged into the v3 branch.

andrewda:
> Also. Would love to hear your thoughts on why you believe this is a breaking change.

For the most part, I think this is a pretty non-breaking change. However, this change will inevitably change some queries (see the tests and my examples in the description). It's unlikely that anyone is actually using these queries since they're currently broken, and if someone is, this _should_ correct the queries to return the correct data. My only concern is the rare/unlikely case that someone is depending on these queries using the table index + a filter expression instead of the more appropriate index.

fishcharlie (approved):

fishcharlie:
Not going to consider this a breaking change. It seems like it'll fix an existing bug and not disrupt valid workflows. Previously this would result in an error. So it's just fixing a bug as opposed creating a breaking change.

fishcharlie:
@andrewda This was released as part of v2.8.0. However I'm getting some reports of breaking behavior. Look at #1243 and #1240. Mind reviewing those and let me know what your thoughts are? I'm not sure that it's directly a result of this work, but this work is the closest thing to the issues that are being reported.

PR #1240, opened 2021-08-17
Fix "Index can't be found for query." error for a query without an index

daria-garkusha:
### Summary:

Query with a hash key condition and a filter expression without an index should not fail with "Index can't be found for query." error message because it's a valid query in DynamoDB. 
Automatic detection of an index is a nice feature, but it should not limit the functionality of DynamoDB. So a check for an index should be removed or as another possible solution, there should be a setting to disable this behavior.

#### Model
```
const ExampleModel = dynamoose.model(
  "example", {
    "id": {
      "hashKey": true,
      "type": Number,
    },
    "name": {
      "type": String,
    },
   ...   
  });
```

#### General
```
// Should not fail
const response = await ExampleModel.query("id").eq(111).filter("name").eq("Charlie").exec()
```

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [ ] Yes
- [ ] No

- [ ] Yes
- [x] No

### Other:
  - [ ] `npm test`
  - [ ] `npm run lint`

### DIFF
--- a/lib/DocumentRetriever.ts
+++ b/lib/DocumentRetriever.ts
@@ -167,9 +167,6 @@ DocumentRetriever.prototype.getRequest = async function (this: DocumentRetriever
 		}, {});
 		if (!utils.can_use_index_of_table(this.internalSettings.model.getHashKey(), this.internalSettings.model.getRangeKey(), comparisonChart)) {
 			object.IndexName = utils.find_best_index(indexes, comparisonChart);
-			if (!object.IndexName) {
-				throw new CustomError.InvalidParameter("Index can't be found for query.");
-			}
 		}
 	}
 	function moveParameterNames (val, prefix): void {
--- a/test/unit/Query.js
+++ b/test/unit/Query.js
@@ -800,6 +800,14 @@ describe("Query", () => {
 					});
 				});
 
+				it("Should not throw error if query contains hash key and additional filter expression", () => {
+					queryPromiseResolver = () => ({"Items": []});
+					Model = dynamoose.model("Cat", new dynamoose.Schema({"id": {"type": Number, "hashKey": true}, "name": String}));
+
+					return expect(callType.func(Model.query("id").eq(111).filter("name").eq("Charlie").exec).bind(
+						Model.query("id").eq(111).filter("name").eq("Charlie"))()).to.not.be.rejectedWith("Index can't be found for query.");
+				});
+
 				it("Should throw error if no indexes exist on model", () => {
 					queryPromiseResolver = () => ({"Items": []});
 					Model = dynamoose.model("Cat", new dynamoose.Schema({"id": Number, "name": String}));


fishcharlie (changes requested):
CI failures

fishcharlie:
@daria-garkusha Did this work on v2.7.3 but is failing only on v2.8.0? This looks similar to #1243.

fishcharlie:
@daria-garkusha Got it. I wonder why that PR passed the tests then. It's weird that this is working in v2.7.3 but not in v2.8.0. I agree it has something to do with that PR that you linked likely. Just not sure why the tests didn't fail on that one but are failing with this fix.

Haven't had time to dig in a lot tho.

Would like to hear from @andrewda about this tho.

daria-garkusha:
@fishcharlie Tests fail in this PR because as a fix I removed `throw new CustomError.InvalidParameter("Index can't be found for query.");` [here](https://github.com/dynamoose/dynamoose/blob/master/lib/DocumentRetriever.ts#L171) and there is no check for a missing index now. I think those failing tests could be removed as it's probably not possible to determine if an index is really missing or a user would like to execute a query without an index. [find_best_index](https://github.com/dynamoose/dynamoose/blob/master/lib/DocumentRetriever.ts#L169) will still recommend a suitable index if it's not explicitly specified in a query but will be not so strict

fishcharlie:
@daria-garkusha 

> I think those failing tests could be removed as it's probably not possible to determine if an index is really missing or a user would like to execute a query without an index

That makes sense to me, however, I'm unsure why those tests didn't cause problems before.

---

Obviously this is a regression. And in my opinion a pretty major one. So here are my goals ranked by priority right now:

1. Get this bug fixed and get back to v2.7.3 behavior (that was working correctly).
2. Add test cases for this bug so that we don't run into this problem again.
3. Release v2.8.1 version
4. Get #1222 working again, without breaking the test created in step 2.

I saw you added a test to this PR. I'm curious what the result is, if we branched off of v2.7.3, and added in that test. Would that help to prevent this issue and #1243 in the future? And satisfy the step 2 requirement?

If so, I think the solution is to revert #1222. Add in that test, then have @andrewda redo #1222 while ensuring that it passes that test.

I'm **highly** concerned about regression at this point. I don't like the idea at this point of removing those failing tests. I'd like to get back to a clean working state, add tests to prevent this issue in the future, then work forward from there.

andrewda:
Thanks for bringing this up, I'm looking into this further right now.

andrewda:
> This line of [code](https://github.com/dynamoose/dynamoose/blob/master/lib/utils/can_use_index_of_table.ts?rgh-link-date=2021-08-18T15%3A21%3A44Z#L14) will return `true` if there is a hash key and no range key in a query.

I believe this is the correct behavior. If the hash key is the only key in the query, the table index can be used.

> But then [find_best_index](https://github.com/dynamoose/dynamoose/blob/master/lib/DocumentRetriever.ts?rgh-link-date=2021-08-18T15%3A21%3A44Z#L169) will not be able to find an index, which is not required for such query

We are checking the following:

```ts
if (!utils.can_use_index_of_table(this.internalSettings.model.getHashKey(), this.internalSettings.model.getRangeKey(), comparisonChart))
```

So if `can_use_index_of_table` returns `true`, then the `find_best_index` code and accompanying `InvalidParameter` throw will not be run. Removing the throw is a temporary work-around, but will lead to more query errors when an index actually can't be found by Dynamoose.

---

One thing that could certainly cause this is if the `comparisonChart` includes filters, but I believe that would have been an issue before #1222 as well. The only way I can see the following query throwing "Index can't be found for query" is if the filter is included in `comparisonChart` (since otherwise that part of the code should be skipped).

```ts
const response = await ExampleModel.query("id").eq(111).filter("name").eq("Charlie").exec()
```

I'm going to do some more testing around this, but if that really is the issue it should be a pretty straight-forward fix.

fishcharlie:
@andrewda Thanks for your investigation so far. Tagging @markelrod due to #1243.

daria-garkusha:
To be honest, reverting [#1222 ](https://github.com/dynamoose/dynamoose/pull/1222) would be also not the best option because it fixes another issue with a local secondary index... And queries, which are in failing tests, should actually fail not because an index is missing, but because they don't contain a hash key (for the table or an index). Can we maybe check it additionally and raise an error in such a case? It will fix failing tests and at the same time will fulfill the test, which I added

fishcharlie:
@daria-garkusha I'm not saying that it won't be fixed ever. I'm just saying that bug has existed for a long time. Handling regressions is the **top** priority currently. So even if it means reverting a bug fix to solve a regression, I'm **all** for that.

andrewda:
Looks like `filter`s are being included in the chart:

```ts
const response = await ExampleModel.query("id").eq(111).filter("name").eq("Charlie").exec()

// comparisonChart = { id: { type: 'EQ' }, name: { type: 'EQ' } }
```

Additionally, I just tested on v2.7.3 and the test still does not pass, so the issue is present even before #1222.

Working on a fix now, and will put up a PR when I find something.

daria-garkusha:
@fishcharlie I understand that it will be also fixed later. I was just considering if it's possible somehow to keep that fix and solve a current problem. But from the priority, I fully agree that regression has a top priority and should be fixed as soon as possible

fishcharlie:
@daria-garkusha Yeah I'm not opposed to it being fully fixed all at once. However, I also want to limit risk. v2.8.0 has been one of the releases with highest regression for Dynamoose that I can remember in a while. Which isn't good. Just want to reduce the risk of breaking people's integrations.

We'll see what @andrewda is able to find. Thanks for your work on this @andrewda!!

andrewda:
@fishcharlie @daria-garkusha I've put up https://github.com/dynamoose/dynamoose/pull/1244 with a potential fix. I've copied over the test you added in this PR and it is passing.

fishcharlie:
Superseded by #1244

Issue #1243, opened 2021-08-18
[BUG]


markelrod:
### Summary:

Unable to perform query using primary key in 2.8. Works in 2.7.3.

#### Schema
```
// Code here

const BroadcastSchema = new dynamoose.Schema(
  {
    user_id: {
      hashKey: true,
      type: String,
      required: true,
    },
    timestamp: {
      type: Date,
      rangeKey: true,
    },
    session_id: {
      type: String,
      required: true,
      index: {
        name: "session-index",
        global: true,
        throughput: { read: 10, write: 10 },
      },
    },
```

#### Model
```
// Code here
```

#### General
```
// Code here
let broadcasts = await Broadcast.query("user_id")
    .eq(user_id)
    .filter("recording")
    .exists()
    .sort("descending")
    .all(100)
    .exec();
```

### Current output and behavior (including stack trace):

InvalidParameter: Index can't be found for query.
--
at Query.DocumentRetriever.getRequest (/usr/src/app/node_modules/dynamoose/lib/DocumentRetriever.ts:171:11)
at runMicrotasks (<anonymous>)
at processTicksAndRejections (internal/process/task_queues.js:97:5)
at /usr/src/app/node_modules/dynamoose/lib/DocumentRetriever.ts:114:20

### Expected output and behavior:

List of records

### Environment:

Operating System:
Operating System Version:
Node.js version (`node -v`):
NPM version: (`npm -v`):
Dynamoose version:

### Other:
- [X ] I have read through the Dynamoose documentation before posting this issue
- [X ] I have searched through the GitHub issues (including closed issues) and pull requests to ensure this issue has not already been raised before
- [ X] I have searched the internet and Stack Overflow to ensure this issue hasn't been raised or answered before
- [X ] I have tested the code provided and am confident it doesn't work as intended
- [ X] I have filled out all fields above
- [X ] I am running the latest version of Dynamoose (HAVE TO REVERT BC OF ISSUE)

fishcharlie:
@markelrod Wanna test the #1240 PR and see if that solves your issue?

markelrod:
Happy to check it. I need to add it to my package.json so Docker build pulls it in. Never tried to reference a PR before. 

I think I use the branch. Is what I need to do?

[...]

PR #1244, opened 2021-08-18
Fix queries with hash key and filter

andrewda:
See:
#1240
#1243

### Summary:

Queries using the table hash key and a filter are currently failing due to how we are trying to find the best index. This resolves this issue by performing an additional check to ensure the table index is being used as a last resort, if possible (i.e. if there is an "EQ" operation on the hash key).

#### Schema
```ts
const ExampleModel = dynamoose.model(
  "example", {
    "id": {
      "hashKey": true,
      "type": Number,
    },
    "name": {
      "type": String,
    },
   ...   
  });
```

#### Model
```ts
// Should not fail
const response = await ExampleModel.query("id").eq(111).filter("name").eq("Charlie").exec()
```

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [x] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


fishcharlie:
This looks **really** good at first glance. Wish that @markelrod was able to test this PR, but I'm not sure how to get it working with TypeScript and pointing to a branch. 😢 

@daria-garkusha are you able to review this PR as well?

andrewda:
You should be able to add this to your `package.json` to use this commit:

```
    "dynamoose": "github:andrewda/dynamoose#42767a019c24902984e34387fa903e936b3ae2f1",
```

markelrod:
Let me know when 2.8.1 is ready and I'll test it right away!

test/unit/Query.js
```diff
@@ -800,6 +800,14 @@ describe("Query", () => {
 					});
 				});
 
+				it("Should not throw error if query contains hash key and additional filter expression", () => {
+					queryPromiseResolver = () => ({"Items": []});
+					Model = dynamoose.model("Cat", new dynamoose.Schema({"id": {"type": Number, "hashKey": true}, "name": String}));
+
+					return expect(callType.func(Model.query("id").eq(111).filter("name").eq("Charlie").exec).bind(
+						Model.query("id").eq(111).filter("name").eq("Charlie"))()).to.not.be.rejectedWith("Index can't be found for query.");
```
fishcharlie, line 808: I think maybe we should change this to be a truthy value as opposed to just verifying that it's not rejected with that specific error. Probably checking to ensure the result is an empty array. That way if it returns a different error this test won't pass.

lib/utils/find_best_index.ts
```diff
@@ -18,5 +23,11 @@ export default function (indexes: ModelIndexes, comparisonChart: ConditionStorag
...
 
 	const index = validIndexes.find((index) => comparisonChart[index._rangeKey]) || validIndexes[0];
-	return index?.IndexName ?? null;
+
+	// If no index is found and hash key exists, use table index
+	if (!index && comparisonChart[tableHashKey]?.type === "EQ") {
+		return {"tableIndex": true};
+	}
+
+	return {"tableIndex": false, "indexName": index?.IndexName ?? null};
```
fishcharlie, line 32: One question. Without this change, is at least 1 of the tests you modified/added failing?

fishcharlie (commented):
Looks good. Couple of questions. Would also like to have @daria-garkusha review if possible before merging.

test/unit/Query.js
andrewda: Agreed, we should check the output and the throw check will be implied.

lib/utils/find_best_index.ts
andrewda: Yep! Anything with `tableHashKey` should fail without this change.

https://github.com/andrewda/dynamoose/blob/42767a019c24902984e34387fa903e936b3ae2f1/test/unit/utils/find_best_index.js#L27-L40

andrewda:
Alright, I've made another update which restructures how we decide to use the table index. First of all, the `ModelIndexes` interface now includes `TableIndex`, which is the index of the table. This exposes the table hash and range keys in `find_best_index`.

1. Our first choice for index is one with both hash key and range key defined.
2. If no index has both hash and range keys defined in the query, then our next choice is the table index (assuming its hash key is present).
3. If we still haven't found an index, then we use the first index with a hash key defined in the query.
4. If still no index, throw an `Index can't be found for query` error.

This also allows us to remove the `can_use_index_of_table` util, since `find_best_index` now also takes the table index into account. This centralizes the logic a little more, making testing more effective and straight forward.

andrewda:
Would definitely love someone who was having the issue previously to test with the latest commit 340ec36f01346e37f5d7b678f91c644f16c5e7af.

```
    "dynamoose": "github:andrewda/dynamoose#340ec36f01346e37f5d7b678f91c644f16c5e7af",
```

daria-garkusha:
@fishcharlie @andrewda I checked this PR, and it works fine for me. The new implementation of `find_best_index` is definitely more precise than before and should cover all cases.

lib/Schema.ts
```diff
@@ -336,11 +336,17 @@ export class Schema {
...
 			AttributeDefinitions,
-			KeySchema,
-			...await this.getIndexes(model)
+			KeySchema
 		};
+
+		const {GlobalSecondaryIndexes, LocalSecondaryIndexes} = await this.getIndexes(model);
+
+		if (GlobalSecondaryIndexes) response.GlobalSecondaryIndexes = GlobalSecondaryIndexes;
+		if (LocalSecondaryIndexes) response.LocalSecondaryIndexes = LocalSecondaryIndexes;
```
fishcharlie, line 347:
```suggestion
		if (GlobalSecondaryIndexes) {
			response.GlobalSecondaryIndexes = GlobalSecondaryIndexes;
		}
		if (LocalSecondaryIndexes) {
			response.LocalSecondaryIndexes = LocalSecondaryIndexes;
		}
```

fishcharlie (approved):

PR #1315, opened 2021-12-09
Remove dynamic require of ModelStore in Model.ts

avsmithy:
### Summary:

We encountered a strange bug which is hard to pin down, but
this particular commit has fixed it. My current hypothesis:

The instanceof check in ModelStore.ts is not operating correctly in
our environment. I suspect the inline require of ModelStore in Model.ts
is causing multiple instances of the ModelStore class to be loaded.

I believe our node runtime is loading a new version of the ModelStore file
every time the `require('../ModelStore`)` statement is being run. If this require was put
at the top of the file as you would do normally, then you would end up with a circular dependency.
This would explain why the require is inline.

Passing the ModelStore singleton into the Model class as a parameter
(and hence removing the circular dependency and dynamic require) appears to have solved this
issue.

I can't actually create a minimal working example as it is a huge combination of factors to do with
node/lambdas/webpack that cause this, but this change might be common sense anyway, so I thought I'd raise it!

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x ] No
- [] I'm not sure

- [x ] Yes
- [ ] No

- [ x] Yes
- [ ] No

### Other:
- [ x] I have read through and followed the Contributing Guidelines
- [ x] I have searched through the GitHub pull requests to ensure this PR has not already been submitted
- [ x] I have updated the Dynamoose documentation (if required) given the changes I made
- [ x] I have added/updated the Dynamoose test cases (if required) given the changes I made
- [ x] I have ensured the following commands are successful from the root of the project directory
  - [x ] `npm test`
  - [ x] `npm run lint`
- [ x] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoose/dynamoose/blob/main/LICENSE)
- [x ] All of my commits and commit messages are detailed, explain what changes were made, and are easy to follow and understand
- [ x] I have filled out all fields above


fishcharlie:
@avsmithy Thanks for the contribution! Few questions:

1. `We encountered a strange bug`, can you describe more about what this bug was?
2. Thanks for merging this into the v3 branch. However it raises the question, does this bug only exist on v3? Or does it exist on the v2 line as well?
3. Do you use `dynamoose.model(name)` at all to retrieve previously created models? This is what the ModelStore is used for. Just curious how widespread usage of this feature is.

I'm fine to make an exception to merge this in without tests. However, please be aware that without tests, I can't guarantee that future changes won't occur that will break this down the road. This means that patch versions, minor versions, or even major versions can occur that will break your setup. I'd highly encourage you to add tests on your end, to verify that this works as intended.

Finally, do me one last favor and change the breaking changes question in the original template to `No`.

fishcharlie:
Oh **right** as I posted that I notice that `coveralls` failed since the coverage decreased. That changes the situation of doing this without tests. Really can't merge PRs that cause a decrease in code coverage.

If you get a chance and can take a look at that I'd appreciate it.

avsmithy:
Hi @fishcharlie, thanks for engaging!

> We encountered a strange bug, can you describe more about what this bug was?

Dynamoose was throwing the error "You must pass in a Model or table name as a string", which is the error thrown on line 14 of ModelStore.

This would happen inconsistently when we ran our application. Sometimes it would start successfully and continue to run without errors. After killing the process and restarting (with absolutely no other changes to the environment) we would then see this error immediately.

We debugged this and are certain its due to the instanceof check on line 8 of ModelStore failing. Strangely, the `input` variable _was_ of type _Model_, but `input instanceof Model` evaluated to false. Hence why I believe it's something to do with multiple instances of either Model or ModelStore being loaded by webpack/node.

> Do you use dynamoose.model(name) at all to retrieve previously created models? This is what the ModelStore is used for. Just curious how widespread usage of this feature is.

Yes, we use `dynamoose.model(tableName, schema, ...)` to create the model, and use the return value of that function to access to model later. If I recall correctly, the error was thrown during the `ModelStore(this)` call at the end of the `Model constructor.`.

> Thanks for merging this into the v3 branch. However it raises the question, does this bug only exist on v3? Or does it exist on the v2 line as well?

We haven't been using v2 for a while, and this error only cropped up recently, so I can't say for sure. I can't even figure out which commit introduced this on our side, as it started off very infrequently and then became more common.

I understand you can't guarantee this won't happen again without tests, but since I'm unable to reproduce a simplified version of the issue, I wouldn't know where to start writing a test for this.

One more thing, this only ever happened when webpack was in development mode. The production bundle was always fine.

If it breaks again, we will see it in development and not in production, so it's an acceptable risk to us.

In the future we will be bundling all of our node code via webpack and stop using nodeExternals, at which point I consider the risk completely mitigated on our side.

> Oh right as I posted that I notice that coveralls failed since the coverage decreased. That changes the situation of doing this without tests. Really can't merge PRs that cause a decrease in code coverage .

The coverage decreased by **-0.0003%**, which I guess probably due to the one line of code being removed in ModelStore! 😄  Is this ok?

> Finally, do me one last favor and change the breaking changes question in the original template to No.

Done

Cheers!

fishcharlie:
@avsmithy 

> we use `dynamoose.model(tableName, schema, ...)` to create the model

Got it. Do you ever use `dynamoose.model(name)` tho? Without passing in the schema or other options. That will return your already created model so you can use it in other places. For example:

```js
const User = dynamoose.model("User", {id: String, name: String});
const UserB = dynamoose.model("User");

// User & UserB will be identical
```

Question is if you ever do something like `UserB`?

> The coverage decreased by -0.0003%, which I guess probably due to the one line of code being removed in ModelStore! 😄 Is this ok?

Sadly no. In v2 the code coverage is 100% across the board. I know it has decreased a little bit in v3. But before v3 launch, I will be getting that back up to 100%. Since it's all my code and changes I made, I know I'll be able to get it back up to 100% very easily.

I'd be happy to take a look later to see if I can spot something to fix this, and maybe add a test, but this does need to be done before this gets merged.

---

As for everything else that I didn't directly reply to, it sounds great, and thanks for the information.

avsmithy:
> Got it. Do you ever use dynamoose.model(name) tho? Without passing in the schema or other options. That will return your already created model so you can use it in other places. 

No, we have an abstraction layer over dynamoose which calls `dynamoose.model(name, schema)` and then stores the return value. Every call after that uses the stored returned value. 

> I'd be happy to take a look later to see if I can spot something to fix this, and maybe add a test, but this does need to be done before this gets merged.

I understand - I'm not sure what tests to write to get the coverage back up though, I'd really appreciate if you could point me in the right direction? I'm than happy to write them.

Cheers!

fishcharlie:
I can't figure out the test coverage discrepancy here. I'm gonna try to prioritize getting code coverage back to 100% on the v3 branch, then merge that into here and see what the outcome is.

avsmithy:
Thanks @fishcharlie!

PR #1339, opened 2022-01-14
Update Model.md

Drarig29:
I think you should talk about TypeScript interfaces here, not classes.

The snippet code has errors:

> Property 'id' has no initializer and is not definitely assigned in the constructor

image

### DIFF
--- a/docs/docs/guide/Model.md
+++ b/docs/docs/guide/Model.md
@@ -18,14 +18,14 @@ const Cat = dynamoose.model("Cat", new dynamoose.Schema({"name": String}));
 const Cat = dynamoose.model("Cat", new dynamoose.Schema({"name": String}), {"create": false});
 ```
 
-An optional TypeScript class which extends `Document` can be provided right before the function bracket. This provides type checking when using operations like `Model.create()`.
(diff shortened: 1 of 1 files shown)


fishcharlie (changes requested):
Take a look at this type test:

https://github.com/dynamoose/dynamoose/blob/9f5b240aa383a1cfce858398ec7b0b49c200bd2d/test/types/Model.ts#L66-L91

This matches almost identically what the documentation says currently (without this PR).

So I'm curious if you could give a little bit more explanation and reasoning behind this proposed change. What error are you getting with the current code? What does your tsconfig file look like?

Drarig29:
Hey, sorry for the delay! For my use I thought it would better to use TS interfaces because I didn't want to use the null assertion operator.

But I ended up using a class extending `Document` and implementing a TS interface instead.

Sorry for the loss of time!

fishcharlie:
@Drarig29 I agree. As far as I'm aware tho, this is a limitation of TypeScript.

If you want to dig into it a bit more and see if there is a workaround to use interfaces here and submit a PR that'd be awesome.

But no pressure. I'm glad you got it resolved!

PR #1398, opened 2022-05-07
Updating appId and apiKey for Algolia

fishcharlie:


fishcharlie:
CI failing, but merging anyway because I think this will fix the problem.

PR #1421, opened 2022-06-17
Export all interfaces/classes from Schema definitions.

dipack95:
### Summary:

Export interfaces / classes from the Schema definition for consumption by users of Dyanmoose.

### Type (select 1):
- [ ] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [x] Something not listed here

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No

### Other:

### DIFF
--- a/packages/dynamoose/lib/Schema.ts
+++ b/packages/dynamoose/lib/Schema.ts
@@ -782,7 +782,7 @@ export interface SchemaGetAttributeSettingValue {
 	typeIndexOptionMap?: any;
 }
 
-export interface SchemaInternalProperties {
+interface SchemaInternalProperties {
 	schemaObject: SchemaDefinition;
 	settings: SchemaSettings;
 	getMapSettingValuesForKey: (key: string, settingNames?: string[]) => string[];
--- a/packages/dynamoose/lib/Schema.ts
+++ b/packages/dynamoose/lib/Schema.ts
@@ -35,7 +35,7 @@ export interface DynamoDBTypeResult {
 	set?: DynamoDBSetTypeResult;
 }
 
-interface DynamoDBTypeCreationObject {
+export interface DynamoDBTypeCreationObject {
 	name: string;
 	dynamicName?: ((typeSettings?: AttributeDefinitionTypeSettings) => string);
 	dynamodbType?: DynamoDBAttributeType | DynamoDBAttributeType[] | DynamoDBType | ((typeSettings: AttributeDefinitionTypeSettings) => string | string[]);
@@ -45,7 +45,7 @@ interface DynamoDBTypeCreationObject {
 	customType?: {functions: (typeSettings: AttributeDefinitionTypeSettings) => {toDynamo?: (val: ValueType) => ValueType; fromDynamo?: (val: ValueType) => ValueType; isOfType: (val: ValueType, type: "toDynamo" | "fromDynamo") => boolean}};
 }
 
-class DynamoDBType implements DynamoDBTypeCreationObject {
+export class DynamoDBType implements DynamoDBTypeCreationObject {
 	// TODO: since the code below will always be the exact same as DynamoDBTypeCreationObject we should see if there is a way to make it more DRY and not repeat it
 	name: string;
 	dynamicName?: ((typeSettings?: AttributeDefinitionTypeSettings) => string);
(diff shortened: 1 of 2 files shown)


packages/dynamoose/lib/Schema.ts
```diff
@@ -774,15 +774,15 @@ interface AttributeDefinition {
...
 	unknownAttributeAllowed: boolean;
 }
-interface SchemaGetAttributeSettingValue {
+export interface SchemaGetAttributeSettingValue {
 	returnFunction: boolean;
 	typeIndexOptionMap?: any;
 }
 
-interface SchemaInternalProperties {
+export interface SchemaInternalProperties {
```
fishcharlie, line 785: Why should something named `SchemaInternalProperties` be exported? That makes **ZERO** sense to me. Why are you needing this to be exported?

fishcharlie (changes requested):
Couple of things:

1. This should be documented in the `PENDING_CHANGELOG.md` file
2. We need tests for this. See `packages/dynamoose/test/types`
3. Why is this necessary? What code is the client running that is failing currently, but will succeed after this PR.

packages/dynamoose/lib/Schema.ts
dipack95: Yeah that's fair, I can remove this export.

dipack95:
@fishcharlie 

Something like:

```typescript
const schema = new dynamoose.Schema({ id: { type: String, required: true, hashKey: true } });

type ExtractType<T extends AttributeDefinition> = T['type'] extends String ? string : unknown;

type ExtractSchema<T extends SchemaDefinition> = { [k: string]: ExtractType<T[k]> };

// The following `SchemaInterface` would ensure that any object cast to it,
// would have the `id` property, which would need to be a String.
interface SchemaInterface extends ExtractSchema<typeof schema['Schema']['schemaObject'] {}
```

I'm still not a 100% sure if the example above will work, but this was where my thoughts were tending towards. For context, we currently use Sequelize for a related project, and we extract and enforce Typescript types from the Sequelize attribute definitions using a similar methodology.

dipack95:
> We need tests for this. See packages/dynamoose/test/types

@fishcharlie I'm not sure what tests I can add for interface exports, tbh. Can you elaborate?

fishcharlie:
@andrewda I'd love your thoughts on this PR. I'm leaning towards rejecting and closing it. It kinda encourages abuse of our rule about [not using undocumented pieces of Dynamoose](https://dynamoosejs.com/other/FAQ#can-i-use-an-undocumented-property-class-method-or-function-in-dynamoose). To me, these interfaces are not solidified. I haven't thought enough to ensure they are ready for public use. They are really meant to be internal at this point. Additionally TypeScript support is still in Beta, which adds to this as well.

@dipack95 I'd suggest using a fork for now. It's really recommended against what you are doing as mentioned in that FAQ post above. I don't want to encourage use of internal components that could change anytime. But I don't have a better solution for you right now other than to use a fork.

andrewda:
Yep, I agree. This is exposing some internals that we probably don't want to commit to not breaking in future re-rewrites. Additionally, similar interfaces in projects like mongoose are not exported, presumably for similar reasons.

https://github.com/Automattic/mongoose/blob/3a4ba27a75622a9fc370683744b8822f3f9dbfdf/types/query.d.ts#L93

It seems like you're roughly trying to create an automatic conversion between your database schema and a TypeScript interface, but I'm not necessarily sure that's the best approach in the first place. Personally, I like to lean towards a more explicit approach which grants you more flexibility if your database model differs from your application logic model. Below is an example from one of my codebases using [`nestjs-dynamoose`](https://github.com/hardyscc/nestjs-dynamoose), although something similar could be achieved using Dynamoose by itself.

```ts
import { Schema } from 'dynamoose';
import { Model, ModelDefinition } from 'nestjs-dynamoose';

export interface NotificationKey {
  id: string;
}

export interface Notification extends NotificationKey {
  user: string;
  type: string;
  timestamp: number;
  message?: string;
}

export type NotificationModel = Model<Notification, NotificationKey>;

export const NotificationModel: ModelDefinition = {
  name: 'notifications',
  schema: new Schema(
    {
      user: {
        type: String,
        index: { name: 'UserTimeIndex', global: true, rangeKey: 'timestamp' },
      },
      type: String,
      timestamp: Number,
      message: String,
    },
    { saveUnknown: true },
  ),
};
```

Would something like that work for you?

PR #1481, opened 2022-08-25
TypeMismatch field to be of type object, instead found type object.

halan:
### Summary:

Currently, some libs use objects with null prototype to avoid function injection. The issue is explained here: https://github.com/graphql/graphql-js/issues/484#issuecomment-250344120.

Technically there are objects with a null prototype and they need to be recognized by Dynamoose schema. Without that, we'll have to deep copy those objects recovering the prototype (or just the `constructor = Object`).

It's a bug because those types are object and Dynamoose getting confused about that and raise a confused message:
> TypeMismatch: Expected {the name of the field} to be of type object, instead found type object.

I'm offering a safe and quick solution to recognize those objects and allow them to be stored on DynamoDB as regular objects.

#### Schema
```ts
User = dynamoose.model("User", {
  "address": {
    "type": Object,
    "schema": {"country": String}
  }
});
```

#### General
```ts
user = new User({
  "address": Object.assign(Object.create(null), {"country": "world"})
});
// It will raise the weird Error: 
// TypeMismatch: Expected address to be of type object, instead found type object.

```

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [x] Test added to report bug 

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/packages/dynamoose/lib/Schema.ts
+++ b/packages/dynamoose/lib/Schema.ts
@@ -147,7 +147,7 @@ const attributeTypesMain: DynamoDBType[] = ((): DynamoDBType[] => {
 		new DynamoDBType({"name": "Buffer", "dynamodbType": "B", "set": true, "jsType": Buffer}),
 		booleanType,
 		new DynamoDBType({"name": "Array", "dynamodbType": "L", "jsType": {"func": Array.isArray}, "nestedType": true}),
-		new DynamoDBType({"name": "Object", "dynamodbType": "M", "jsType": {"func": (val): boolean => Boolean(val) && Object.prototype.toString.call(val) === '[object Object]'}, "nestedType": true}),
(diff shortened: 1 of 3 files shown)


fishcharlie (commented):
@halan Can you please check to ensure we have a unit test that ensures that it fails if you try to use a class instance as an object type?

halan:
> @halan Can you please check to ensure we have a unit test that ensures that it fails if you try to use a class instance as an object type?

Probably not, all tests pass but my validation doesn't distinguish plain objects from class instances (yet).

I can add that test and improve the check. We can use `val.constructor === Object || val.constructor === undefined`, it can distinguish from classes instances. There is some ideas here https://github.com/jonschlinkert/is-plain-object/blob/master/is-plain-object.js. Tomorrow I add the test and improve the check.

fishcharlie (changes requested):
Need to fix lint errors

halan:
Done

fishcharlie (approved):

Issue #1260, opened 2021-09-07
[BUG] Slow scan operations


Josefnm:
### Summary:

I'm using dynamoose for my project, and overall it works well. However scan operations are several times slower than the same operations using AWS DocumentClient. 

#### Schema
It's the same with all my different schemas. Here is an example of the simplest one. None of them use Buffer type.
```
const userSchema = new dynamoose.Schema(
  {
    id: {
      type: String,
      required: true,
      hashKey: true,
    },
    phoneNumber: {
      type: String,
      index: {
        project: true,
        global: true,
        name: 'UserPhoneNumberIndex',
      },
    },
    name: String,
    status: String,
    email: String,
    // a few more fields with strings, numbers and booleans
  },
  { timestamps: true }
)
```

#### Model
```
const model = dynamoose.model('myTable', mySchema, {
    create: false,
    waitForActive: false,
  })
```

#### General
```
// dynamoose
MyModel.scan().all().exec()

// DocumentClient
export const scanTable = async <T = unknown>(
  table: string,
  extraParams?: Partial<AWS.DynamoDB.DocumentClient.ScanInput>
): Promise<T[]> => {
  const params: AWS.DynamoDB.DocumentClient.ScanInput = {
    TableName: `${constants.ENV}-${table}`,
    ...extraParams,
  }

  const scanResults = []
  let response: PromiseResult<
    AWS.DynamoDB.DocumentClient.ScanOutput,
    AWS.AWSError
  >
  do {
    // eslint-disable-next-line no-await-in-loop
    response = await getDocumentClient().scan(params).promise()
    scanResults.push(...response.Items)
    params.ExclusiveStartKey = response.LastEvaluatedKey
  } while (typeof response.LastEvaluatedKey !== 'undefined')

  return <T[]>scanResults
}
```

### Current output and behavior (including stack trace):
Example 1:
Scanning ~450 items that are ~1 kb each (according to aws dynamodb console).
Using aws DocumentClient and scanning all: ~400 ms
Using dynamoose scan all: ~4500 ms

Example 2:
Scanning ~23000 items that are ~230 bytes each.
Using aws DocumentClient and scanning all: ~6000 ms
Using dynamoose scan all: ~24000 ms

### Expected output and behavior:

Somewhat similar performance

### Environment:

Operating System: Amazon Linux
Operating System Version: 2
Node.js version (`node -v`): v14.x
NPM version: (`npm -v`): 7.18.1
Dynamoose version: 2.8.1

[...]

PR #1572, opened 2023-02-08
Add skip conform to schema

vasily-polonsky:
### Summary:
The library shows poor performance when you need to query a large number of items from the database, especially if they have a nested structure. This is especially true for serverless, where the environment is not a cluster/EC2 instance and the CPU is usually limited. For example, during development on a local machine the operation of scanning 250 items with dynamoose takes about 3 seconds, in a lambda with 0.1 vCPU it takes over 30 seconds. This makes it not possible to use on production. Using aws-sdk it takes 600ms for the same requests in both environments (local and serverless). 

The problem is that dynamoose consumes huge CPU resources during the `confromToSchema` method call, so the solution I propose is to make it possible to skip this method call during large queries where a large number of items must be returned. Mongoose has a solution to this problem with [lean](https://mongoosejs.com/docs/tutorials/lean.html). `Lean` works a little differently, but the idea is the same - make it possible to handle a large number of items with less performance degradation. Meanwhile, I think the `confromToSchema` method itself needs to be refactored.

I haven't written documentation for the `skipConformToSchema` method, because I think we should first agree on the name and whether it's appropriate to introduce this method.

### GitHub linked issue:
https://github.com/dynamoose/dynamoose/issues/1260
https://stackoverflow.com/questions/74356120/slow-on-fetching-records-from-dynamodb-table

#----

### Type (select 1):
- [ ] Bug fix
- [ X ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [ X ] No
- [ ] I'm not sure

- [ X ] Yes
- [ ] No

- [ X ] Yes
- [ ] No

### Other:
- [ X ] I have read through and followed the Contributing Guidelines
- [ X ] I have searched through the GitHub pull requests to ensure this PR has not already been submitted
- [ X ] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoose/dynamoose/blob/main/LICENSE)
- [ X ] All of my commits and commit messages are detailed, explain what changes were made, and are easy to follow and understand

### DIFF
--- a/packages/dynamoose/lib/ItemRetriever.ts
+++ b/packages/dynamoose/lib/ItemRetriever.ts
@@ -38,6 +38,7 @@ interface ItemRetrieverInternalProperties {
 		index?: string;
 		startAt?: ObjectType;
 		limit?: number;
+		skipConformToSchema?: boolean
 	}
 }
 
@@ -46,6 +47,7 @@ interface ItemRetrieverInternalProperties {
 abstract class ItemRetriever extends InternalPropertiesClass<ItemRetrieverInternalProperties> {
 	getRequest: (this: ItemRetriever) => Promise<any>;
 	all: (this: ItemRetriever, delay?: number, max?: number) => this;
+	skipConformToSchema: (this: ItemRetriever) => this;
 	limit: (this: ItemRetriever, value: number) => this;
 	startAt: (this: ItemRetriever, value: ObjectType) => this;
 	attributes: (this: ItemRetriever, value: string[]) => this;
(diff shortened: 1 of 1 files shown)


vasily-polonsky:

nopol10:
Hi, any chance for this or something similar to get merged soon? 🙂
The >5-10 times speed improvement is sorely needed as dynamoose becomes unusable in parts of my application 😢
I'm open to helping out in any way if needed

fishcharlie:
@nopol10 & @vasily-polonsky I'm debating this one a lot. I get the performance benefit. But I then question what the value is in Dynamoose. The whole idea of Dynamoose is to have strict schemas for your DynamoDB data.

I'm curious if either of you could explain why with this PR you still choose Dynamoose over numerous other options out there? It seems like this is getting rid of a huge benefit and purpose of Dynamoose. So I'm curious about your thoughts and how you are viewing that.

Regardless, this PR needs unit tests.

vasily-polonsky:
Hi @fishcharlie 
I understand your concerns, but there is another side to it. When I use Dynamoose to work with DDB tables, I'm sure the data there is correct, and Dynamoose works really well, if you need to fetch one record or say 20, the performance issue is not critical. But sometimes I need to fetch 300-500 records (sometimes it's actually a whole table and I need to do some aggregation or calculations) and this is where problems arise. If I know for sure that the data is correct, can I have a flag to skip this check to improve my response time? This issue is really critical for serverless environments, where the vCPU is usually lower than in a kubernetes pod.

nopol10:
Yup, as @vasily-polonsky says, there are cases where I am certain that the schema will be correct so I will not be worried about non-conformance. 

In addition, dynamoose provides an API that makes dynamodb related code easier to write and maintain so I'll like to have retain that if possible with a small code change rather than rewriting the query to use the dynamodb ask directly.

ThienNDDi:
hi, @fishcharlie it has been a while since this PR is opened. Sadly, i think i got this issue too.
I used scan a table having more than 200 record. Immediately, any request to that table spent nearly 25s.
I hope this will be merge soon.

fishcharlie:
This can't be merged until unit tests are added.

jimmyn:
I'm facing the same issue, locally it takes 2–3 seconds to fetch the batch of 500 records, it lambda it is not usable at all

Issue #1432, opened 2022-07-06
[FEATURE] Make explicitly choose "primary index"(table index) with `using` method


goonerholic:
### Summary:
When filter condition is set to the property which is a hash key of some secondary index, the index query uses automatically set to that secondary index, instead of the table index(primary key)

#### Schema
```typescript
const schema = new dynamoose.Schema({
  PK: {
    type: String,
    hashKey: true
  },
  SK: {
    type: String,
    rangeKey: true,
  },
  type: {
    type: String,
    index: {
      name: "global-index",
      global: true,
      rangeKey: "otherProp"
    }
  }
})
```

#### Model
```typescript
const model = dynamoose.model(Table, schema)
```

#### General

Let's say I would like to query with condition below.
```typescript
const queryResult = await model.query("PK").eq("hello").where("SK").beginsWith("world").filter("type").eq("wow").exec();
```

What I really intended is like this
```json
{
    "FilterExpression": "#qha = :qhv",
    "ExpressionAttributeNames": {
        "#a0": "PK",
        "#qha": "type",
        "#qra": "SK"
    },
    "ExpressionAttributeValues": {
        ":v0": {
            "S": "hello"
        },
        ":qhv": {
            "S": "wow"
        },
        ":qrv": {
            "S": "world"
        }
    },
    "KeyConditionExpression": "#a0 = :v0 AND begins_with (#qra, :qrv)"
}
```

But the actual result seems like this
```json
{
    "FilterExpression": "#a0 = :v0",
    "ExpressionAttributeNames": {
        "#a0": "PK",
        "#qha": "type",
        "#qra": "SK"
    },
    "ExpressionAttributeValues": {
        ":v0": {
            "S": "hello"
        },
        ":qhv": {
            "S": "wow"
        },
        ":qrv": {
            "S": "world"
        }
    },
    "IndexName": "global-index",
    "KeyConditionExpression": "#qha = :qhv AND begins_with (#qra, :qrv)"
}
```

It seems that dynamoose automatically choose optimal index from the condition. With `using` method, I can choose which index I'm gonna use but can't pick the `primary key`(or `table index`).

I can walkround this with using `contain` condition instead of `eq` condition, because `find_best_index` only check for the `eq` condition to decide which index to use.

But It would be great if I can set "primary key" explicitly with `query.using` method or just make "primary key" prior to the other indices.

### Environment:

Operating System: Apple silicon M1
Operating System Version: Mac OS 12
[...]

PR #1578, opened 2023-03-10
Make .query() respect .using() index

yuliswe:
### Summary:

Proposes a solution to #1432 by an additional usage of the `.using(indexName)` API.

#### Before the change:

`.using(indexName: string)`

#### After the change:

`.using(indexName: string | {auto: boolean})`

#### Behavior before the change:

When not specifying `.using(indexName: string)`, `.query()` automatically selects an index.
When specifying an index name (string) with `.using(indexName)`, `.query()` and all other methods use that index name.

#### Behavior after the change: 

When not specifying `.using()`: Same as before.
When specifying an index name (string): Same as before
When specifying `.query({auto: true})`:  Same as the current behavior when .using() is not present
When specifying `.query({auto: false})`:  `.query()` uses the main primary / range key instead of auto-selecting an index.

####

#### Schema

```
SomeModel.query(..).exec()
// behavior unchanged

SomeModel.query(..).using("someIndex").exec()
// behavior unchanged

SomeModel.query(..).using({auto: true})
// same as SomeModel.query(..)

SomeModel.query(..).using({auto: false}).exec()
// disables auto selecting index
```

### Type (select 1):

Feature implementation

No

Yes

Yes

### DIFF
--- a/packages/dynamoose/lib/ItemRetriever.ts
+++ b/packages/dynamoose/lib/ItemRetriever.ts
@@ -35,7 +35,7 @@ interface ItemRetrieverInternalProperties {
 		attributes?: string[];
 		count?: number;
 		consistent?: boolean;
-		index?: string;
+		index?: string | { auto: boolean };
 		startAt?: ObjectType;
 		limit?: number;
 	}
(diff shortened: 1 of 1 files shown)


ylilarry:

fishcharlie (changes requested):
@ylilarry Thanks for this! Few thoughts:

1. This PR basically makes it so if you pass in `{auto: false}` it will use the table itself and not include `IndexName` in the request. Is that correct?
2. If so, I wonder if `auto` is the best term. It feels like `.using({"table": true})` or something like that would be better. Or even `.using(dynamoose.Table)` to represent that you want to query the table itself. Since there isn't really a true or false option here. It's an option you want to use. There isn't really a true/false nature there.
3. We also need unit tests & documentation here.

PR #1587, opened 2023-04-14
move '@aws-sdk' modules from 'dependencies' to 'peerDependencies'

KanghoonYi:
### Summary:
I moved '@aws-sdk/client-dynamodb', '@aws-sdk/util-dynamodb' to peerDependecies.
On AWS Lambda, '@aws-sdk/*' modules is already embedded([Reference](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html)).
Even though i exclude that, when i installed 'dynamoose', '@aws-sdk/client-dynamodb' is included to bundled package.

### Type (select 1):
- [ ] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [x] Something not listed here

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/packages/dynamoose/package-lock.json
+++ b/packages/dynamoose/package-lock.json
@@ -19,13 +19,15 @@
       ],
       "license": "Unlicense",
       "dependencies": {
-        "@aws-sdk/client-dynamodb": "^3.142.0",
-        "@aws-sdk/util-dynamodb": "^3.142.0",
-        "dynamoose-utils": "^3.0.0-beta.6",
+        "dynamoose-utils": "^3.1.0",
         "js-object-utilities": "^2.1.0"
       },
       "devDependencies": {
-        "dynamoose-logger": "^3.0.0-beta.6"
+        "dynamoose-logger": "^3.1.0"
+      },
+      "peerDependencies": {
+        "@aws-sdk/client-dynamodb": "^3.142.0",
+        "@aws-sdk/util-dynamodb": "^3.142.0"
       }
     },
     "node_modules/@aws-crypto/ie11-detection": {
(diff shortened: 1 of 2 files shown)


KanghoonYi:

fishcharlie:
@KanghoonYi Thanks for this. I agree this is something we need to consider. However, I question if this is a breaking change or not. In npm versions prior to v7 there was a few versions that did not automatically install peerDependencies. Which requires an extra install step. Pretty sure since Dynamoose supports Node.js >= v12, there are a few npm versions we have to think about.

I'm not opposed to doing this. But it feels like a breaking change to me if users are running those older versions of Node.js.

What do you think?

KanghoonYi:
@fishcharlie 
I see...
You are right. This is breaking change.
How about use 'engines' property on package.json?
```json
{
   "engines": {
     "npm": ">= 7.0.0"
   }
}
```
If npm version lower than 7, users get error that require 'npm' upgrade.
NPM upgrade("npm install -g npm@7") is possible on node 12.
[Reference](https://github.com/pixohq/babystory-data-model/pull/15)

> @KanghoonYi Thanks for this. I agree this is something we need to consider. However, I question if this is a breaking change or not. In npm versions prior to v7 there was a few versions that did not automatically install peerDependencies. Which requires an extra install step. Pretty sure since Dynamoose supports Node.js >= v12, there are a few npm versions we have to think about.
> 
> I'm not opposed to doing this. But it feels like a breaking change to me if users are running those older versions of Node.js.
> 
> What do you think?

fishcharlie:
Not a bad idea. I think that'd still be a breaking change tho.

PR #1598, opened 2023-04-30
pendingTasks error handling 

jinseok0:
### Summary:

when I put in the wrong AWS_SECRET_ACCESS_KEY, I get an error in the console but can't get out of the await promise.
so I checked the code and saw that there is no exception handling for `await setupFlowPromise` in `/lib/Table/index.ts`.

screenshot1

#### General
```typescript
const dynamoose = require("dynamoose");

(async () => {
  console.log(1);
  const Cat = dynamoose.model("Cat", new dynamoose.Schema({ name: String }));
  try {
    console.log(2);
    await Cat.create({
      name: "cat",
    });
    console.log(3);
  } catch (error) {
    console.log(4);
  }
  console.log(5);
})();
```

#### Expectation

screenshot2

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [ ] No
- [x] I'm not sure

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No

### Other:

### DIFF
--- a/packages/dynamoose/lib/Table/index.ts
+++ b/packages/dynamoose/lib/Table/index.ts
@@ -150,8 +150,8 @@ export class Table extends InternalPropertiesClass<TableInternalProperties> {
 					return Promise.reject(new CustomError.OtherError(`Table ${this.name} has not been initialized.`));
 				}
 
-				return internalPropertiesObject.ready ? Promise.resolve() : new Promise((resolve) => {
-					internalPropertiesObject.pendingTasks.push(resolve);
+				return internalPropertiesObject.ready ? Promise.resolve() : new Promise((resolve, reject) => {
+					internalPropertiesObject.pendingTasks.push({resolve, reject});
 				});
 			},
 			"models": models.map((model: any) => {
@@ -243,13 +243,21 @@ export class Table extends InternalPropertiesClass<TableInternalProperties> {
 					});
 				}, Promise.resolve());
 
-				await setupFlowPromise;
+				try {
+					await setupFlowPromise;
 
-				this.getInternalProperties(internalProperties).ready = true;
-				this.getInternalProperties(internalProperties).setupFlowRunning = false;
+					this.getInternalProperties(internalProperties).ready = true;
+					this.getInternalProperties(internalProperties).setupFlowRunning = false;
 
-				this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task());
-				this.getInternalProperties(internalProperties).pendingTasks = [];
+					this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task.resolve());
+					this.getInternalProperties(internalProperties).pendingTasks = [];
+				} catch (error) {
+					this.getInternalProperties(internalProperties).ready = false;
+					this.getInternalProperties(internalProperties).setupFlowRunning = false;
+
+					this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task.reject(error));
+					this.getInternalProperties(internalProperties).pendingTasks = [];
+				}
 			}
 		});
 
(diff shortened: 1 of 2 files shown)


jinseok0:

packages/dynamoose/lib/Table/index.ts
```diff
@@ -243,13 +243,21 @@ export class Table extends InternalPropertiesClass<TableInternalProperties> {
...
 
-				this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task());
-				this.getInternalProperties(internalProperties).pendingTasks = [];
+					this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task.resolve());
+					this.getInternalProperties(internalProperties).pendingTasks = [];
+				} catch (error) {
+					this.getInternalProperties(internalProperties).ready = false;
+					this.getInternalProperties(internalProperties).setupFlowRunning = false;
+
+					this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task.reject(error));
```
fishcharlie, line 258: I think this is the only line we need in the `catch` block, right?

fishcharlie (changes requested):
What about unit tests here?

packages/dynamoose/lib/Table/index.ts
jinseok0: yes, this is the most necessary line in that block. I put the 2 lines above because I saw the logic and thought it might be necessary.

jinseok0: @fishcharlie oh I forgot to run the test code.

packages/dynamoose/lib/Table/index.ts
```diff
@@ -257,6 +257,8 @@ export class Table extends InternalPropertiesClass<TableInternalProperties> {
 
 					this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task.reject(error));
 					this.getInternalProperties(internalProperties).pendingTasks = [];
+
+					throw error;
```
jinseok0, line 261: what if I added one more line of code and it looked like this?

jinseok0:
is there anything else we should consider or modify? @fishcharlie

packages/dynamoose/lib/Table/index.ts
```diff
@@ -243,13 +243,23 @@ export class Table extends InternalPropertiesClass<TableInternalProperties> {
...
+					this.getInternalProperties(internalProperties).ready = true;
+					this.getInternalProperties(internalProperties).setupFlowRunning = false;
 
-				this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task());
-				this.getInternalProperties(internalProperties).pendingTasks = [];
+					this.getInternalProperties(internalProperties).pendingTasks.forEach((task) => task.resolve());
+					this.getInternalProperties(internalProperties).pendingTasks = [];
+				} catch (error) {
+					this.getInternalProperties(internalProperties).ready = false;
+					this.getInternalProperties(internalProperties).setupFlowRunning = false;
```
fishcharlie, line 256: Still not sure why these lines are required.

packages/dynamoose/lib/Table/index.ts
fishcharlie: Not sure what you mean here.

fishcharlie (changes requested):
Still need unit tests.

fishcharlie:
@jinseok0 Still need unit tests

packages/dynamoose/lib/Table/index.ts
jinseok0: If you look at the existing unit tests, there are already test cases for errors in that function. Here, if I prevent the error from occurring with a catch statement, I have to rewrite the related test cases. However, this may cause problems for other projects that already import that library. But since all I want is to raise a reject on the error so that the pendingTasks don't get stuck forever, I've made sure to raise the error again at the end of that catch statement. @fishcharlie

packages/dynamoose/lib/Table/index.ts
jinseok0: If I really don't need it, I'll delete the line. @fishcharlie

jinseok0:
I removed the unnecessary lines and added a unit test. @fishcharlie

jinseok0:
test

Issue #1647, opened 2023-12-21
[FEATURE] Type safe custom methods on models and items


tranhl:
### Summary:

Currently, there is no way to type custom methods set on models and items. See this [this discussion](https://github.com/dynamoose/dynamoose/discussions/1646#discussion-5981044) for context.

Having the ability to provide type information to Dynamoose around these methods will increase type safety while using the library.

For model methods, a straightforward solution would be to introduce a second generic type parameter to the `model` that allows us to strongly type the model's methods. However, for item methods, there are three options. One is to add a generic to the `Item` itself. The second is to instruct consumers to supply custom methods while extending the `Item` type. The third is to add another generic to the `model` factory function.

@fishcharlie Keen to hear your perspective on which opens would be best in terms of DX and feasibility. My team is pretty keen on having this as a feature, so we might make PR if we can come to an agreement on how the enhancement should look.

#### Model Methods
```ts
import { Schema, model } from 'dynamoose';
import { Item } from 'dynamoose/dist/Item';

type UserItem = Item & {
  id: string;
  email: string;
};

type ModelMethods = {
  getByEmail: () => Promise<UserItem>; // 👈 Define type for model method
};

export const User = model<UserItem, ModelMethods>( // 👈 Pass it to model factory to apply type
  'User',
  new Schema({
    id: { type: String, hashKey: true },
    email: { type: String, required: true }
  })
);

User.methods.set('getByEmail', function () {});
```

#### Item Methods (Generic on Item type)
```ts
import { Schema, model } from 'dynamoose';
import { Item } from 'dynamoose/dist/Item';

type UserItem = Item<{
  resetPassword: () => Promise<void>; // 👈 Item method defined as part of defining item type
}> & {
  id: string;
  email: string;
};

type ModelMethods = {
  getByEmail: () => Promise<UserItem>;
};

export const User = model<UserItem, ModelMethods>(
  'User',
  new Schema({
    id: { type: String, hashKey: true },
    email: { type: String, required: true }
  })
);

User.methods.set('getByEmail', function () {});
User.methods.item.set('resetPassword', function() {});
```

#### Item Methods (Extend Item type)
```ts
import { Schema, model } from 'dynamoose';
import { Item } from 'dynamoose/dist/Item';

type UserItem = Item & {
  id: string;
  email: string;
} & {
  resetPassword: () => Promise<void>; // 👈 Same as previous example, except the method is defined as an intersection instead
};

type ModelMethods = {
[...]

PR #1659, opened 2024-02-03
feat: strongly typed item methods

tranhl:
### Summary:

This PR adds support for strongly-typed custom item methods via a second generic type parameter to `dynamoose.model`.

### Code sample:
```ts
// 1. Create a custom item method type that extends from `ItemMethods`.
// This is ensures that only functions are present in the type.
interface UserItemMethods extends ItemMethods {
	resetPassword: () => Promise<void>;
}

// 2. Create a type for `Item` as usual.
interface UserItem extends Item {
	id: string;
	name: string;
	age: number;
}

// 3. Create a schema.
const userSchema = new dynamoose.Schema({
	"id": String,
	"name": {
		"type": String,
		"index": {
			"type": IndexType.global
		}
	},
	"age": Number
});

// 4. Pass item & custom method types to `dynamoose.model`
const UserModel = dynamoose.model<UserItem, UserItemMethods>(
	"User",
	userSchema
);

const user = await UserModel.get('id');
// 5. Access item methods with type safety.
user.resetPassword() // () => Promise<void>
```

### GitHub linked issue:

Implements part of #1647 

### Type (select 1):
- [ ] Bug fix
- [x] Feature implementation
- [ ] Testing improvement
- [ ] Test added to report bug

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/packages/dynamoose/lib/General.ts
+++ b/packages/dynamoose/lib/General.ts
@@ -1,4 +1,4 @@
-import {Item} from "./Item";
+import {Item, ItemMethods} from "./Item";
 import {Model} from "./Model";
 
 // - General
@@ -14,7 +14,9 @@ export type KeyObject = {[attribute: string]: string | number};
 // An item representing a DynamoDB key
 export type InputKey = string | number | KeyObject;
 
-interface ModelItemConstructor<T extends Item> {
+type Expand<T> = {[KeyType in keyof T]: T[KeyType]} & {};
+
+interface ModelItemConstructor<T extends Item, U extends ItemMethods> {
 	/**
 	 * In order to create a new item you just pass in your object into an instance of your model.
 	 *
@@ -29,10 +31,10 @@ interface ModelItemConstructor<T extends Item> {
 	 * // myUser is now an item instance of the User model
 	 * ```
 	 */
-	new (object: {[key: string]: any}): T;
-	Model: Model<T>;
+	new (object: Expand<Omit<T, keyof Item>>): T & U;
+	Model: Model<T, U>;
 }
-export type ModelType<T extends Item> = T & Model<T> & ModelItemConstructor<T>;
+export type ModelType<T extends Item, U extends ItemMethods> = T & Model<T, U> & ModelItemConstructor<T, U>;
 
 // This represents an item array. This is used for the output of functions such as `scan`, `query`, and `batchGet`. These functions can extend this property to add additional properties or functions. However this represents the shared properties/functions for all item arrays.
 export interface ItemArray<T> extends Array<T> {
--- a/packages/dynamoose/lib/General.ts
+++ b/packages/dynamoose/lib/General.ts
@@ -1,5 +1,6 @@
 import {Item, ItemMethods} from "./Item";
 import {Model} from "./Model";
+import {Flatten} from "./Types";
 
 // - General
 export type CallbackType<R, E> = (error?: E | null, response?: R) => void;
@@ -13,8 +14,6 @@ export type DeepPartial<T> = {[P in keyof T]?: DeepPartial<T[P]>};
 export type KeyObject = {[attribute: string]: string | number};
 // An item representing a DynamoDB key
 export type InputKey = string | number | KeyObject;
-
-type Expand<T> = {[KeyType in keyof T]: T[KeyType]} & {};
 
 interface ModelItemConstructor<T extends Item, U extends ItemMethods> {
 	/**
@@ -31,7 +30,7 @@ interface ModelItemConstructor<T extends Item, U extends ItemMethods> {
 	 * // myUser is now an item instance of the User model
 	 * ```
 	 */
-	new (object: Expand<Omit<T, keyof Item>>): T & U;
+	new (object: Flatten<Omit<T, keyof Item>>): T & U;
 	Model: Model<T, U>;
 }
 export type ModelType<T extends Item, U extends ItemMethods> = T & Model<T, U> & ModelItemConstructor<T, U>;
--- a/packages/dynamoose/lib/General.ts
+++ b/packages/dynamoose/lib/General.ts
@@ -15,7 +15,7 @@ export type KeyObject = {[attribute: string]: string | number};
 // An item representing a DynamoDB key
 export type InputKey = string | number | KeyObject;
 
-interface ModelItemConstructor<T extends Item, U extends ItemMethods> {
+interface ModelItemConstructor<T extends Item, U extends ItemMethods<U>> {
 	/**
 	 * In order to create a new item you just pass in your object into an instance of your model.
 	 *
@@ -33,7 +33,7 @@ interface ModelItemConstructor<T extends Item, U extends ItemMethods> {
 	new (object: Flatten<Omit<T, keyof Item>>): T & U;
 	Model: Model<T, U>;
 }
-export type ModelType<T extends Item, U extends ItemMethods> = T & Model<T, U> & ModelItemConstructor<T, U>;
+export type ModelType<T extends Item, U extends ItemMethods<U>> = T & Model<T, U> & ModelItemConstructor<T, U>;
 
 // This represents an item array. This is used for the output of functions such as `scan`, `query`, and `batchGet`. These functions can extend this property to add additional properties or functions. However this represents the shared properties/functions for all item arrays.
 export interface ItemArray<T> extends Array<T> {
--- a/packages/dynamoose/lib/Item.ts
+++ b/packages/dynamoose/lib/Item.ts
@@ -30,6 +30,14 @@ export interface ItemSettings {
 	type?: "fromDynamo" | "toDynamo";
 }
 
+export interface ItemMethods {
+	[key: string]: Function;
+}
+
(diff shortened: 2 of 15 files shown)


tranhl:
@fishcharlie Hoping to get your input on my design/implementation of strongly typed item methods. I've created a draft PR for now just in case the design needs rework.

I've done the best I can to make sure that anywhere an `Item` is returned, the `ItemMethods` are now also being returned as well (e.g. after executing a `Scan` or `Query` operation). Please do point out if I've missed anything.

tranhl:
I dedicate any and all copyright interest in this software to the
public domain. I make this dedication for the benefit of the public at
large and to the detriment of my heirs and successors. I intend this
dedication to be an overt act of relinquishment in perpetuity of all
present and future rights to this software under copyright law.

packages/dynamoose/lib/Model/index.ts
```diff
@@ -439,21 +439,21 @@ export class Model<T extends ItemCarrier = AnyItem> extends InternalPropertiesCl
 	// latestTableDetails: DynamoDB.DescribeTableOutput;
 	// pendingTaskPromise: () => Promise<void>;
 	Item: typeof ItemCarrier;
```
tranhl, line 441: @fishcharlie Is being able to access the underlying `Item` class from a model part of public interface? If it is not, then there aren't any issues. However if it is, then we'll have to figure out a way to better type this property.

For example, this is possible:

```js
const UserModel = dynamoose.model('User', userSchema)
const user = new UserModel.Item({ id: 'foobar' })
```

This works fine in JS, however typing this correctly in TS has proven challenging. I made a naive attempt at fixing the types: 

```suggestion
	Item: T & typeof ItemCarrier;
```

However, TypeScript understandably isn't happy with this, as the internal `Item` class created within `Model` is not assignable to the generic we pass in:

```
Type 'typeof Item' is not assignable to type 'T & typeof Item'.
  Type 'typeof Item' is not assignable to type 'T'.
    'T' could be instantiated with an arbitrary type which could be unrelated to 'typeof Item'.
```

Haven't been able to find a good solution to this yet.

tranhl:
recheck

tranhl:
recheck

packages/dynamoose/lib/General.ts
```diff
@@ -14,7 +14,9 @@ export type KeyObject = {[attribute: string]: string | number};
 // An item representing a DynamoDB key
 export type InputKey = string | number | KeyObject;
 
-interface ModelItemConstructor<T extends Item> {
+type Expand<T> = {[KeyType in keyof T]: T[KeyType]} & {};
```
fishcharlie, line 17: Why the `& {}`?

Also, can you give an example of what this would look like? For some reason this type seems very complex at first glance. It's not obviously clear what it's doing and what it's purpose is. Maybe add a comment above with an example or something? Or a better description?

fishcharlie: Also should this be in a separate file?

packages/dynamoose/lib/Item.ts
```diff
@@ -30,6 +30,14 @@ export interface ItemSettings {
 	type?: "fromDynamo" | "toDynamo";
 }
 
+export interface ItemMethods {
+	[key: string]: Function;
+}
+
+export interface AnyItemMethods {
+	[key: string]: any;
+}
```
fishcharlie, line 39: It's not immediately clear what this is doing. Seems more general than just `AnyItemMethods`. Why do we have to use `any` vs `Function`? If it's item methods, shouldn't that be a function?

packages/dynamoose/lib/Model/index.ts
fishcharlie: @tranhl 

> Is being able to access the underlying Item class from a model part of public interface?

It might work. But it's not supported. If someone tries to do that in their code I would point them to this: https://dynamoosejs.com/other/FAQ#can-i-use-an-undocumented-property-class-method-or-function-in-dynamoose.

packages/dynamoose/test/types/Item.ts
```diff
@@ -11,3 +11,5 @@ const shouldPassSaveWithReturnItemCallback = user.save({"return": "item"}, () =>
 
 // @ts-expect-error
 const shouldFailWithInvalidReturnType = user.save({"return": "invalid-return-type"});
+
+const shouldPassCustomMethodAccess = user.resetPassword();
```
fishcharlie, line 15: Maybe we should also have a test with a method that doesn't exist? We can use `// @ts-expect-error` for that one.

packages/dynamoose/test/types/Model.ts
```diff
@@ -89,7 +95,7 @@ const userSchema = new dynamoose.Schema({
 	"age": Number
 });
 
-export const UserTypedModel = dynamoose.model<User>(
+export const UserTypedModel = dynamoose.model<User, UserMethods>(
```
fishcharlie, line 98: What happens here if you don't pass in `UserMethods`? Does it still work? I kinda think we should be breaking this out into two different tests. The existing tests are just fine. We should have new tests for these UserMethods that don't modify the old tests.

packages/dynamoose/test/types/Model.ts
```diff
@@ -110,3 +116,6 @@ UserTypedModel.update({"id": "foo"}, {
 UserTypedModel.update({"id": "foo"}, {
 	"$REMOVE":{"age":null}
 });
+
+// @ts-expect-error
+const shouldFailItemCustomMethodAccess = UserTypedModel.resetPassword();
```
fishcharlie, line 121: So are model methods outside the scope of this PR? This PR only deals with item methods it looks like?

fishcharlie (changes requested):
Overall this looks really good.

Beyond the code level comments I made the only other high level suggestion is that we need documentation for this. All the documentation either exists in the code itself using JSDoc or in [`docs/docs_src`](https://github.com/dynamoose/dynamoose/tree/main/docs/docs_src).

packages/dynamoose/test/types/Model.ts
tranhl: That's correct, I've decided to address them separately to keep scope focused. Happy to include model methods as well if that suits you?

packages/dynamoose/lib/General.ts
tranhl: Good points! This type definitely could use a JSDoc explaining its function & intended use. Also agree that it should be moved to a separate file. Best location for it from what I can see would be `lib/Types.ts`. How does that sound?

packages/dynamoose/lib/Item.ts
tranhl: Fair point. I encountered an issue with TS early on omitting the `ItemMethods` generic when calling `dynamoose.model`, however I believe that requiring custom methods to extend `ItemMethods` addresses this. Let me revisit this and see if it's possible to completely remove it.

packages/dynamoose/lib/Model/index.ts
tranhl: Okay perfect 👍 .

It's also mentioned in the TS docs that [documentation should be treated as the source of truth](https://dynamoosejs.com/getting_started/TypeScript#should-i-trust-the-documentation-or-typescript-typings-more), so we can set this aside until `Model.Item` becomes public interface.

packages/dynamoose/test/types/Item.ts
tranhl: Definitely. Test coverage across edge cases was minimal as I was unsure if you'd be happy with the approach of the PR. Will update/refactor tests to be more comprehensive, and update the documentation as well.

packages/dynamoose/test/types/Model.ts
tranhl: As part of the interface, it should work. As per [my previous comment](https://github.com/dynamoose/dynamoose/pull/1659#discussion_r1477551646), I will make sure this is covered in tests.

packages/dynamoose/lib/General.ts
fishcharlie: `lib/Types.ts` works for me 🎉

packages/dynamoose/test/types/Model.ts
fishcharlie: @tranhl Ideally I'd like to review both PRs before merging either of them. I don't want to be left in a state where one piece is done and people ask why the other piece wasn't implemented.

You can decide how to best structure it. I'd personally prefer 2 separate PRs so that it's easier to review and the context is easier to understand. Sadly I don't think GitHub has the concept of having one PR depend on another.

jimmyn:
It would be great to have this feature live!

tranhl:
FYI my team has chosen to adopt [DynamoDB Toolbox V1](https://www.dynamodbtoolbox.com/) as solves all of the friction of dealing with the DynamoDB Client. As such I'm going to close this PR and recommend that anyone looking for type-safe DynamoDB use Toolbox instead.

Issue #1716, opened 2024-12-30
[BUG] Item.save issue


nowukh:
### Summary:

If you attempt to save an object with multiple values referencing the same object, only one value will be saved correctly, and the others will be ignored.

It seems to be due to the isCircular handling when dealing with objects in the deep_copy utility.

#### General
```
const date = new Date()
const a = {
    birtdate: date,
    createdAt: date
}
Model.create(a)
```

### Environment:

Operating System:
Operating System Version:
Node.js version (`node -v`):
NPM version: (`npm -v`):
Dynamoose version:

### Other:
- [ ] I have read through the Dynamoose documentation before posting this issue
- [ ] I have searched through the GitHub issues (including closed issues) and pull requests to ensure this issue has not already been raised before
- [ ] I have searched the internet and Stack Overflow to ensure this issue hasn't been raised or answered before
- [ ] I have tested the code provided and am confident it doesn't work as intended
- [ ] I have filled out all fields above
- [ ] I am running the latest version of Dynamoose

PR #1673, opened 2024-04-14
fix: query perf degradation from `deep_copy` util

tranhl:
### Summary:

I believe this fixes #1260.

[See this comment for context & details](https://github.com/dynamoose/dynamoose/issues/1260#issuecomment-2053897337)

### GitHub linked issue:

#1260

### Type (select 1):
- [ ] Bug fix
- [ ] Feature implementation
- [ ] Documentation improvement
- [ ] Testing improvement
- [ ] Test added to report bug (GitHub issue #---- @---)
- [x] Something not listed here

### Is this a breaking change? (select 1):
- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

### Is this ready to be merged into Dynamoose? (select 1):
- [x] Yes
- [ ] No

### Are all the tests currently passing on this PR? (select 1):
- [x] Yes
- [ ] No

### Other:
- [x] I have read through and followed the Contributing Guidelines
- [x] I have searched through the GitHub pull requests to ensure this PR has not already been submitted
- [x] I have updated the Dynamoose documentation (if required) given the changes I made
- [x] I have added/updated the Dynamoose test cases (if required) given the changes I made
- [x] I agree that all changes made in this pull request may be distributed and are made available in accordance with the [Dynamoose license](https://github.com/dynamoose/dynamoose/blob/main/LICENSE)
- [x] All of my commits and commit messages are detailed, explain what changes were made, and are easy to follow and understand
- [x] I have filled out all fields above

PR #1758, opened 2025-10-14
fix: deep_copy dropping shared references in sibling properties

### LINKED ISSUE #1781: [BUG] `deep_copy` silently drops nested object properties
### Summary:

- When the same JavaScript object reference is assigned to two sibling keys in an item .
   - For example, consider an `Order` whose `billingAddress` and `shippingAddress` point to the same `address` object.
   - The payload generated by `Model.batchPut` (and `Model.create` ,`Model.update`) silently drop  the second occurrence from the request payload that Dynamoose sends to DynamoDB.
- The corruption originates in `utils.deep_copy`, which is called by `Item.objectFromSchema` (and the `Item` constructor used by `Model.create` / `Model.batchPut`) on every user-supplied write payload.

#### Schema
```js
  customerId: { type: String, required: true },
  total: { type: Number, required: true },
  billingAddress: {
    type: Object,
    schema: {
      line1: { type: String },
      city: { type: String },
      country: { type: String },
    },
  },
  shippingAddress: {
    type: Object,
    schema: {
      line1: { type: String },
      city: { type: String },
      country: { type: String },
    },
  },
});
```

#### Model
```js
const Order = dynamoose.model("Order", schema, {
  create: false,
  waitForActive: false,
});
```

#### General
```js
// A common case: the customer used the same address for billing and shipping,
// so the application reuses one `address` object as both fields.
// This is just a demo version, and I acknowledge that it might not be common.
const address = { line1: "1 Infinite Loop", city: "Cupertino", country: "US" };

const orders = [
  {
    id: "o1",
    customerId: "c1",
    total: 100,
    billingAddress: address,
    shippingAddress: address,
  },
  {
    id: "o2",
    customerId: "c2",
    total: 200,
    billingAddress: address,
    shippingAddress: address,
  },
];

// We use `return: "request"` so this repro doesn't need a live DynamoDB.
const req = await Order.batchPut(orders, { return: "request" });

// Print the actual PutRequest.Item that Dynamoose would send to DynamoDB.
const writes = req.RequestItems[Object.keys(req.RequestItems)[0]];
const writes = req.RequestItems?.[tableName];
 (writes?? []).forEach((w, i) => {
    // unmarshall converts a DynamoDB AttributeValue tree back to Plain Old JavaScript Objects
    const decoded = unmarshall(w.PutRequest?.Item ?? {}) as Partial<Order>;
    console.log(`order[${i}] id=${decoded.id}`);
    console.log('  billingAddress  =', decoded.billingAddress);
    console.log('  shippingAddress =', decoded.shippingAddress);
  });
```

```
order[0] id=o1
  billingAddress  = { line1: '1 Infinite Loop', city: 'Cupertino', country: 'US' }
  shippingAddress = undefined
order[1] id=o2
  billingAddress  = { line1: '1 Infinite Loop', city: 'Cupertino', country: 'US' }
  shippingAddress = undefined
```
- `shippingAddress` is silently omitted from the `PutRequest.Item` for **every** order in the batch. No error is thrown.
- The original `orders` array is not mutated. `orders[0].billingAddress === orders[0].shippingAddress` is still `true` after the call, and both still contain the full address. Only the deep copy that becomes the wire payload is missing the field.
- The same drop happens through `Model.create(orders[0])` and `Model.update({"id": "o1"}, {"billingAddress": address, "shippingAddress": address})`.

```
order[0] id=o1
  billingAddress  = { line1: '1 Infinite Loop', city: 'Cupertino', country: 'US' }
  shippingAddress = { line1: '1 Infinite Loop', city: 'Cupertino', country: 'US' }
order[1] id=o2
  billingAddress  = { line1: '1 Infinite Loop', city: 'Cupertino', country: 'US' }
  shippingAddress = { line1: '1 Infinite Loop', city: 'Cupertino', country: 'US' }
```

- A non-circular shared reference should be deep-copied as two independent equal sub-trees on the wire the same way `lodash.cloneDeep`, `JSON.parse(JSON.stringify(…))`, and Dynamoose `<4.x` all behaved.

### Environment:

Operating System: Ubuntu (also reproduces on macOS)
Operating System Version: 22.04
Node.js version (`node -v`): v20.x
NPM version (`npm -v`): 10.x
Dynamoose version: `4.1.5` (also confirmed on `4.0.0`  i.e. likely every release since this [PR](https://github.com/dynamoose/dynamoose/pull/1673))

The bug is silent and no error is thrown, the original input is untouched, and the corruption only surfaces when the data is read back and validated. Real-world patterns that hit it:
- An `Order` (above) where the customer reused one address for billing and shipping.
- An audit-style record where `createdBy` and `updatedBy` are set to the same `user` object.
- A `batchPut` whose items each carry the same shared `metadata` blob, used at two sibling keys (e.g. `metadata` and `enrichedMetadata`).
- Any code path that builds an item via `{...base, foo: shared, bar: shared}` where `shared` is a sub-object built once and used twice.

It can be reproduced even more minimally against `utils.deep_copy` directly, with no model:
```js
const deep_copy = require("dynamoose/dist/utils/deep_copy").default;

const address = { line1: "1 Infinite Loop", city: "Cupertino", country: "US" };
const order = { id: "o1", billingAddress: address, shippingAddress: address };
console.log(deep_copy(order));
// expected: { id: 'o1', billingAddress: {…}, shippingAddress: {…} }
// actual:   { id: 'o1', billingAddress: {…} }       <-- shippingAddress dropped
```
- The reproduction repository can be found [here](https://github.com/JonnavithulaGirish/dynamoose-deep-copy-repo)

### Other:

EeeasyCode:
### Summary:

Fixes `deep_copy` dropping sibling properties that share the same object reference.

**Root cause.** The `refs` Set introduced in #1673 is a cumulative visited-set: once an object is seen anywhere in the traversal, any later occurrence is treated as a circular reference and omitted. Because true cycles and legitimate shared references share the same Set, two sibling properties holding the same value (e.g. `createdAt === updatedAt`, or `billingAddress === shippingAddress`) lose the second property.

**Fix.** Change `refs` from a cumulative visited-set to an ancestor-path stack — `refs.add(obj)` on entering an object/array, `refs.delete(obj)` after the child recursion returns. Only values that lie on the *current* recursion path are treated as cycles, which matches the documented "omit circulars" semantics without conflating it with shared references.

**Scope of this PR (all share the same root cause):**
- Same `Date` / `Buffer` / `Set` / `Map` / `Uint8Array` instance in multiple properties (the originally reported case).
- Same plain object or class instance in multiple properties — closes #1781.
- Self-referencing arrays (`arr.push(arr)`) no longer infinite-recurse.
- Circular properties on class instances no longer leak the original reference through the removed `Object.assign(Object.create(proto), obj)` shallow copy.

**Additional review feedback addressed:**
- Buffer clone now uses `Buffer.from(obj)` (single operation) instead of `Buffer.alloc` + `copy` — thanks @fishcharlie for the callout; the previous "TypeScript compatibility" framing was wrong.
- Indentation normalized to tabs per `.eslintrc.js` (`"indent": ["error", "tab"]`).
- Rebased onto latest `main`.
- History squashed to a single commit.

#### General
```javascript
// Before
const now = new Date("2021-03-01T07:00:00Z");
deep_copy({ id: "1", createdAt: now, updatedAt: now });
// → { id: "1", createdAt: Date, updatedAt: undefined }   ❌

const address = { city: "Cupertino" };
deep_copy({ id: "1", billingAddress: address, shippingAddress: address });
// → { id: "1", billingAddress: {city:"Cupertino"}, shippingAddress: undefined }   ❌ (#1781)

// After — both siblings preserved, each as an independent deep copy.
```

### GitHub linked issue:

Closes #1781
Related: #1716

### Other information:

**Test coverage.** Four regression tests added to `test/utils/deep_copy.js`:
1. Same `Date` instance in multiple properties.
2. Same plain object in multiple properties (#1781).
3. Self-referencing array — completes without stack overflow.
4. Class instance with a circular self-property — omitted from copy, constructor preserved.

The existing "omit circular references" invariant (`original.self = original`, `original.c.c = original.c`, `original.c.f.c = original.c` → the cycles are omitted) is preserved by the ancestor-stack semantics.

**Performance.** No regression vs. the current `Set`-based detection — all operations are O(1). Space complexity improves from O(n) (cumulative set) to O(depth) (ancestor stack). The pre-#1673 `isCircular` O(n²) issue is not reintroduced.

**Verification.** `npm run lint` 0 warnings · `npm test` 2,384 passed / 0 failed / 32 suites · `npm run test:types` pass.

### Type (select 1):
- [x] Bug fix
- [ ] Feature implementation
- [ ] Testing improvement

- [ ] 🚨 YES 🚨
- [x] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

### DIFF
--- a/PENDING_CHANGELOG.md
+++ b/PENDING_CHANGELOG.md
@@ -5,3 +5,7 @@
 ### Features
 
 - Added support for enabling and configuring DynamoDB Streams through Table options
+
+### Bug Fixes
+
+- Fixed `deep_copy` dropping sibling properties that share the same object reference (e.g. a single `Date` used for both `createdAt` and `updatedAt`, or a single address used for both `billingAddress` and `shippingAddress`). Circular reference detection now tracks only the current ancestor path rather than every object ever visited, which also removes an infinite-recursion risk on self-referencing arrays and a shared-reference leak on class instances containing circular properties
--- a/packages/dynamoose/lib/utils/deep_copy.ts
+++ b/packages/dynamoose/lib/utils/deep_copy.ts
@@ -3,14 +3,19 @@ export default function deep_copy<T> (obj: T, refs = new Set()): T {
 
 	// Handle Date
 	if (obj instanceof Date) {
-		copy = new Date();
-		copy.setTime(obj.getTime());
+		copy = new Date(obj.getTime());
 		return copy;
 	}
 
-	// Handle Array
-	if (obj instanceof Array) {
-		copy = obj.map((i) => deep_copy(i, refs));
+	// Handle Buffer (must be before Uint8Array since Buffer extends Uint8Array)
+	if (obj instanceof Buffer) {
+		copy = Buffer.from(obj);
+		return copy;
+	}
+
+	// Handle Uint8Array
+	if (obj instanceof Uint8Array) {
+		copy = new Uint8Array(obj);
 		return copy;
 	}
 
@@ -26,49 +31,35 @@ export default function deep_copy<T> (obj: T, refs = new Set()): T {
 		return copy;
 	}
 
-	// Handle Buffer
-	if (obj instanceof Buffer) {
-		copy = Buffer.from(obj);
-		return copy;
-	}
-
-	// Handle Uint8Array
-	if (obj instanceof Uint8Array) {
-		copy = new Uint8Array(obj);
-		return copy;
-	}
-
 	if (obj instanceof Function) {
 		// This is not technically correct, but required for unit test purposes. We currently have a unit test that passes in a function where it shouldn't. So in order to handle this case we need to do something here. Ideally we would clone the function somehow to create a copy of it. But that is lower priority for now.
 		return obj;
 	}
 
-	// Handle Object
+	// Handle Array
+	if (obj instanceof Array) {
+		refs.add(obj);
+		copy = obj.map((item) => item !== null && typeof item === "object" && refs.has(item) ? undefined : deep_copy(item, refs));
+		refs.delete(obj);
+		return copy;
+	}
+
+	// Handle Object (plain object or class instance)
 	if (obj instanceof Object) {
 		refs.add(obj);
-
-		if (obj.constructor !== Object) {
-			copy = Object.assign(Object.create(Object.getPrototypeOf(obj)), obj);
-		} else {
-			copy = {};
-		}
+		copy = obj.constructor !== Object ? Object.create(Object.getPrototypeOf(obj)) : {};
 
 		for (const attr in obj) {
 			if (Object.prototype.hasOwnProperty.call(obj, attr)) {
 				const value = obj[attr];
-				const isObjValue = typeof value === "object" && !Array.isArray(value) && value !== null;
-
-				if (isObjValue) {
-					const isCircular = refs.has(value);
-					// Omit circular property from copy
-					if (isCircular) continue;
-
-					refs.add(value);
+				// Omit true circular references: value is an ancestor on the current recursion path.
+				if (value !== null && typeof value === "object" && refs.has(value)) {
+					continue;
 				}
-
 				copy[attr] = deep_copy(value, refs);
 			}
 		}
+		refs.delete(obj);
 		return copy;
 	}
 
--- a/packages/dynamoose/test/utils/deep_copy.js
+++ b/packages/dynamoose/test/utils/deep_copy.js
@@ -184,4 +184,61 @@ describe("utils.deep_copy", () => {
 		expect(original.toString()).toEqual("\u0000ello World!");
 		expect(copy.toString()).toEqual("Hello World!");
 	});
+
+	it("Should return a deep copy of object with same Date instance in multiple properties", () => {
+		const now = new Date("Mon, 01 Mar 2021 07:00:00 GMT");
+		const original = {"id": "1", "createdAt": now, "updatedAt": now};
+
+		const copy = utils.deep_copy(original);
+
+		expect(copy.createdAt).toBeInstanceOf(Date);
+		expect(copy.updatedAt).toBeInstanceOf(Date);
+		expect(copy.createdAt.toUTCString()).toEqual("Mon, 01 Mar 2021 07:00:00 GMT");
+		expect(copy.updatedAt.toUTCString()).toEqual("Mon, 01 Mar 2021 07:00:00 GMT");
+
+		now.setUTCDate(2);
+		expect(copy.createdAt.toUTCString()).toEqual("Mon, 01 Mar 2021 07:00:00 GMT");
+		expect(copy.updatedAt.toUTCString()).toEqual("Mon, 01 Mar 2021 07:00:00 GMT");
+	});
+
+	it("Should return a deep copy of object with same plain object reference in multiple properties", () => {
+		const address = {"city": "Cupertino", "country": "US"};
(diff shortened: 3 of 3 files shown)


fishcharlie (changes requested):
Can we please revert all the formatting changes that weren't directly part of this PR please?

fishcharlie:
@EeeasyCode Ok few more things. Please merge in the `main` branch. Your indentation isn't correct. You are using spaces instead of tabs. Which doesn't fit this projects style. Based on #1773 the linting should now fail on this PR.

Additionally, I'm a bit confused about this comment:

> Enhanced Buffer copying method to use Buffer.alloc() + copy() for better TypeScript compatibility

How does that improve TypeScript compatibility?  Buffer.alloc creates a new Buffer, and writes 0s to it. Then copy copies the data to it.

Whereas Buffer.from does it all in 1 single operation. Which increases the performance of the operation.

I fail to see how this change is beneficial. And I don't understand at all how it relates to TypeScript compatibility at all.

damienbeaufils:
Hi @EeeasyCode! Any news about this PR and the required changes? Thanks! 🙏

damienbeaufils:
For those who will end up in the same issue, my workaround: instead of patching dynamoose's `deep_copy`, I uses `structuredClone()` to create an independent deep copy before passing it to the entity. This breaks the shared reference so dynamoose's `deep_copy` never encounters the same object twice.

EeeasyCode:
@fishcharlie @damienbeaufils — pushed a squashed rewrite of this PR addressing the review feedback and, after revisiting the code, moving from a symptomatic patch to a root-cause fix.

**What changed vs. the previous attempt**

The earlier version special-cased `Date` / `Buffer` / `Set` / `Map` / `Uint8Array` via an `isPrimitiveLike` bypass. That only masked the symptom: the actual bug is that `refs` was being used as a cumulative visited-set rather than an ancestor-path stack, so *any* shared sibling reference (including plain objects, as reported in #1781) was being dropped.

This PR changes `refs` to an ancestor stack (`add` on enter, `delete` on exit). One small change, same O(1) lookup, but it correctly distinguishes cycles from shared references. As a side effect it also closes #1781 and removes two latent bugs in the same function (infinite recursion on self-referencing arrays, and a shared-reference leak on class-instance circular properties via the now-removed `Object.assign` shallow copy).

**Review feedback addressed**
- `Buffer.from(obj)` instead of `Buffer.alloc` + `copy` — you were right, single-op is better and the TypeScript-compatibility framing was incorrect.
- Indentation normalized to tabs; `npm run lint` 0 warnings.
- Rebased onto latest `main`.
- History squashed to a single commit.

**Verification**
- `npm test` — 2,384 passed / 0 failed / 32 suites (including 4 new regression tests).
- `npm run test:types` — pass.
- `npm run lint` — 0 warnings.

@damienbeaufils your `structuredClone()` workaround should no longer be necessary once this lands — thanks for filing the repro context.

Happy to split out the array / class-instance fixes into separate PRs if you'd prefer tighter scope, but since they trace back to the same `refs`-semantic root cause and are covered by dedicated regression tests, bundling felt cleaner. Let me know either way.

fishcharlie (approved):

fishcharlie (approved):
This looks good to me. I reverted the code changes and ensured the tests fail, and have verified the tests pass when the code changes exist.

Code level review looks good as well.

PR #1778, opened 2026-03-09
Merge nested object properties on update instead of replacing

### LINKED ISSUE #1776: [BUG] Model.update() replaces entire nested Object instead of merging partial properties
### Summary:

When calling `Model.update()` with a partial nested object, Dynamoose replaces the entire nested attribute in DynamoDB instead of only updating the provided properties.

For example, if an item has `settings: { theme: "dark", notifications: true }` and you call `update({ id }, { settings: { theme: "light" } })`, the resulting item will have `settings: { theme: "light" }` with `notifications` completely removed.

This happens because the update method in `Model/index.js` (around line 544) maps the entire nested object as a single `SET #a0 = :v0` expression, where `:v0` is the partial object. DynamoDB then replaces the full `settings` map attribute.

Related to #1589 and #1775 (which covers the TypeScript side of this problem).

#### Schema
```ts
const ProductSchema = new Schema({
  config: {
    type: Object,
    schema: {
      featureA: { type: Boolean, default: false },
      featureB: { type: Boolean, default: false },
    },
    default: { featureA: false, featureB: false },
  },
});
```

#### Model
```ts
const ProductModel = dynamoose.model("Product", ProductSchema);
```

#### General
```ts
// Create an item
await ProductModel.create({
  id: "123",
  config: { featureA: true, featureB: true },
});

// Update only one nested property
await ProductModel.update({ id: "123" }, { config: { featureA: false } });

// Fetch the item
const item = await ProductModel.get({ id: "123" });
console.log(item.config);
// Actual:   { featureA: false }              <-- featureB is gone!
// Expected: { featureA: false, featureB: true }
```

The update generates a DynamoDB `UpdateExpression` of `SET #a0 = :v0` where:
- `ExpressionAttributeNames: { "#a0": "config" }`
- `ExpressionAttributeValues: { ":v0": { "featureA": false } }`

This tells DynamoDB to replace the entire `config` attribute with `{ featureA: false }`, discarding any properties not included in the partial update.

Dynamoose should generate individual SET expressions for each nested property:

```
SET #a0.#a1 = :v0
ExpressionAttributeNames: { "#a0": "config", "#a1": "featureA" }
ExpressionAttributeValues: { ":v0": false }
```

This is how DynamoDB natively supports partial updates on nested map attributes. Only the specified nested properties should be updated, while others remain unchanged.

Alternatively, a schema-level or update-level option (e.g. `{ mergeNestedObjects: true }`) could control this behavior to avoid breaking changes.

### Environment:

Operating System: Windows
Operating System Version: 11 Pro
Node.js version (`node -v`): v22.20.0
NPM version: (`npm -v`): 11.6.2
Dynamoose version: 4.1.5

Current workaround requires a read-then-write pattern:

```ts
const existing = await ProductModel.get({ id: "123" });
await ProductModel.update(
  { id: "123" },
  { config: { ...existing.config, featureA: false } }
);
```

This adds an extra read operation and introduces a potential race condition if two updates happen concurrently.

### Other:

Adi1231234:
## Summary

`Model.update()` replaces entire nested objects in DynamoDB instead of merging partial properties. When calling `update({ id }, { config: { featureA: false } })`, DynamoDB receives `SET #config = :v` which overwrites the entire `config` attribute, discarding any properties not included in the partial update.

**Root cause:** The update expression builder treats nested objects as single values, generating one `SET` expression for the whole attribute.

**Fix:** When a `$SET` operation targets a Map attribute (`"M"` type) that has a sub-schema, flatten the nested object into individual dot-notation `SET` expressions. This uses DynamoDB's native support for updating nested map attributes.

### Before
```
SET #a0 = :v0
ExpressionAttributeNames:  { "#a0": "config" }
ExpressionAttributeValues: { ":v0": { "M": { "featureA": { "BOOL": false } } } }
```
DynamoDB replaces the entire `config` map, removing any properties not in the update.

### After
```
SET #a0.#a1 = :v2
ExpressionAttributeNames:  { "#a0": "config", "#a1": "featureA" }
ExpressionAttributeValues: { ":v2": { "BOOL": false } }
```
DynamoDB updates only `config.featureA`, leaving other properties untouched.

### Changes

- **`packages/dynamoose/lib/Model/index.ts`**: Detect partial nested updates (Map type + sub-schema + plain object value) and recursively generate dot-notation SET expressions instead of a single whole-object SET
- **`packages/dynamoose/test/Model.js`**: Update 2 tests that expected the old "replace" behavior to verify the new "merge" behavior with correct dot-notation expressions

### Design decisions

- Only applies to `$SET` operations on Map attributes with a defined sub-schema
- Objects without a sub-schema (generic `{ type: Object }`) retain the current replace behavior
- Recursively flattens deeply nested objects to their leaf values
- Passes `required: false` to `objectFromSchema` so partial updates don't trigger validation errors for missing sibling properties

## Test plan
- [x] `npm run build` passes
- [x] Type tests pass (`tsc --noEmit`)
- [x] All 710 runtime tests pass (4 updated to match new merge behavior)

Closes #1776

### DIFF
--- a/packages/dynamoose/lib/Model/index.ts
+++ b/packages/dynamoose/lib/Model/index.ts
@@ -703,6 +703,7 @@ export class Model<T extends ItemCarrier = AnyItem> extends InternalPropertiesCl
 		const table = this.getInternalProperties(internalProperties).table();
 		const {instance} = table.getInternalProperties(internalProperties);
 		let index = 0;
+		const isObject = (v: any): boolean => typeof v === "object" && v !== null && !Array.isArray(v);
 		const getUpdateExpressionObject: () => Promise<any> = async () => {
 			const updateTypes = [
 				{"name": "$SET", "operator": " = ", "objectFromSchemaSettings": {"validate": true, "enum": true, "forceDefault": true, "required": "nested", "modifiers": ["set"]}},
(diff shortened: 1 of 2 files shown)


Adi1231234:

PR #1783, opened 2026-06-20
fix: nested/partial Model.update, custom-type consistency, and array traversal

KennethWKZ:
### Summary:

Six independent `Item` / `Model` fixes, rebased into one commit per concern. Two are behavioral, the rest are a crash fix, a build fix, a type-only change, and a new opt-in setting.

The branch was restructured after the first review round. The nested `$SET` change that prompted the breaking-change question is now **opt-in** (`{merge: true}`) and the default behavior is unchanged. The only remaining breaking change is `Item.save` returning custom types consistently with `Model.get`.

| Commit | What |
|--------|------|
| `fix(model)` DeepPartial typing | Type-only. `Partial<T>` did not recurse, so partially specified nested attributes were rejected at the type level |
| `fix(item)` array attribute crash | `attributesWithSchema` threw `node.forEach is not a function` when an object was supplied for an attribute the schema declares as an array |
| `fix(utils)` merge_objects cast | `main` does not currently compile: `Operator '+=' cannot be applied to types 'number' and 'T'` |
| `feat(utils)` is_plain_object | New helper + 18 unit tests |
| `feat(update)` `{merge: true}` | Opt-in nested attribute update via document paths. Default unchanged |
| `fix(item)` save callback | The save callback was invoked twice when the callback itself threw |
| `fix(item)` save custom types | 🚨 Breaking. `save` now resolves custom types in the same representation as `get` |

#### Schema

```js
	"id": Number,
	"data": {
		"type": Object,
		"schema": {
			"name": String,
			"age": {"type": Number, "required": true}
		}
	}
});
```

#### Model

```js
// Default — unchanged. The whole attribute is replaced as a map.
await User.update({"id": 1}, {"data": {"name": "Charlie", "age": 30}});
// SET #a0 = :v0   ->   data = {"name": "Charlie", "age": 30}

// Omitting a nested required property still fails validation, as before.
await User.update({"id": 1}, {"data": {"name": "Charlie"}});
// ValidationError: data.age is a required property but has no value when trying to save item

// New — {merge: true} writes each provided property to its own document path.
await User.update({"id": 1}, {"data": {"name": "Charlie"}}, {"merge": true});
// SET #a0.#a1 = :v2   ->   data.age is left untouched
```

#### General

```js
// Merging works at any depth, and on saveUnknown objects with no nested schema.
await User.update({"id": 1}, {"$SET": {"data": {"profile": {"name": "Charlie"}}}}, {"merge": true});
// SET #a0.#a1.#a2 = :v3
```

### Other information:

**Why `{merge: true}` exists.** On `main` there is no way to express a nested attribute update, which DynamoDB supports natively as `SET data.#name = :value`. The whole-object form replaces the map and trips nested `required` checks, and the dotted-key form does not produce a document path:

```js
await User.update({"id": 1}, {"data.name": "Charlie"});
// ExpressionAttributeNames: {"#a0": "data.name"}
// #a0 substitutes the entire string, so this targets a top-level attribute
// whose NAME contains a dot — not data -> name.
```

`{merge: true}` emits `SET #a0.#a1 = :v2` with separate name placeholders, which is the shape the raw SDK uses. Only plain objects are flattened — arrays, Sets, Dates and Buffers are written as leaf values, since recursing into them would emit paths keyed by array indices or byte offsets.

A nested document path requires the parent map to already exist on the item, so `merge` is left unset when the attribute may not exist yet. This is documented in the new "Updating nested objects" section.

**On the breaking change.** `Item.save` applied custom type conversion only to the item it constructed for the return value, not to the item being saved, so `create` and `get` disagreed:

| | before | after |
|---|---|---|
| `save()` → `Date` attribute | `number` | `Date` |
| `save()` → `createdAt` | `number` | `Date` |
| `get()` → `Date` attribute | `Date` | `Date` |

`get` was already correct; this makes `save` agree with it. The point is not the return type as such — it is that a schema declared type should be what you get back from every method. Right now `Date` in the schema means a `Date` from `get` and a `number` from `save`, so the schema means one thing on read and another on write.

The clearest evidence is a test this branch previously had to disable. `Should return correct result after saving with timestamps` asserts `result.toJSON()` equals `result2.toJSON()` after a save and a re-save. That assertion is restored and passing — save and re-save now serialize identically, and so do `save` and `get`.

Because of that I would rather not split this commit out. Landing the rest without it leaves the read/write inconsistency in place, which seems worse than the breaking change itself. If you would prefer it target the next major, that works too — but as a version decision rather than by separating it from the fixes it belongs with.

### Type (select 1):
- [x] Bug fix
- [x] Feature implementation
- [x] Documentation improvement
- [x] Testing improvement

- [x] 🚨 YES 🚨
- [ ] No
- [ ] I'm not sure

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### Other:

---

This pull request was created by an AI agent (Claude Code, Opus 5). The human contributor directed the work and made the design decisions — including the decision to make the nested `$SET` change opt-in rather than the default, and to revert the `.vscode` and `tsconfig` changes — reviewed every change, challenged and corrected the agent's analysis on several points, and approved the submission. The AI agent wrote the code changes and tests, ran the test suite, restructured the commit history, and drafted this description.

### DIFF
--- a/packages/dynamoose/lib/Item.ts
+++ b/packages/dynamoose/lib/Item.ts
@@ -482,8 +482,14 @@ export class Item extends InternalPropertiesClass<ItemInternalProperties> {
 			await promise;
 			this.getInternalProperties(internalProperties).storedInDynamo = true;
 
+			// First apply custom types conversion to the current item
+			await this.conformToSchema({"customTypesDynamo": true, "type": "fromDynamo", "typeCheck": false});
+
 			const returnItem = new (this.getInternalProperties(internalProperties).model).Item(savedItem as any);
 			returnItem.getInternalProperties(internalProperties).storedInDynamo = true;
+
+			// Apply custom types conversion to ensure consistent behavior with get
+			await returnItem.conformToSchema({"customTypesDynamo": true, "type": "fromDynamo"});
 
 			return returnItem;
 		})();
@@ -723,12 +729,14 @@ Item.objectFromSchema = async function (object: any, model: Model<Item>, setting
 			const isValueUndefined = typeof value === "undefined" || value === null;
 			if (!isValueUndefined) {
 				const typeDetails = utils.dynamoose.getValueTypeCheckResult(schema, value, key, settings, {typeIndexOptionMap}).matchedTypeDetails as DynamoDBTypeResult;
-				const {customType} = typeDetails;
-				const {"type": typeInfo} = typeDetails.isOfType(value as ValueType);
-				const isCorrectTypeAlready = typeInfo === (settings.type === "toDynamo" ? "underlying" : "main");
-				if (customType && customType.functions[settings.type] && !isCorrectTypeAlready) {
-					const customValue = customType.functions[settings.type](value);
-					utils.object.set(returnObject, key, customValue);
+				if (typeDetails) {
+					const {customType} = typeDetails;
+					const {"type": typeInfo} = typeDetails.isOfType(value as ValueType);
+					const isCorrectTypeAlready = typeInfo === (settings.type === "toDynamo" ? "underlying" : "main");
+					if (customType && customType.functions[settings.type] && !isCorrectTypeAlready) {
+						const customValue = customType.functions[settings.type](value);
+						utils.object.set(returnObject, key, customValue);
+					}
 				}
 			}
 		});
--- a/packages/dynamoose/lib/Item.ts
+++ b/packages/dynamoose/lib/Item.ts
@@ -478,26 +478,21 @@ export class Item extends InternalPropertiesClass<ItemInternalProperties> {
 			return ddb(table.getInternalProperties(internalProperties).instance, "putItem", putItemObj);
 		});
 
+		const resultPromise = (async (): Promise<Item> => {
+			await promise;
+			this.getInternalProperties(internalProperties).storedInDynamo = true;
+
+			const returnItem = new (this.getInternalProperties(internalProperties).model).Item(savedItem as any);
+			returnItem.getInternalProperties(internalProperties).storedInDynamo = true;
+
+			return returnItem;
+		})();
+
 		if (callback) {
 			const localCallback: CallbackType<Item, any> = callback as CallbackType<Item, any>;
-			promise.then(() => {
-				this.getInternalProperties(internalProperties).storedInDynamo = true;
-
-				const returnItem = new (this.getInternalProperties(internalProperties).model).Item(savedItem as any);
-				returnItem.getInternalProperties(internalProperties).storedInDynamo = true;
-
-				localCallback(null, returnItem);
-			}).catch((error) => callback(error));
+			resultPromise.then((returnItem) => localCallback(null, returnItem), (error) => localCallback(error));
 		} else {
-			return (async (): Promise<Item> => {
-				await promise;
-				this.getInternalProperties(internalProperties).storedInDynamo = true;
-
-				const returnItem = new (this.getInternalProperties(internalProperties).model).Item(savedItem as any);
-				returnItem.getInternalProperties(internalProperties).storedInDynamo = true;
-
-				return returnItem;
-			})();
+			return resultPromise;
 		}
 	}
 
--- a/packages/dynamoose/lib/Item.ts
+++ b/packages/dynamoose/lib/Item.ts
@@ -574,8 +574,8 @@ Item.attributesWithSchema = async function (item: Item, model: Model<Item>): Pro
 
 		Object.keys(treeNode).forEach((attr) => {
 			if (attr === "0") {
-				// We check for empty objects here (added `typeof node === "object" && Object.keys(node).length == 0`, see PR https://github.com/dynamoose/dynamoose/pull/1034) to handle the use case of 2d arrays, or arrays within arrays. `node` in that case will be an empty object.
-				if (!node || node.length == 0 || typeof node === "object" && Object.keys(node).length == 0) {
+				// Anything that isn't a non-empty array has no indices to walk, so we fake the path. This covers a missing node, an empty array, a non-array object, and the empty object used for 2d arrays, or arrays within arrays (see PR https://github.com/dynamoose/dynamoose/pull/1034).
+				if (!Array.isArray(node) || node.length == 0) {
 					node = [{}]; // fake the path for arrays
 				}
 				node.forEach((a, index) => {
--- a/packages/dynamoose/lib/Model/index.ts
+++ b/packages/dynamoose/lib/Model/index.ts
@@ -8,7 +8,7 @@ import Internal from "../Internal";
 import {Serializer, SerializerOptions} from "../Serializer";
 import {Condition, ConditionInitializer} from "../Condition";
 import {Scan, Query} from "../ItemRetriever";
-import {CallbackType, ObjectType, FunctionType, ItemArray, ModelType, KeyObject, InputKey} from "../General";
+import {CallbackType, DeepPartial, ObjectType, FunctionType, ItemArray, ModelType, KeyObject, InputKey} from "../General";
 import {PopulateItems} from "../Populate";
 import {AttributeMap} from "../Types";
 import * as DynamoDB from "@aws-sdk/client-dynamodb";
@@ -21,11 +21,11 @@ import returnModel from "../utils/dynamoose/returnModel";
 const {internalProperties} = Internal.General;
 
 type UpdatePartial<T> =
-  | Partial<T>
-  | { $SET: Partial<T> }
-  | { $ADD: Partial<T> }
+  | DeepPartial<T>
+  | { $SET: DeepPartial<T> }
+  | { $ADD: DeepPartial<T> }
   | { $REMOVE: { [K in keyof T]?: null } | string[] }
-  | { $DELETE: Partial<T> };
+  | { $DELETE: DeepPartial<T> };
 
 // Transactions
 type GetTransactionResult = Promise<GetTransactionInput>;
@@ -712,7 +712,7 @@ export class Model<T extends ItemCarrier = AnyItem> extends InternalPropertiesCl
 			].reverse();
 
 			if (!updateObj) {
-				updateObj = utils.deep_copy(keyObj) as Partial<T>;
+				updateObj = utils.deep_copy(keyObj) as DeepPartial<T>;
 				Object.keys(await this.getInternalProperties(internalProperties).convertKeyToObject(keyObj)).forEach((key) => delete updateObj[key]);
 			}
 
--- a/packages/dynamoose/lib/Model/index.ts
+++ b/packages/dynamoose/lib/Model/index.ts
@@ -87,6 +87,8 @@ interface ModelUpdateSettings {
 	return?: "item" | "request";
 	condition?: Condition;
 	returnValues?: DynamoDB.ReturnValue;
+	// How nested objects passed to `$SET` are applied. `false` (default) replaces the whole attribute as a map (`SET attr = {..}`), `true` flattens the object into document paths (`SET attr.sub = :v`) so subfields that were not passed in are left untouched.
+	merge?: boolean;
 }
 interface ModelBatchGetItemsResponse<T> extends ItemArray<T> {
 	unprocessedKeys: ObjectType[];
@@ -703,6 +705,33 @@ export class Model<T extends ItemCarrier = AnyItem> extends InternalPropertiesCl
 		const table = this.getInternalProperties(internalProperties).table();
 		const {instance} = table.getInternalProperties(internalProperties);
 		let index = 0;
+		// `merge: true` opts IN to flattening a nested object into document paths (`SET attr.sub = :v`), which updates the given subfields and leaves the rest of the map alone. By default the whole attribute is replaced as a map (`SET attr = {..}`).
+		const mergeEnabled = typeof settings === "object" && (settings as ModelUpdateSettings).merge === true;
+
+		const flattenForMerge = (
+			parentPathParts: string[],
+			obj: ObjectType,
+			acc: {ExpressionAttributeNames: ObjectType; ExpressionAttributeValues: ObjectType; UpdateExpression: ObjectType},
+			setOperator: string
+		): void => {
+			for (const key of Object.keys(obj)) {
+				const val = obj[key];
+				const nameKey = `#a${index}`;
+				acc.ExpressionAttributeNames[nameKey] = key;
+				const pathParts = [...parentPathParts, nameKey];
+
+				if (utils.is_plain_object(val) && Object.keys(val).length > 0) {
+					index++;
+					flattenForMerge(pathParts, val, acc, setOperator);
+				} else {
+					const valueKey = `:v${index}`;
+					acc.ExpressionAttributeValues[valueKey] = val;
+					acc.UpdateExpression.SET.push(`${pathParts.join(".")}${setOperator}${valueKey}`);
+					index++;
+				}
+			}
+		};
+
 		const getUpdateExpressionObject: () => Promise<any> = async () => {
 			const updateTypes = [
 				{"name": "$SET", "operator": " = ", "objectFromSchemaSettings": {"validate": true, "enum": true, "forceDefault": true, "required": "nested", "modifiers": ["set"]}},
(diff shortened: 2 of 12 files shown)


KennethWKZ:

KennethWKZ:
@fishcharlie let me know is it this ready to merge in, tested on my side works fine, there another #1784 require this PR to merge in 1st

.vscode/settings.json
```diff
@@ -4,5 +4,6 @@
 		"dynamoose",
 		"serializers",
 		"unmarshall"
-	]
-}
+	],
+	"js/ts.tsdk.path": "node_modules/typescript/lib"
```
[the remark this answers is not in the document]
fishcharlie: I think this is a legitimate comment.

packages/dynamoose/lib/Model/index.ts
```diff
@@ -87,6 +87,10 @@ interface ModelUpdateSettings {
 	return?: "item" | "request";
 	condition?: Condition;
 	returnValues?: DynamoDB.ReturnValue;
+	// Nested object ($SET) handling: `undefined` (default) shallow-merges into the existing map,
+	// `true` deep-merges, `false` replaces the whole attribute as a map (`SET attr = {..}`) — the
+	// only shape that works when the item has no existing map for that attribute yet.
+	merge?: boolean;
```
fishcharlie, line 91: Have we updated the documentation for this?

packages/dynamoose/lib/Model/index.ts
```diff
@@ -703,6 +707,45 @@ export class Model<T extends ItemCarrier = AnyItem> extends InternalPropertiesCl
...
+		const mergeReplace = typeof settings === "object" && (settings as ModelUpdateSettings).merge === false;
+
+		const isPlainObject = (val: any): boolean => {
+			if (val === null || val === undefined || typeof val !== "object") return false;
+			if (Array.isArray(val)) return false;
+			if (val instanceof Set || val instanceof Date) return false;
+			if (typeof Buffer !== "undefined" && val instanceof Buffer) return false;
+			if (val instanceof Uint8Array) return false;
+			return val.constructor === undefined || val.constructor === Object;
+		};
```
fishcharlie: Should this maybe be moved into a helper utility file? That way it's also easier to write dedicated tests for it.

packages/dynamoose/lib/Model/index.ts
```diff
@@ -703,6 +707,45 @@ export class Model<T extends ItemCarrier = AnyItem> extends InternalPropertiesCl
...
+					index++;
+					flattenForMerge(pathParts, val, acc, setOperator);
+				} else {
+					const valueKey = `:v${index}`;
+					acc.ExpressionAttributeValues[valueKey] = val;
+					acc.UpdateExpression.SET.push(`${pathParts.join(".")}${setOperator}${valueKey}`);
+					index++;
+				}
+			}
+		};
```
fishcharlie, line 733: Same as above comment.

packages/dynamoose/lib/utils/merge_objects.ts
```diff
@@ -47,7 +47,7 @@ const main = (settings: MergeObjectsSettings = {"combineMethod": MergeObjectsCom
 					if (settings.combineMethod === MergeObjectsCombineMethod.ArrayMergeNewArray) {
 						returnObject[key] = [returnObject[key], arg[key]];
 					} else if (typeof returnObject[key] === "number") {
-						returnObject[key] += arg[key];
+						returnObject[key] += arg[key] as any;
```
fishcharlie: Any reason we can't make this a more specific type?

packages/dynamoose/lib/Item.ts
```diff
@@ -480,11 +480,16 @@ export class Item extends InternalPropertiesClass<ItemInternalProperties> {
 
 		if (callback) {
 			const localCallback: CallbackType<Item, any> = callback as CallbackType<Item, any>;
-			promise.then(() => {
+			promise.then(async () => {
```
fishcharlie: Not quite sure about this. Feels like a bit of an anti-pattern to have a promise callback with an async function.

packages/dynamoose/test/types/tsconfig.json
```diff
@@ -2,7 +2,7 @@
 	"compilerOptions": {
 		"module": "esnext",
 		"moduleResolution": "node",
-		"lib": ["es2015"],
+		"lib": ["es2015", "dom"],
```
fishcharlie: What is the reasoning behind this change?

packages/tsconfig.base.json
```diff
@@ -2,7 +2,7 @@
 	"compilerOptions": {
 		"module": "commonjs",
 		"moduleResolution": "node",
-		"lib": ["es2019"],
+		"lib": ["es2019", "dom"],
```
fishcharlie: What is the reasoning behind this change?

fishcharlie (changes requested):
I'm kinda of the opinion that this is a breaking change. What do you think @KennethWKZ?

KennethWKZ:
Thanks for the review. I force-pushed a restructured branch — 8 commits, one concern each, same final content. The important change since your review:

**The nested `$SET` behavior is now opt-in and the default is unchanged.** You were right that it was a breaking change. I had read replace-on-`$SET` as a bug, but your tests named it deliberately (`"Should throw error if trying to replace object without nested required property"`), so the intent was clear and I was wrong to change it. Both of those tests are restored verbatim and passing.

What I was actually hitting is that there was no way to express a nested attribute update at all. DynamoDB supports `SET data.#name = :value` natively, but on `main`:

- the whole-object form replaces the map and trips the nested `required` check
- the dotted-key form `{"data.name": "x"}` produces `ExpressionAttributeNames: {"#a0": "data.name"}`, and since `#a0` substitutes the whole string, that targets a top-level attribute whose *name* contains a dot rather than a document path

So `{merge: true}` is now purely additive — it emits `SET #a0.#a1 = :v2` with separate placeholders. Nothing changes unless you ask for it.

**Still breaking:** `fix(item): return custom types consistently from save and get`. `save` returned the underlying value (`number` for a `Date`) while `get` returned the custom type. That commit is independent of the other six — happy to pull it into its own PR against the next major and let the rest land as a patch, whichever you prefer.

Also worth flagging: `main` does not currently compile. `lib/utils/merge_objects.ts:50` fails with `Operator '+=' cannot be applied to types 'number' and 'T'`. One commit here fixes it.

Test suite is 2401 passing. The 29 failures in `test/Logger.js` are pre-existing on `main` — same count, same tests, unrelated to this branch.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, made the design decisions, reviewed all changes, and approved this comment before posting.

.vscode/settings.json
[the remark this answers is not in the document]
KennethWKZ: Reverted — the file is now byte-identical to `main`, so it's out of the diff entirely.

For the record though, I think the Copilot comment has it backwards. Checking VS Code 1.130.0's `typescript-language-features` manifest, both keys are declared, and it's `typescript.tsdk` that carries the deprecation notice pointing at `js/ts.tsdk.path` — not the other way around.

The intent was that contributors get IntelliSense from the repo's own TypeScript rather than whatever version their editor ships. It's editor-only; builds and CI already resolve `node_modules/.bin/tsc` either way, so nothing about running the repo depends on it. Removing it changed nothing in build, `test:types`, or the suite.

Happy to raise it separately as its own change if you think the convention is worth having repo-wide. It doesn't belong in this PR.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/test/types/tsconfig.json
KennethWKZ: No good reasoning — that was a local environment artifact on my side, not something the code needed. Both reverted to `["es2019"]` and `["es2015"]`.

I verified after reverting: all three packages build and `test:types` passes. Adding `dom` to a Node library's `lib` would also have let browser globals type-check in code that would throw at runtime, so it was actively wrong to include.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/tsconfig.base.json
KennethWKZ: Same answer as the `test/types/tsconfig.json` thread: this was a local environment artifact on my side, not something the code needed. Reverted to `["es2019"]`.

Verified after reverting — all three packages build and `test:types` passes.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/lib/utils/merge_objects.ts
KennethWKZ: No reason — changed to `as number`. The branch is already guarded by `typeof returnObject[key] === "number"` directly above, so the narrow cast is accurate.

Worth noting this line is why `main` doesn't currently compile: `Operator '+=' cannot be applied to types 'number' and 'T'`.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/lib/Model/index.ts
KennethWKZ: Moved `isPlainObject` to `lib/utils/is_plain_object.ts` with 18 dedicated unit tests in `test/utils/is_plain_object.js`, covering object literals, null-prototype objects, arrays, Sets, Dates, Buffers, `Uint8Array`, class instances and primitives.

I left `flattenForMerge` inline. It closes over the `index` counter that drives expression-attribute naming, so extracting it would mean threading mutable state through the signature for no real gain. Can still do it if you'd rather.

One thing that fell out of writing the tests: the original had both `val instanceof Buffer` and `val instanceof Uint8Array`, but `Buffer` extends `Uint8Array`, so the Buffer branch was unreachable. Dropped it.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/lib/Model/index.ts
KennethWKZ: Covered in the `isPlainObject` thread above: that helper moved to `lib/utils/is_plain_object.ts` with dedicated unit tests.

I left `flattenForMerge` inline — it closes over the `index` counter that drives expression-attribute naming, so extracting it would mean threading mutable state through the signature. Happy to do it if you would prefer.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/lib/Model/index.ts
KennethWKZ: It wasn't, and that's fixed now. Added `merge` to the `model.update` settings table plus an "Updating nested objects" section in `docs/docs_src/guide/Model.md` covering the default replace behavior, what `{merge: true}` changes, and the constraint that a nested document path requires the parent map to already exist.

Every expression string in those examples was taken from actual output rather than written by hand. Also added a `PENDING_CHANGELOG.md` entry.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/lib/Item.ts
KennethWKZ: Agreed, and there was a real bug behind it. The callback branch resolved inside `.then` with `.catch` chained after, so an error thrown by the user's callback was caught by that `.catch` and passed back into the same callback a second time. Confirmed: two invocations, the second carrying the user's own error.

Restructured so the result is built once and both branches share it, with the callback attached via the two-argument `.then(onSuccess, onError)`.

One deviation worth surfacing: every other callback site in the codebase uses `.then().catch()`. This is the only two-argument form. Matching that convention would reintroduce the double-call, so I chose correctness over consistency — but say the word if you'd rather it match and handle the double-call another way.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/lib/Item.ts
```diff
@@ -575,7 +586,8 @@ Item.attributesWithSchema = async function (item: Item, model: Model<Item>): Pro
 		Object.keys(treeNode).forEach((attr) => {
 			if (attr === "0") {
 				// We check for empty objects here (added `typeof node === "object" && Object.keys(node).length == 0`, see PR https://github.com/dynamoose/dynamoose/pull/1034) to handle the use case of 2d arrays, or arrays within arrays. `node` in that case will be an empty object.
-				if (!node || node.length == 0 || typeof node === "object" && Object.keys(node).length == 0) {
+				// Also adding `!Array.isArray(node)` to the check for non array objects
+				if (!node || !Array.isArray(node) || node.length == 0 || typeof node === "object" && Object.keys(node).length == 0) {
 					node = [{}]; // fake the path for arrays
 				}
```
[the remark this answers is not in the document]
KennethWKZ: The conclusion was right, the reasoning wasn't. It doesn't collapse populated objects — I checked array-of-objects input and old and new produce identical paths. What it actually fixes is a crash: for a non-array object, `!node` is false, `node.length == 0` is false (`.length` is `undefined`), and the empty-object check is false, so it fell through to `node.forEach(...)` and threw `node.forEach is not a function`.

I simplified the condition to `!Array.isArray(node) || node.length == 0` because your `!Array.isArray` short-circuit made the remaining clauses unreachable. Compared old and new across 15 input shapes: identical on 11, and the 4 differences are all inputs that previously threw.

Added two regression cases to the `attributesWithSchema` table, both verified failing without the fix.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/lib/Model/index.ts
```diff
@@ -739,6 +782,27 @@ export class Model<T extends ItemCarrier = AnyItem> extends InternalPropertiesCl
...
+						(function flatten (path: string, obj: ObjectType): void {
+							for (const [attr, val] of Object.entries(obj)) {
+								accumulator.ExpressionAttributeNames[`#a${index}`] = attr;
+								const k = `${path}.#a${index++}`;
+								if (val !== null && typeof val === "object") {
+									flatten(k, val);
+								} else {
+									accumulator.ExpressionAttributeValues[`:v${index}`] = val;
+									accumulator.UpdateExpression["SET"].push(`${k} = :v${index++}`);
+								}
```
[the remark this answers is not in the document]
KennethWKZ: Correct, and worse than described. Probed each leaf type through a nested map:

| leaf | before | after |
|---|---|---|
| `["a","b"]` | `#a0.#a3.#a4`, `#a0.#a3.#a6` — list turned into a map keyed `"0"`/`"1"` | `#a0.#a3 = :v4` → `{"L":[...]}` |
| `Buffer.from("hi")` | two paths with `{"N":"104"}`, `{"N":"105"}` | `{"B": <Buffer>}` |
| `new Set(["a"])` | **absent from the expression entirely** | `{"SS":["a"]}` |

The Set case was silent data loss, not just a malformed path. Fixed by using the extracted `is_plain_object` helper. Added four regression tests (array, Set, Buffer, and a nested map to confirm real maps still recurse), all verified failing without the fix.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/test/Model.js
```diff
@@ -2011,18 +2011,42 @@ describe("Model", () => {
...
+			const timestamp = Date.now();
+			const item = await DateModel.create({
+				"id": "test",
+				"someDate": timestamp
+			});
+
+			// Verify item.someDate is a Date object after create
+			expect(item.someDate).toBeInstanceOf(Date);
+			expect(item.someDate.getTime()).toEqual(timestamp);
+		});
```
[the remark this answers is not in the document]
KennethWKZ: Fair. Added a `getItem` mock and a `get` call, and it now asserts both paths resolve to the same `Date`.

Being straight about coverage though: I reverted each half of the `Item.save` change to see what this test actually guards. It catches the `returnItem` conform, but passes with the create-side conform reverted. So the in-memory mutation — the half that drives the breaking change — isn't covered by it.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

packages/dynamoose/test/Model.js
```diff
@@ -4732,7 +4824,10 @@ describe("Model", () => {
 
 			it("Should not delete keys from object", () => {
 				const obj = {"id": 1, "name": "Bob"};
+				const oldWarn = console.warn; // eslint-disable-line no-console
+				console.warn = () => {}; // eslint-disable-line no-console
 				User.transaction.update(obj, utils.empty_function);
+				console.warn = oldWarn; // eslint-disable-line no-console
```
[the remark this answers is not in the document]
KennethWKZ: Done, wrapped in `try/finally`.

Note there are five other identical `console.warn` overrides in this file that predate this branch and have the same issue. Left them alone as out of scope, but happy to fix them in a follow-up.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, reviewed all changes, and approved this comment before posting.

KennethWKZ:
Follow-up on one thing I said earlier, and a small change to the branch.

I offered to split `fix(item): return custom types consistently from save and get` into its own PR. I'd like to withdraw that offer, because I think it framed the commit wrongly.

The change isn't really "save returns a `Date` now". It's that a schema declared type should be what you get back from every method. As it stands, `Date` in the schema means a `Date` from `get` and a `number` from `save` — the schema means one thing on read and another on write. Splitting the commit out would land the rest of the branch while leaving that inconsistency in place, which seems worse than the breaking change it avoids.

While checking this I found a loose end of my own. The earlier version of this branch had commented out an assertion in `Should return correct result after saving with timestamps`:

```js
// Skip toJSON comparison test due to Date serialization differences
// expect(result.toJSON()).toEqual(result2.toJSON());
```

That comment was stale — left over from an earlier iteration. The assertion passes with the fix as it stands now, so disabling it was unnecessary and it hid the best evidence for the change. I've restored it and amended it into the same commit (no fixup commit; branch is force-pushed at `c562c2b`).

That assertion is the useful one here: it checks that an item, saved and then re-saved, serializes identically. It's the invariant that was actually broken, and it now holds. `save` and `get` also serialize identically, which is the consistency the commit is about.

So: still a breaking change, and still your call on which release it belongs to. I'd just rather it be a version decision than a reason to separate it from the fixes it belongs with. If you'd prefer it target the next major, that's completely reasonable — I'd only push back on shipping the other commits while `save` and `get` still disagree.

Test suite is unchanged at 2401 passing, with the same 29 pre-existing `test/Logger.js` failures that are present on `main`.

---

This comment was drafted by an AI agent (Claude Code, Opus 5). The human contributor directed the work, made the design decisions, reviewed all changes, and approved this comment before posting.

PR #1785, opened 2026-07-12
perf: ~2x faster scan/query item conversion (fixes #1719)

### LINKED ISSUE #1719: [BUG] significant scan performance difference between AWS DynamoDB SDK and Dynamoose
### Summary:
Scanning of table containing 370 items is significantly slower using Dynamoose vs AWS SDK for the same query. In Dynamoose config, timeout was set to 1000ms, however the total time taken was ~4000ms using Dynamoose parallel scan. The equivalent AWS SDK approach sending parallel scan commands and calling unmarshall on the items gives about 1000-2000ms (closer to 1000 on average). The object schema is fairly big, possible computational bottleneck processing list of complex objects.

#### Schema
```
const anonymizedSchema = {
  field1: { type: String, hashKey: true },
  field2: { type: String, rangeKey: true }, // <schema-version>-<data-version>-<build-version>
  field3: { type: String, required: true },
  field4: { type: String, required: true },
  field5: { type: String, enum: ["UI", "API"], required: true },
  field6: {
    type: String,
    required: false,
    index: { name: "origin-index", global: true, rangeKey: "field2", project: true }
  },
  field7: { type: String, required: true },
  field8: { type: String },
  field9: { type: String, enum: ["API", "PORTAL"], required: true },
  field10: { type: [String, null], required: false },
  field11: searchConfigSchema, // Substituted with the previously anonymized nested schema
  field12: { type: Boolean, required: false },
  field13: {
    type: Object,
    schema: {
      field14: Boolean,
      field15: Boolean,
      field16: { type: Array, schema: [String] },
      field17: Boolean,
      field18: {
        type: Object,
        schema: {
          field19: String,
          field20: String,
          field21: String,
          field22: String
        }
      }
    }
  },
  field23: {
    type: Object,
    schema: {
      field24: Boolean,
      field25: {
        type: Array,
        schema: [
          {
            type: Object,
            schema: {
              field26: String,
              field27: String,
              field28: { type: String, enum: ["RULE_BASED", "ML_BASED"] },
              field29: String,
              field30: { type: Array, schema: [String] },
              field31: Boolean,
              field32: {
                type: Object,
                schema: {
                  field33: {
                    type: Object,
                    schema: {
                      field34: String,
                      field35: String,
                      field36: String,
                      field37: String,
                      field38: Number,
                      field39: {
                        type: Object,
                        schema: {
                          field40: { type: Array, schema: [{ type: Object, schema: { field41: String, field42: String } }] },
                          field43: { type: Array, schema: [{ type: Object, schema: { field44: String, field45: String } }] }
                        }
                      }
                    }
                  },
                  field46: {
                    type: Array,
                    schema: [
                      {
                        type: Object,
                        schema: {
                          field47: String,
                          field48: String,
                          field49: String,
                          field50: {
                            type: Object,
                            schema: {
                              field51: { type: String, enum: ["VISITED"] },
                              field52: { type: Array, schema: [String] }
                            }
                          },
                          field53: {
                            type: Array,
                            schema: [{ type: Object, schema: { field54: String, field55: String } }]
                          }
                        }
                      }
                    ]
                  },
                  field56: {
                    type: Object,
                    schema: {
                      field57: { type: String, enum: ["CARD_BANNER", "LINK_BANNER"] },
                      field58: {
                        type: Object,
                        schema: {
                          field59: {
                            type: String,
                            enum: ["TWO_COLUMN", "ONE_COLUMN", "INLINE"]
                          },
                          field60: String,
                          field61: {
                            type: Object,
                            schema: {
                              desktop: { type: Object, schema: { field62: Number, field63: Number } },
                              tablet: { type: Object, schema: { field64: Number, field65: Number } },
                              mobile: { type: Object, schema: { field66: Number, field67: Number } }
                            }
                          },
                          field68: {
                            type: Object,
                            schema: {
                              desktop: { type: Object, schema: { field69: Number, field70: Number } },
                              tablet: { type: Object, schema: { field71: Number, field72: Number } },
                              mobile: { type: Object, schema: { field73: Number, field74: Number } }
                            }
                          }
                        }
                      },
                      field75: {
                        type: Object,
                        schema: {
                          field76: { type: String, enum: ["LEFT", "CENTER"] },
                          field77: { type: Object, schema: { field78: Boolean, field79: String } },
                          field80: String,
                          field81: {
                            type: Object,
                            schema: {
                              desktop: { type: Object, schema: { field82: Number } },
                              tablet: { type: Object, schema: { field83: Number } },
                              mobile: { type: Object, schema: { field84: Number } }
                            }
                          },
                          field85: String
                        }
                      },
                      field86: {
                        type: Object,
                        schema: {
                          field87: { type: Object, schema: { field88: Boolean, field89: String } },
                          field90: {
                            type: Object,
                            schema: {
                              desktop: { type: Object, schema: { field91: Number } },
                              tablet: { type: Object, schema: { field92: Number } },
                              mobile: { type: Object, schema: { field93: Number } }
                            }
                          },
                          field94: String,
                          field95: String
                        }
                      },
                      field96: {
                        type: Object,
                        schema: {
                          field97: Boolean,
                          field98: { type: String, enum: ["LEFT", "RIGHT"] },
                          field99: String
                        }
                      },
                      field100: {
                        type: Object,
                        schema: {
                          field101: String,
                          field102: {
                            type: Object,
                            schema: {
                              field103: Boolean,
                              field104: String
                            }
                          }
                        }
                      },
                      field105: {
                        type: Object,
                        schema: {
                          field106: { type: Object, schema: { field107: Boolean } },
                          field108: { type: Object, schema: { field109: Boolean } }
                        }
                      }
                    }
                  }
                }
              }
            }
          ]
        }
      }
    }
  }
};
```

#### Model
```
// configuration options
ConfigurationOptions = {
  region: 'region',
  httpOptions: {
    timeout: 1000,
    connectTimeout: 1000
  },
  maxRetries: 4,
  maxRedirects: 10,
  retryDelayOptions: {
    base: 100
  },
  paramValidation: true
};
```

#### General
```
// #region dynamoose query
const beforedynamoose = now();
// eslint-disable-next-line @typescript-eslint/no-explicit-any
let payload: any = await model.model.scan().parallel(16).consistent().exec();
const afterdynamoose = now();
console.log('dynamoose time', afterdynamoose - beforedynamoose);
// #endregion

// #region sdk query
const beforesdk = now();
const client = new DynamoDBClient({region: 'region'});
// generate 16 parallel scans
const scanCommands = new Array(16).fill(0).map((_, index) => new ScanCommand({
  TableName: "table-name",
  ConsistentRead: true,
  Segment: index,
  TotalSegments: 16
}));
const results = await Promise.all(scanCommands.map(async (command) => client.send(command)));
const parsed = flatten(results.map((r) => r.Items !== undefined ? r.Items.map((item) => unmarshall(item)) : []));
const aftersdk = now();
console.log('sdk time', aftersdk - beforesdk);
// #endregion
```

Image
Logger output:
```
aws:dynamodb:describeTable:response - 

aws:dynamodb:scan:request - {
    "TableName": "table-name",
    "ConsistentRead": true,
    "TotalSegments": 16,
    "Segment": 0
}
aws:dynamodb:scan:request -
aws:dynamodb:scan:request -
aws:dynamodb:scan:request -
```

Image

### Environment:

Operating System:
Operating System Version:
Node.js version (`node -v`):
NPM version: (`npm -v`):
Dynamoose version:

### Other:
- [ /] I have read through the Dynamoose documentation before posting this issue
- [ /] I have searched through the GitHub issues (including closed issues) and pull requests to ensure this issue has not already been raised before
- [/ ] I have searched the internet and Stack Overflow to ensure this issue hasn't been raised or answered before
- [ /] I have tested the code provided and am confident it doesn't work as intended
- [ /] I have filled out all fields above
- [ x] I am running the latest version of Dynamoose

KennethWKZ:
# perf: ~2x faster scan/query item conversion (fixes #1719)

Closes dynamoose/dynamoose#1719

## Summary

Converting large scan/query result sets was dominated by redundant work that is independent of the actual item data. This removes it, cutting the CPU cost of `Item.objectFromSchema` — the per-item work behind every scan, query, get, batchGet, and transaction read — with **no behavioral change**.

No public API, type, or serialization change.

## Problem

Reported in #1719: scanning a deeply nested schema was several times slower in Dynamoose than the raw AWS SDK `unmarshall`. Profiling the conversion loop showed the time was not in `deep_copy`; it was in three places, all quadratic or redundant:

1. **Schema introspection recomputed for every item.** `getAttributeValue`, `getAttributeTypeDetails`, `getAttributeType`, and `attributes()` are pure functions of the immutable `schemaObject`, yet were recomputed from scratch on every call — thousands of calls per item on a nested schema.

2. **`js-object-utilities`' `entries()` is O(n²).** It calls `isCircular(value)` before descending into every nested node, and each `isCircular` walks that node's entire subtree. `objectFromSchema` calls it per item, and `Item.fromDynamo` calls it again for the `Uint8Array → Buffer` pass.

3. **`getTypePaths` merged its nested results by spreading.** `result = {...result, ...recurse(...)}` copies the whole accumulated object once per entry, which is quadratic in the length of a list. This one is pre-existing on `main` and unrelated to the other two.

### 1. Memoize schema introspection (`lib/Schema.ts`)

Per-`Schema`-instance `WeakMap` caches for `getAttributeValue`, `getAttributeTypeDetails`, `getAttributeType`, and the no-argument `attributes()` result. The schema definition is immutable after construction, so these are safe. `attributes()` returns a defensive copy, and stores a copy in the cache, so a caller mutating the returned array cannot corrupt it.

The memo key is built only from the requested key's own ancestor paths in `typeIndexOptionMap`. Serializing or otherwise inspecting the whole map would cost O(map size) per lookup, and `getTypePaths` produces one entry per list element — which would make converting a long list quadratic.

**Cache invalidation:** the `expires`/TTL feature injects an attribute into `schemaObject` *after* the Schema is built. That write now goes through a new `setSchemaAttribute` internal property which pairs it with cache invalidation, so a future post-construction mutation cannot silently leave the caches stale.

### 2. Linear-time object flattening (`lib/utils/object_entries.ts`, `lib/Item.ts`)

`utils.object_entries` is a drop-in linear-time replacement for `js-object-utilities`' `entries`, guarding against cycles with an ancestor-path `Set` (the technique `deep_copy` already uses) instead of a full subtree walk per node. Swapped in at the three per-item call sites in `Item.ts`.

> **Note:** per @fishcharlie's review, this belongs upstream in `js-object-utilities` rather than here. It is output-identical to the current `entries` (including on cyclic input — verified by differential, see Testing), so it should be shippable there as a patch release. **This PR is parked as-is until that lands**; once it is published I will delete `lib/utils/object_entries.ts`, revert `Item.ts` to `utils.object.entries`, and bump the dependency. The benchmark table below reports both totals so the split is clear.

### 3. Merge type paths in place (`lib/Schema.ts`)

`getTypePaths` now uses `Object.assign(result, ...)` instead of spreading into a new object.

## Benchmarks

Environment: **Node v24.18.0** (V8 13.6.233.17), macOS 25.5.0 arm64, **Apple M5, 10 logical cores, 24 GB RAM**, default V8 heap limit (~4.3 GB), no `NODE_OPTIONS` set. Each figure is the **median of 5 timed runs after 3 warmup runs**, measured in-process with `process.hrtime.bigint()` and no network (conversion CPU only). Single-threaded, no core pinning, so absolute numbers will vary by machine — the ratios are the point.

The "memoization only" column is what this PR delivers if the `entries` change moves upstream, so the two contributions can be read separately.

| scenario | main | memoization only | + `entries` (this PR) | full vs main |
|---|---:|---:|---:|:---:|
| wide flat schema, 30 attrs × 2000 items | 152.94 ms | 70.80 ms | 72.00 ms | **2.12x** |
| wide flat schema, 80 attrs × 500 items | 103.87 ms | 46.36 ms | 46.05 ms | **2.26x** |
| deeply nested schema (depth 20) × 300 items | 43.88 ms | 18.70 ms | 13.75 ms | **3.19x** |
| 2000 element list, single type | 3.22 ms | 1.90 ms | 1.74 ms | **1.85x** |
| 400 element list, multiple types | 1.19 ms | 0.69 ms | 0.70 ms | **1.70x** |
| 1600 element list, multiple types (fromDynamo) | 4.75 ms | 2.81 ms | 2.75 ms | **1.73x** |
| 1600 element list, multiple types (toDynamo) | 4.49 ms | 2.58 ms | 2.58 ms | **1.74x** |
| 4000 element list, multiple types | 11.25 ms | 6.28 ms | 6.72 ms | **1.67x** |
| 800 element list of multi-type objects | 116.41 ms | 27.24 ms | 24.36 ms | **4.78x** |
| 500 unknown attributes (`saveUnknown`) | 17.59 ms | 16.18 ms | 16.53 ms | 1.06x |
| `attributes()`, 100 attrs, 5000 calls | 170.06 ms | 0.16 ms | 0.16 ms | **1063x** |
| raw type lookups × 50000 | 35.67 ms | 3.98 ms | 3.92 ms | **9.10x** |

Reading the split: **memoization alone is ~1.7–2.3x** on typical item shapes (4.3x on the multi-type-object case, which is mostly the `getTypePaths` fix), and the `entries` change adds most of its remaining benefit on deeply nested objects. The `saveUnknown` case is ~flat by design — every attribute key there is unique, so there is nothing to memoize.

## Testing

- **Differential behavior check.** 150 randomized schemas (nesting, arrays of objects, arrays of arrays, sets, multi-type scalars, dates in ms/seconds/iso, buffers, optionals, defaults, `saveUnknown`), each item run through both `toDynamo` and `fromDynamo`, with all output and every thrown error stable-hashed. **3,000 conversions, `sha256 912315d8…`, byte-identical between `main` and this branch.**
- **New unit tests** for `object_entries` (13): equivalence with `js-object-utilities` `entries` across nested objects, arrays, `Buffer`, `Uint8Array`, `Date`, empty objects and numeric keys; pre-order preservation; termination on direct and indirect cycles.
- **New regression tests** in `test/Schema.js`: TTL cache invalidation after `Table` injects the attribute; `attributes()` cache cannot be corrupted by a caller mutating the returned array; memo key reuse when `typeIndexOptionMap` differs only in unrelated keys, and non-reuse when it changes the matched type; `getTypePaths` returns an entry per element for a long multi-type list.
- **Full suite:** 2404 passing, 0 failing. `npm run lint`, `npm run test:types`, and `npm run build:sourcemap` all clean.

## Backward compatibility

None broken. Conversion output is identical (see the differential hash above). Caches are per-`Schema`-instance `WeakMap`s, so they are collected with the schema and are transparent to callers.

---

This pull request was created by an AI agent (Claude Code, Opus 5). The human contributor reviewed the changes, ran the benchmarks and tests, and approved the submission.

### DIFF
--- a/packages/dynamoose/lib/utils/object_entries.ts
+++ b/packages/dynamoose/lib/utils/object_entries.ts
@@ -1,0 +1,35 @@
+import {ObjectType} from "../General";
+
+// A drop-in, linear-time replacement for `js-object-utilities`' `entries` helper.
+//
+// The library version calls `isCircular(value)` before recursing into every nested
+// node, and each `isCircular` performs a full walk of that node's subtree. That makes
+// flattening a deeply nested object O(n^2), which dominates the cost of converting large
+// scan/query result sets (issue #1719).
+//
+// This version guards against cycles in O(n) by tracking the ancestors on the current
+// recursion path in a Set and refusing to descend into a value already on that path.
+// For acyclic inputs (every DynamoDB item, and any object that has already been run
+// through `deep_copy`, which strips cycles) the output is identical to the library's,
+// key-for-key and in the same order.
+function object_entries (object: ObjectType, existingKey = "", ancestors: Set<unknown> = new Set()): [string, any][] {
+	const accumulator: [string, any][] = [];
+	// Mark this node as being on the current recursion path. Sibling references that share
+	// the same (non-cyclic) object still descend because each node is removed on the way
+	// back up; only a value that is an ancestor of itself (a true cycle) is skipped.
+	ancestors.add(object);
+	for (const [key, value] of Object.entries(object)) {
+		const keyWithExisting = `${existingKey ? `${existingKey}.` : ""}${key}`;
+		accumulator.push([keyWithExisting, value]);
+		if (typeof value === "object" && !(value instanceof Buffer) && value !== null && !ancestors.has(value)) {
+			const nested = object_entries(value, keyWithExisting, ancestors);
+			for (let i = 0; i < nested.length; i++) {
+				accumulator.push(nested[i]);
+			}
+		}
+	}
+	ancestors.delete(object);
+	return accumulator;
+}
+
+export default object_entries;
--- a/PENDING_CHANGELOG.md
+++ b/PENDING_CHANGELOG.md
@@ -6,6 +6,11 @@
 
 - Added support for enabling and configuring DynamoDB Streams through Table options
 
+### Performance
+
+- Improved performance of converting items returned from `scan` and `query` (~2x faster on large result sets). `Schema` now memoizes `attributes`, `getAttributeValue`, `getAttributeType`, and `getAttributeTypeDetails` per schema, and object flattening no longer re-walks each nested value's subtree to detect circular references
+- Improved performance of items containing a long list of attributes with multiple possible types (~4x faster for a 800 element list). Building the type paths for such an item no longer copies the accumulated result once per element
+
 ### Bug Fixes
 
 - Fixed `deep_copy` dropping sibling properties that share the same object reference (e.g. a single `Date` used for both `createdAt` and `updatedAt`, or a single address used for both `billingAddress` and `shippingAddress`). Circular reference detection now tracks only the current ancestor path rather than every object ever visited, which also removes an infinite-recursion risk on self-referencing arrays and a shared-reference leak on class instances containing circular properties
--- a/packages/dynamoose/lib/Item.ts
+++ b/packages/dynamoose/lib/Item.ts
@@ -73,7 +73,7 @@ export class Item extends InternalPropertiesClass<ItemInternalProperties> {
 	}
 	static fromDynamo (object: AttributeMap): ObjectType {
 		const result = awsConverter().unmarshall(object);
-		utils.object.entries(result).forEach(([key, value]) => {
+		utils.object_entries(result).forEach(([key, value]) => {
 			if (value instanceof Uint8Array) {
 				utils.object.set(result, key, Buffer.from(value));
 			}
@@ -696,7 +696,7 @@ Item.objectFromSchema = async function (object: any, model: Model<Item>, setting
 				}
 			}
 		};
-		utils.object.entries(returnObject).filter((item) => item[1] !== undefined && item[1] !== dynamooseUndefined).map(checkTypeFunction);
+		utils.object_entries(returnObject).filter((item) => item[1] !== undefined && item[1] !== dynamooseUndefined).map(checkTypeFunction);
 		keysToDelete.reverse().forEach((key) => utils.object.delete(returnObject, key));
 	}
 
@@ -739,7 +739,7 @@ Item.objectFromSchema = async function (object: any, model: Model<Item>, setting
 		});
 	}
 	// DynamoDB Type Handler (ex. converting sets to correct value for toDynamo & fromDynamo)
-	utils.object.entries(returnObject).filter((item) => typeof item[1] === "object").forEach((item) => {
+	utils.object_entries(returnObject).filter((item) => typeof item[1] === "object").forEach((item) => {
 		const [key, value] = item;
 		let typeDetails;
 		try {
--- a/packages/dynamoose/lib/Schema.ts
+++ b/packages/dynamoose/lib/Schema.ts
@@ -834,6 +834,8 @@ interface SchemaInternalProperties {
 	getDefaultMapAttribute: (attribute: string) => string;
 	getIndexAttributes: () => {index: IndexDefinition; attribute: string}[];
 	getTimestampAttributes: () => GetTimestampAttributesType;
+	clearCaches: () => void;
+	setSchemaAttribute: (attribute: string, definition: AttributeDefinition) => void;
 }
 
 type GetTimestampAttributesType = ({
@@ -1095,6 +1097,20 @@ export class Schema extends InternalPropertiesClass<SchemaInternalProperties> {
 		this.setInternalProperties(internalProperties, {
 			"schemaObject": parsedObject,
 			"settings": parsedSettings,
+			// Drops the per-schema memoization caches. Must be called whenever schemaObject
+			// is mutated after construction (e.g. the expires/TTL attribute injected by Table),
+			// otherwise the caches would serve a stale view of the schema (issue #1719).
+			"clearCaches": (): void => {
+				getAttributeValueCache.delete(this);
+				getAttributeTypeDetailsCache.delete(this);
+				getAttributeTypeCache.delete(this);
+				attributesNoArgCache.delete(this);
+			},
+			// Adds an attribute after construction. Writing to `schemaObject` directly would leave the caches below holding a stale view of the schema.
+			"setSchemaAttribute": (attribute: string, definition: AttributeDefinition): void => {
+				this.getInternalProperties(internalProperties).schemaObject[attribute] = definition;
+				this.getInternalProperties(internalProperties).clearCaches();
+			},
 			"getMapSettingValuesForKey": (key: string, settingNames?: string[]): string[] => utils.array_flatten(mapSettingNames.filter((name) => !settingNames || settingNames.includes(name)).map((mapSettingName) => {
 				const result = this.getAttributeSettingValue(mapSettingName, key);
 				if (Array.isArray(result)) {
@@ -1348,8 +1364,20 @@ export class Schema extends InternalPropertiesClass<SchemaInternalProperties> {
 	}
 	getAttributeType (key: string, value?: ValueType, settings?: SchemaGetAttributeTypeSettings): string | string[] {
 		try {
+			// The cache below is keyed on the standardized key alone since this branch never uses `value` or `settings`. The value dependent `unknownAttributeAllowed` path in the `catch` below is left uncached.
+			const standardKey = key.replace(/\.\d+/gu, ".0");
+			let cache = getAttributeTypeCache.get(this);
+			if (!cache) {
+				cache = new Map();
+				getAttributeTypeCache.set(this, cache);
+			}
+			if (cache.has(standardKey)) {
+				return cache.get(standardKey);
+			}
 			const typeDetails = this.getAttributeTypeDetails(key);
-			return Array.isArray(typeDetails) ? (typeDetails as any).map((detail) => detail.dynamodbType) : typeDetails.dynamodbType;
+			const result = Array.isArray(typeDetails) ? (typeDetails as any).map((detail) => detail.dynamodbType) : typeDetails.dynamodbType;
+			cache.set(standardKey, result);
+			return result;
 		} catch (e) {
 			if (settings?.unknownAttributeAllowed && e.message === `Invalid Attribute: ${key}` && value) {
 				return Object.keys(Item.objectToDynamo(value, {"type": "value"}))[0];
@@ -1487,7 +1515,8 @@ export class Schema extends InternalPropertiesClass<SchemaInternalProperties> {
 			}
 
 			if (isObject) {
-				result = {...result, ...this.getTypePaths(value, {...settings, "previousKey": fullKey})};
+				// Merged in place. Spreading into a new object copied the whole accumulated result once per entry, which made this quadratic in the length of a list.
+				Object.assign(result, this.getTypePaths(value, {...settings, "previousKey": fullKey}));
 			}
 
 			return result;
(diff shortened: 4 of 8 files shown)


packages/dynamoose/lib/utils/object_entries.ts
fishcharlie, line 1: I'm skeptical of putting this in here. `js-object-utilities` is a project I built and is fully open source. Feels like we should just make this patch there instead. Is there any reason why it needs to be Dynamoose specific?

If we move it upstream I can merge that PR and publish a new release so Dynamoose can take advantage of it.

fishcharlie (commented):
Very first review. Still need to do a more in depth review.

packages/dynamoose/lib/utils/object_entries.ts
KennethWKZ: That's fair, and I agree — there's nothing Dynamoose-specific about it. It's a pure object-flattening function with no coupling to schemas or anything else in this codebase. Upstream is the right home for it.

One thing that should make the upstream patch easy to accept: **it's output-identical to the current `entries`, including on cyclic input.** I ran a differential across 2,607 randomized shapes (nested objects, arrays, buffers, dates, nulls, mixed depth) plus targeted cycle cases — direct self-cycle, indirect cycle, deep back-reference, array cycle — and got 0 mismatches. Same for shared-but-acyclic references (sibling shared refs, diamonds, the same `Date` on two keys, a repeated object in an array).

So it isn't a behavior change, just a cheaper way to compute the same answer: `isCircular(value)` walks the value's whole subtree at every node, whereas tracking the ancestors on the current recursion path answers the same question in O(1) per node. That should be shippable as a patch release.

The core of it:

```js
const main = (object, existingKey = "", ancestors = new Set()) => {
	const accumulator = [];
	ancestors.add(object);
	for (const [key, value] of Object.entries(object)) {
		const keyWithExisting = `${existingKey ? `${existingKey}.` : ""}${key}`;
		accumulator.push([keyWithExisting, value]);
		if (typeof value === "object" && !(value instanceof Buffer) && value !== null && !ancestors.has(value)) {
			accumulator.push(...main(value, keyWithExisting, ancestors));
		}
	}
	ancestors.delete(object);
	return accumulator;
};
```

Each node is removed from the set on the way back up, so sibling references to the same (non-cyclic) object still descend — only a value that is its own ancestor is skipped. Take it as-is if it's useful.

While you're in there, two adjacent things worth a look, both pre-existing and neither caused by the above:

1. **`Uint8Array` gets enumerated byte-by-byte.** `entries` treats `Buffer` as a leaf but not other `ArrayBuffer` views, so a `Uint8Array` binary attribute expands into one entry per byte — a 64KB attribute produces 65,538 entries (~8.5ms). This matters for Dynamoose specifically because `Item.fromDynamo` runs on the AWS SDK's `unmarshall` output, where binary attributes come back as `Uint8Array`, not `Buffer`. Current and patched versions behave identically here, so it's orthogonal to the perf patch, but treating `ArrayBuffer` views as leaves would be a nice fix. (Credit to Copilot for spotting this one on the review.)

2. **`circularKeys`-based helpers and shared references.** Separately in this PR I fixed `deep_copy` dropping sibling properties that share an object reference (one `Date` used for both `createdAt` and `updatedAt`, for example) — the cause was global-visited tracking rather than ancestor-path tracking. My differential says `entries` does *not* have that bug, but it may be worth checking whether anything else built on `circularKeys` does.

**Plan from my side:** I'll leave this branch as-is for now rather than churn it, and once you've patched and published upstream I'll drop `lib/utils/object_entries.ts`, revert `Item.ts` to `utils.object.entries`, and bump the dependency.

Worth flagging on the numbers: with the `entries` change removed, this PR on its own lands at roughly **1.8x** rather than 2.4x, with the remaining ~1.3x coming from the dependency bump. I'll update the description with re-measured figures when I make that change — the current benchmark table and differential hash in the description were measured against the first version of the commit and are stale.

I've also addressed both Copilot comments in the meantime — the `attributes()` cache now stores a defensive copy on the initial miss as well, and I'll add the `Uint8Array` equivalence case to the test table. I found and fixed two further issues while re-reviewing: the memo key was serializing the entire `typeIndexOptionMap` on every lookup, which made converting a long list of multi-type values quadratic (worse than `main` at 4,000 elements), and `getTypePaths` was spreading its accumulator once per entry, also quadratic. That second one is pre-existing on `main` rather than something this PR introduced; fixing it took an 800-element multi-type list from 116ms to 27ms.

---

This comment was written by an AI agent (Claude Code, Opus 5). The human contributor reviewed the analysis, ran the benchmarks and tests, and approved the submission.

tayyabferozi:
> Very first review. Still need to do a more in depth review.

Following

